PR #85, opened 2018-08-30
Feature/commercial 50 kyle

KyleMicrocare:
Module is complete, afaik. Can safely be merged into develop50 and further upstream. ~~Has some issues with values being reported via UI.~~ Issue has been fixed


KyleMicrocare:
Need to add last fix for reported values - modbus address values were off by one for SDM630

KyleMicrocare:
I'm an idiot - that commit was automatically added to the PR. carry on.

sfeilmeier:
Hi Kyle,
sorry for the very long delay. I made the following adjustments while merging the code:
- git repository was missing the "test" folder which causes an error when importing the project in eclipse. For git to consider an empty folder you need to add a '.gitignore' file inside.
- Renamed 'SignedIntDoubleWordElement' to 'SignedDoublewordElement'; this is a compromise, but if I wanted to do it right, I would have to rename all of those Elements; which touches every file in the repository - and I would like to avoid that.
- Autoformatting the files (Ctrl + Shift + F in Eclipse)
- Remove obsolete package io.openems.edge.meter.eastron.sdm630

KyleMicrocare:
awesome. I'll rebase on your changes locally. Thanks for the feedback.

PR #118, opened 2018-10-30
Solar log development

abraovic:
read only solar log implementation


sfeilmeier:
Hi Ante,

thank you for your pull request. I was fixing some general problems, trying to be as minimally invasive to your source code as possible, to make it easier for you to follow the changes.

- 21e9e31: I had found a problem in the SymmetricPvInverter interface implementation. I had forgotten to export the meter interface; which caused that weird error you faced. 
- 55c1499: I fixed the way you are applying the power limitation. All of this needs to happen via Channels and without 'simple' interface methods. I am executing the calculation on every setNextWriteValue event.
- 52eb829: I implemented a simple Fix-Power-Limit Controller for PV-Inverters. It allows setting a power limit via configuration.

Some points I did not touch:

- You need to make sure, that (as many as possible) Channels of the inherited interfaces are being filled. For example in your implementation you should not introduce a new 'PAC' channel, but rather use the provided 'ACTIVE_POWER' channel. This way your implementation will be usuable as a generic SymmetricMeter throughout OpenEMS - e.g. for visualizing it in the UI.
- We stick to the code formatting that is provided by default in Eclipse IDE. It makes all the source code in the project having the same style, making it easier to read. Please apply it by calling 'Source' -> 'Format' in the menu; or easier: Ctrl + Shift + F.

I was able to validate that reading data from a SolarLog interface is already working. Writing is not yet working - there seems to be some kind of 'Illegal Address' issue. I also checked the firmware version of the SolarLog to be above 3.3.0, so it should be ok with the newer modbus protocol.

Thanks again and feel free to ask your questions.

Regards,
Stefan

sfeilmeier:
Hi Ante,

I made some improvements and bug-fixes. Please check if the implementation is still working. If you approve, we can merge it.

Regards,
Stefan

abraovic:
Hi @sfeilmeier 
you have removed [line 75](https://github.com/OpenEMS/openems/pull/118/commits/fec8553677327053c0087a6a35d4a9d0c28e0c82#diff-cd7a82be7b2d1b2492016a640fa8ef07L75) from SolarLog.java and [line 33](https://github.com/OpenEMS/openems/pull/118/commits/7a72180cf62d02adff44ea39fe881447e19d9bf3#diff-05113ca19b4938cd95d8dd817acc1582L33) from Config.java which was used to set limit in max active power. 

And since that is used on [line 180](https://github.com/OpenEMS/openems/pull/118/commits/fec8553677327053c0087a6a35d4a9d0c28e0c82#diff-cd7a82be7b2d1b2492016a640fa8ef07L180) now setting power limit won't work this way. It will require percentage. And we can not use that in an emergency cluster mode where we use an actual amount of power in Watts. 

I guess you have removed this because PvInverter interface no longer implements SymmetricMeter. But still, we need a way to set some kind of of max power in order to calculate percentage. 

Rest of code seems to work as it should.

abraovic:
Hi @sfeilmeier, everything seems to work as it was before. So I guess you can merge it.

PR #124, opened 2018-10-31
KACO / Energy Depot App Project

andreashummelsberger:
Added environment "App" which can be built with cordova.
Added login form and modified authentication using wordpress api.
Updated some translations.
Added checkbox in history to get cumulative sums
Added Wordpress API user and edge authentication.
Timedata extended with cumulative option.


sfeilmeier:
Hi Andreas,

I spent the last few hours trying to merge all your changes. The approach I took was to split this huge pull request in smaller ones - starting with the wordpress backend metadata provider and your changes to the UI. This pull request is here: https://github.com/OpenEMS/openems/pull/125. Unfortunately the UI is not building right now. Maybe you and @fabianfnc can have a look?

In general I can see that you are missing some best practices which make it really difficult to merge your code to OpenEMS.
- Split your commits and pull-requests in as small junks of code as possible/feasible. A commit is the smallest unit in git and it cannot be split apart. This is what I was trying now with this pull request - and it is a mess...
- Keep a separate branch with all your changes that are not planned for pushing, e.g. changes of "OpenEMS" to "PRIMUS", colors, and so on. Like this you can easily merge with this branch before going in production and the rest of the code can be kept as close as possible to the OpenEMS develop branch.
- Avoid naming components "energydepot" when in fact they should be generic. In my opinion the "uiwebsocket" components should be merged. What is the reason to have a separate one?

Sorry for being so specific. You need to keep in mind that the project is starting to grow and if we do not keep a certain quality level from now, we will have big problems in future with all kinds of different situations and configurations. I also realize that I still had a lot of references to "FEMS" and "FENECON". Those are removed from code now entirely.

- The login feature seems to be very specific to your own wordpress backend. It is also not easily configurable via environment files. Because of this I left it out of my pull-request for now.
- You introduced a new 'env.backend' = "App". Can you explain why? From my perspective "App" also connects to backend, right?

Thanks again for your contribution.

Regards,
Stefan

sfeilmeier:
Hallo Andreas,

ich denke dieser Pull-Request hier ist veraltet, deshalb schließe ich ihn. Ich würde mich aber freuen, wenn wir deine Änderungen für Wordpress und am UI noch ins Projekt bekommen würden.

Gruß,
Stefan

PR #175, opened 2018-11-20
Refactor Channel options to use OptionsEnum

sfeilmeier:
- Adds an "**Undefined**" enum value that always needs to exist. It is used when the Channel has no value.
- This simplifies reading and writing option values. It can be used like this:
```
IntegerWriteChannel channel = this.channel(ChannelId.PCS_MODE);`
```
Read:
```
PcsMode pcsMode = this.getPcsModeChannel().value().asEnum();
if (this.getPcsMode() != PcsMode.REMOTE) {
    ...
}
```
Write:
```
channel.setNextWriteValue(PcsMode.REMOTE);
```
or:
```
channel.setNextWriteValueFromObject("Remote");
```

### DIFF
--- a/io.openems.edge.common/src/io/openems/edge/common/channel/doc/Level.java
+++ b/io.openems.edge.common/src/io/openems/edge/common/channel/doc/Level.java
@@ -3,16 +3,32 @@ package io.openems.edge.common.channel.doc;
 /**
  * Severity/visibility Level
  */
-public enum Level {
-	INFO(1), WARNING(2), FAULT(3);
-	
+public enum Level implements OptionsEnum {
+	OK(0, "Ok"), //
+	INFO(1, "Info"), //
+	WARNING(2, "Warning"), //
+	FAULT(3, "Fault");
+
 	private final int value;
-	
-	private Level(int value) {
+	private final String name;
+
+	private Level(int value, String name) {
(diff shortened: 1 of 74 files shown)


sfeilmeier:
Hallo Wolfgang,

ich warte hier seit 23 Tagen auf dein Review. Denkst du, du kannst das nächste Woche erledigen?

io.openems.edge.common/src/io/openems/edge/common/channel/doc/Level.java
```diff
@@ -3,16 +3,32 @@
 /**
  * Severity/visibility Level
  */
-public enum Level {
-	INFO(1), WARNING(2), FAULT(3);
-	
+public enum Level implements OptionsEnum {
+	OK(0, "Ok"), //
```
wgerbl: What is thje difference between OK and Info?

sfeilmeier: Good point. I added the following javadoc:
- "OK" indicates, that everything is OK and there are no messages.
- "Info" indicates, that everything is OK, but there is at least one informative messages available.

Differentiation is required for the UI. With "Info" it knows, that it needs to search trough the StateChannels to get the informative message.

PR #185, opened 2018-11-28
[UI] Grid Mode + Apperant Power for Energymonitor

fabianfnc:


ui/src/app/edge/index/energymonitor/chart/section/gridsection.component.ts
```diff
@@ -8,12 +8,13 @@ import { AbstractSection, SvgSquarePosition, SvgSquare, EnergyFlow, SvgEnergyFlo
     templateUrl: './section.component.html'
 })
 export class GridSectionComponent extends AbstractSection {
-
+    private GridMode: number;
```
sfeilmeier: Variables should start with a lower case letter. As mentioned above: create a enum type to keep the gridMode instead of using a plain number.
https://www.typescriptlang.org/docs/handbook/enums.html

ui/src/app/edge/index/energymonitor/chart/section/gridsection.component.ts
```diff
@@ -34,7 +38,12 @@ export class GridSectionComponent extends AbstractSection {
     }
 
     protected getImagePath(): string {
-        return "grid.png";
+        if (this.GridMode == 0 || this.GridMode == 2) {
+            return "offgrid.png"
+        }
+        else {
```
sfeilmeier: The extra line break is not a big deal, but common formatting would be:
```
if (...) {
} else {
}
```

ui/src/app/shared/edge/config.2018.7.ts
```diff
@@ -76,10 +77,23 @@ export class ConfigImpl_2018_7 extends ConfigImpl implements DefaultTypes.Config
         let simulatorDevices: string[] = [];
         let evcsDevices: string[] = [];
         let thresholdDevices: string[] = [];
+        let essType: string[] = [];
```
sfeilmeier: Instead of storing the essType here, you could also store the maxApparentPower here directly. It would save you from the extra 'if' in CurrentDataAndSummary.

ui/src/app/shared/edge/currentdata.2018.7.ts
```diff
@@ -205,6 +219,9 @@ export class CurrentDataAndSummary_2018_7 extends CurrentDataAndSummary {
             let ratio = 0;
             let maxSell = 0;
             let maxBuy = 0;
+            for (let thing of config.essType) {
+                let type = config.things[thing];
+            }
```
sfeilmeier: What is this for? type is not used again later (and could anyway not be used later, because it is only valid within the for-block.

ui/src/app/shared/translate/en.ts
```diff
@@ -3,6 +3,7 @@ export const TRANSLATION = {
         Grid: "Grid",
         GridBuy: "Buy from grid",
         GridSell: "Sell to grid",
+        GridMode: "No Grid Connection!",
```
sfeilmeier: The name should be something like "OffGrid" or "NoGridConnection" instead of "GridMode"

PR #357, opened 2019-02-28
Feature/commercial 50 60

wgerbl:


pooran-c:
Two classes for error channels are added and their respective changes in the Utils.java.

PR #590, opened 2019-06-11
Advanced Access Control and Role Management

sWalbrun:
Hey Stefan,

Metadata.UserBased can be ignored
I guess it is time for a first check of the code: AccessControl has been changed currently only with UI and Edge. For setting up some tests you have to configure two AccessControlProvidersJSON. Therefore you have to downloads the attached files and add them via their file location

[RoleMasterDataV2_I.txt](https://github.com/OpenEMS/openems/files/3276424/RoleMasterDataV2_I.txt)
[RoleMasterDataV2_II.txt](https://github.com/OpenEMS/openems/files/3276425/RoleMasterDataV2_II.txt)
WebConsoleConfiguration

The image shows how to link to the attached files

When logging in with the default password, the channels (of getting started example) should be shown on the UI as before but the AccessControl did an intersection before

Please let me know if everything worked or if this description was too short.


sfeilmeier:
Hi Sebastian,

thanks for the Pull-Request. I tried reading through it, but there are a lot of changes in files that I suspect you only touched by accident, or were auto-formatted using an auto-formatting algorithm that is different to the one Eclipse uses. I don't want to be picky, but could you try to reduce the pull-request to the files you actually need to change?

There is even a [Checkstyle](http://checkstyle.sourceforge.net/) definition in OpenEMS that I am trying to embrace: [cnf/checkstyle.xml](https://github.com/OpenEMS/openems/blob/develop/cnf/checkstyle.xml). 

For the package name `io.openems.backend.metadata.user_based`: underscores in package names are not recommended [by convention](https://stackoverflow.com/questions/3179216/what-is-the-convention-for-word-separator-in-java-package-names). Actually I would suggest for this new implementation to replace the `metadata.file` bundle. From what I know only we as FENECON use it - and only in a couple of specific projects. In my opinion there is not too much sense in keeping multiple file-based metadata providers around.

I tried activating the UserBased component, but the above JSON-files do not seem to be current anymore as "users" is excepted to be a JsonArray.

io.openems.backend.metadata.user_based/src/io/openems/backend/metadata/user_based/UserBased.java
```diff
@@ -0,0 +1,207 @@
...
+        super("Metadata.UserBased");
+    }
+
+    @Activate
+    void activate(Config config) {
+        log.info("Activate [path=" + config.path() + "]");
+        this.path = config.path();
+
+        // Read the data async
+        CompletableFuture.runAsync(this::refreshData);
```
sfeilmeier: The downside of reading the file async is, that you will never get notified about an Exception.

io.openems.backend.metadata.user_based/src/io/openems/backend/metadata/user_based/UserBased.java
```diff
@@ -0,0 +1,207 @@
...
+            // exception occurred. File could not be read
+            return;
+        }
+
+        List<Edge> edges = new ArrayList<>();
+
+        // parse to JSON
+        try {
+            JsonElement config = JsonUtils.parse(sb.toString());
+            JsonArray jUsers = JsonUtils.getAsJsonArray(config, JsonKeys.USERS.value());
```
sfeilmeier: In my opinion, this JsonKeys enum is over-engineered and makes the code harder to read. If I were you I would just use plain string identifiers here. I do not quite see the benefit of making the Enum reusable within `io.openems.common.utils`

io.openems.common/src/io/openems/common/utils/FileUtils.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.io.BufferedReader;
+import java.io.FileReader;
+import java.io.IOException;
+
+public class FileUtils {
+
+    private static final Logger log = LoggerFactory.getLogger(FileUtils.class);
+
+    public static StringBuilder checkAndGetFileContent(String path) {
```
sfeilmeier: I like the idea of introducing a FileUtils class. In my opinion `checkAndGetFileContent` should throw an (OpenEMS)Exception if it fails, otherwise it is hard to handle the error case in the calling method.

In my experience at FENECON, once you have multiple OpenEMS Edge instances setup, nobody is going to actually read the logs and detect errors like that. It's crucial to always throw Exceptions and handle them in a way, they finally become visible - e.g. using StateChannels; see the "RunFailed"-Controller-Channel as an example: https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.core/src/io/openems/edge/core/cycle/CycleWorker.java#L101. The StateChannel "RunFailed" is defined as a "Fault"-Channel and is set when a Controller run()-method failes. This causes eventually the "_sum/State" to be "Fault", which makes it detectable for a central Dashboard in our office.

sfeilmeier:
Puuuh... this is a huge amount of code, touching a lot of files. I don't feel it's quite finished yet. I took some hours to merge it with develop and adjusted some parts where I felt it's fitting better with the rest of the codebase. I also marked some places with "FIXME" where I see need for discussions or improvements.

For my TODO in AccessControlProviderStatic (https://github.com/OpenEMS/openems/pull/590/commits/f33d4aee6b99c0752d9f051ed5bdb54a120ea34a#diff-40866391f1f3790541c229f6bc56412cR54): It will be impossible to define all possible Channel-Adresses. Can we not use some kind of wildcard or regular expression to allow everything? Or at least make a special handling for a "admin" Role that is allowed to do anything? "Installer" and "Owner" are not that important for me, as they are hardly used in reality currently.

PR #760, opened 2019-08-28
Require SymmetricEss only; improve calculation of maxEssCharge.

sfeilmeier:
ATTENTION: this change will require EvcsController to be executed before BalancingController; but therefor it will show the real possible value


sfeilmeier:
@sebastianasen please review + test if it works

PR #853, opened 2019-10-09
General Adjustments and Fixes

fabianfnc:
Fix -> CHP & EVCS Widget Modes
(*you couldnt see third mode on android devices)

Before:

After:

- EVCS Widget now shows correct Component Alias if given
- EVCS also show EnergySession value when charging process is finished
- Fix of colors in History (Energy + SOC Chart) -> they stay the same now
- Option to search for producttype in FEMS Overview
- General Code Cleanup


ui/src/app/edge/history/energy/energy.component.ts
```diff
@@ -61,6 +61,11 @@ export class EnergyComponent extends AbstractHistoryChart implements OnChanges {
 
   ngOnInit() {
     this.service.setCurrentComponent('', this.route);
+    this.service.getConfig().then(config => {
+      if (!config.hasProducer()) {
+        this.colors.splice(0, 1);
+      }
+    })
```
sfeilmeier: Wie besprochen, ist das nicht die optimale Lösung. Aber funktional und deshalb lassen wir es für's Erste.

PR #865, opened 2019-10-11
Added self consumption controller

venu-sagar:
self-consumption Optimization controller.


sfeilmeier:
Are these the latest changes?

venu-sagar:
Created another new one, so deleting this one

PR #909, opened 2019-10-24
Feature/selfconsumption dc charger

venu-sagar:
Added the Self-consumption controller and minimum power constraints for AC and DC.


venu-sagar:
Created another new one, so deleting this one

PR #911, opened 2019-10-24
Added self consumption controller and predictive delay charge control…

venu-sagar:
Added Self consumption controller. 
Added Predictive delay charge controller.
Made a slight change in the order of alias, id on Production, consumption predictor.
Made a minor change to remove a couple of warnings in BYD container(Asked Pooran before using it.)


io.openems.edge.controller.ess.predictivedelaycharge/src/io/openems/edge/controller/ess/predictivedelaycharge/ac/Config.java
```diff
@@ -0,0 +1,30 @@
+package io.openems.edge.controller.ess.predictivedelaycharge.ac;
+
+import org.osgi.service.metatype.annotations.AttributeDefinition;
+import org.osgi.service.metatype.annotations.ObjectClassDefinition;
+
+@ObjectClassDefinition( //
+		name = "Controller Predictive Delay Charge AC", //
+		description = "controller which calculates the power limitations every second to optimize the charging from PV")
```
sfeilmeier: Not quite well explaining what this Controller actually does and why there are two versions of it...

io.openems.edge.controller.ess.predictivedelaycharge/src/io/openems/edge/controller/ess/predictivedelaycharge/ac/Config.java
```diff
@@ -0,0 +1,30 @@
...
+	boolean enabled() default true;
+
+	@AttributeDefinition(name = "Ess-ID", description = "ID of Ess device.")
+	String ess_id() default "ess0";
+
+	@AttributeDefinition(name = "Grid-Meter-Id", description = "ID of the Grid-Meter.")
+	String meter_id() default "meter0";
+
+	@AttributeDefinition(name = "buffer hour", description = "number of hours before the final hour.")
+	int Buffer_hours() default 2;
```
sfeilmeier: Description is not sufficient to understand in my opinion. Also names of variables should be in camel-case starting with lower case

io.openems.edge.controller.symmetric.selfconsumption/readme.adoc
```diff
@@ -0,0 +1,5 @@
+= Symmetric Peak-Shaving
```
sfeilmeier: Readme is missing

io.openems.edge.controller.symmetric.selfconsumption/src/io/openems/edge/controller/symmetric/selfconsumption/SelfConsumption.java
```diff
@@ -0,0 +1,117 @@
...
+			int calculatedPower = this.gridPower + this.config.Maximum_Feed_In();
+
+			Power power = ess.getPower();
+			calculatedPower = power.fitValueIntoMinMaxPower(this.id(), ess, Phase.ALL, Pwr.ACTIVE, calculatedPower);
+
+			/*
+			 * set result
+			 */
+			ess.addPowerConstraintAndValidate("SelfConsmption-OptimizationController", Phase.ALL, Pwr.ACTIVE,
+					Relationship.EQUALS, calculatedPower); //
```
sfeilmeier: I don't think this is the best way to design this controller. I would expect you to only give a GREATER_OR_EQUALS contraint and letting the normal 'Balancing'-Controller do the rest. Then this controller would not be a 'SelfConsumption' controller but a 'AvoidMaximumFeedIn'-Controller. Otherwise you need to document why you need this controller additionally to the existing Balancing-Controller,

i.e. how does Service know, when they need to activate this Controller and when to activate traditional Balancing-Controller?

io.openems.edge.controller.ess.predictivedelaycharge/src/io/openems/edge/controller/ess/predictivedelaycharge/AbstractPredictiveDelayCharge.java
```diff
@@ -0,0 +1,204 @@
...
+		}
+
+		if (now.getHour() == 1 && executed) {
+			this.executed = false;
+		}
+
+		// target hour = 0 --> not enough production or Initial run(no values)
+		if (this.targetHour == 0) {
+			this.setChannels(State.TARGET_HOUR_NOT_CALCULATED, 0);
+			return;
```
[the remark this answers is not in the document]
venu-sagar: Well the whole point is that, buffer hour wont be calculated if the target hour is not calculated in the first place. So in my understanding it doesn't make any sense to show/check PASSED_BUFFER_HOUR.
But in the logic, due to the PASSED_BUFFER_HOUR condition, the system never checks for PASSED_Target_HOUR condition. so i changed it accordingly.

io.openems.edge.controller.symmetric.selfconsumption/src/io/openems/edge/controller/symmetric/selfconsumption/SelfConsumption.java
venu-sagar: well its my mistake. I had two repositories when i was working on this. one with KACO implementation(fems) and one with normal openems one. i was testing and making changes on one but didn't do this change in this repository and pushed it without checking.

PR #949, opened 2019-11-04
Introduce PID filter

sfeilmeier:
A proportional-integral-derivative (PID) filter uses three weighted variables to reach a process target. (see https://en.wikipedia.org/wiki/PID_controller). This pull-request introduces a PID filter functionality in OpenEMS Edge.

- `PidFilter` class resides in `io.openems.edge.common`, so it is available everywhere
- Controllers that control an energy storage system, can receive an instance of `PidFilter` via the `Power` component in `io.openems.edge.ess.core`.
- The P, I and D variables for `PidFilter` can be configured via the `PowerComponent` in `io.openems.edge.ess.core`.

This commit also comes with a few minor changes:
- `AbstractOpenemsComponent`s `addChannelsForProperties()` method can now handle a 'null' Context. This was required for unit tests.
- `LimitDischargeCellVoltage` tests now use the global `PowerDummy`. No need to implement it again.
- The Controllers `Balancing` and `PeakShaving` use the PID filter. They still have some log messages which should be useful while testing.
- The `maxPowerAdjustmentRate` config setting in the Controllers is obsolete.
- A Unit test in PeakShavingController shows the changed behaviour with PID

Next step: test this algorithm and find better initial values for P, I and D.


sfeilmeier:
Thanks. We should test and fine-tune the implementation on various FEMS devices before merging it, as it affects also Balancing.

Please test on your project; I'll try to roll it out on various others - fems7 is already working.

PR #963, opened 2019-11-11
Feature/active power and one full cycle

huseyinsaht:


sfeilmeier:
@huseyinsaht should we close this old pull-request or is there still anything to come?

PR #1041, opened 2020-01-27
Peak-Shaving within timeslot ("Hochlastzeitfenster")

pooran-c:
Time slot peak shaving controller, Which contains 
1) Peak Shaving within the timeslot.
2) Selection of specific days from the config.
3) Hysteresis Soc check inside the Timeslot and outside the time slot check.


pooran-c:
Tested with the simulator, 
Capture

PR #1048, opened 2020-02-05
Symmetric Peakshaving Widget

fabianfnc:
UI:
-add symmetric peakshaving widget

Flat Widget:
grafik

Advanced Widget:
grafik

Made with Angular Forms:
grafik

Restrictions:
-no leading zero
-no minus values
-zero as value is allowed as long:
upper value is bigger than lower value
but both zero as value is allowed


fabianfnc:
todo: add custom validators

PR #1069, opened 2020-03-06
Minor adjustments in the bcontrol meter component.

sebastianasen:
Added the invert function.
Fixed specific problems of the meter.


sfeilmeier:
Well done. Just a few things @venu-sagar, @sebastianasen:

- invertIfTrue-method should be called with "this"; or make it static
- Why does invertIfTrue return Integer and not int?
- Why is invertIfTrue public? Private should be sufficient.
- Please always delete the branch after merging a pull-request, to keep the repository clean.

PR #1077, opened 2020-03-12
GoodWe ET write implementation

venu-sagar:
Added apply power implementation
              - adjusted according to different power modes based on the availability of PV.
Added Allowed charge/Discharge based on Max apparent power
Added Energy value registers (Production, consumption, grid buy and grid sell)
          - a slight adjustment in the SUM implementation file (Energy values handler).


io.openems.edge.goodwe.et/src/io/openems/edge/goodwe/et/ess/Config.java
```diff
@@ -19,9 +19,15 @@
...
 	boolean enabled() default true;
 
+	@AttributeDefinition(name = "Read-Only mode", description = "Enables Read-Only mode")
+	boolean readOnlyMode() default false;
+
 	@AttributeDefinition(name = "Modbus Unit-id", description = "Unit-id")
 	int unit_id() default GoodWeEtConstants.DEFAULT_UNIT_ID;
 
+	@AttributeDefinition(name = "Power Mode", description = "Mode of the power")
+	PowerModeEms Powermode() default PowerModeEms.AUTO;
```
sfeilmeier, line 29: @venu-sagar I see that this config setting is not used anywhere. Also from the description it is not clear what it is supposed to be. Can we remove it?

PR #1087, opened 2020-03-25
Simulator + more

sfeilmeier:


sfeilmeier:
ToDos:
- Add Direct Datasource that can handle an array of integers - to avoid having to convert to a string
- Add a mode to merge Cycle-runs, so that it is possible to run Cycle-Times of e.g. 15 minutes length and still getting proper results e.g. for Balancing-Controller (problem is, that Controller writes are always one Cycle delayed due to the nature of the input-process-output cycle)

PR #1104, opened 2020-04-11
Update PAX Logging to 2.0.2

sfeilmeier:
Fixes an issue that caused an exception at startup which caused all components listed in 'runbundles' after org.ops4j.pax.logging.pax-logging-service to stop working.

The problem came up in OpenEMS Community Forum: https://community.openems.io/t/cannot-activate-template-controller/198/10. Thanks to Valentin and @clehne for helping to track it down and fix it.


zoernert:
might be related to https://github.com/OpenEMS/openems/issues/1296

PR #1120, opened 2020-04-27
BYD Battery-Box Commercial C130

huseyinsaht:


sfeilmeier:
@huseyinsaht Please go again through all the files, if they are really all for BYD Battery-Box Commercial (https://github.com/OpenEMS/openems/pull/1120/files) 

Do we really need to copy the Batteryhandling-Controller? @wgerbl

PR #1123, opened 2020-04-28
Sum-Component: add EssDischargePower-Channel

sfeilmeier:


sfeilmeier:
@fabianfnc: FYI for future developments. Finally no need to calculate 'effectivePower' anymore

PR #1141, opened 2020-05-12
Apparent Power automatic readable

venu-sagar:
Made the max Apparent power readable directly from the Modbus register instead of entering a fixed value. Makes it automatically adjustable to different inverter versions.

Removed the EMS power mode option from the config file which was of no use, and was there only for debugging purpose.

Reduced the tasks and added dummy registers to make it compact.


io.openems.edge.goodwe.et/src/io/openems/edge/goodwe/et/ess/GoodWeEtBatteryInverterImpl.java
```diff
@@ -488,16 +481,17 @@ private void updatechannels() {
 		 */
 
 		Integer soc = this.getSoc().value().get();
+		Integer maxApparentPower = this.getMaxApparentPower().value().get();
 
 		if (soc == null || soc >= 99) {
 			this.getAllowedCharge().setNextValue(0);
 		} else {
-			this.getAllowedCharge().setNextValue(GoodWeEtConstants.MAX_APPARENT_POWER * -1);
+			this.getAllowedCharge().setNextValue(maxApparentPower * -1);
```
sfeilmeier, line 489: You will get a NullPointerException here, as maxApparentPower could be null. You can use `TypeUtils.multiply()` to safely multiply.

io.openems.edge.goodwe.et/src/io/openems/edge/goodwe/et/ess/GoodWeEtBatteryInverterImpl.java
```diff
@@ -488,16 +481,17 @@ private void updatechannels() {
...
 		}
 		if (soc == null || soc <= 0) {
 			this.getAllowedDischarge().setNextValue(0);
 		} else {
-			this.getAllowedDischarge().setNextValue(GoodWeEtConstants.MAX_APPARENT_POWER);
+			this.getAllowedDischarge().setNextValue(maxApparentPower);
```
sfeilmeier, line 494: this is ok

io.openems.edge.goodwe.et/src/io/openems/edge/goodwe/et/ess/GoodWeEtBatteryInverterImpl.java
venu-sagar: Changed it!. Thanks for pointing out.

PR #1150, opened 2020-05-24
Correct references to build.adoc

HarmOtten:


sfeilmeier:
Thank you for the fix. Unfortunately changing the xref seems to break the Antora build at https://openems.github.io/openems.io/openems/latest/edge/deploy.html. I am yet to find out, how I can make it work in Github and in Antora...

sfeilmeier (approved):
Unfortunately changing the xref seems to still break the Antora build. Still fixing a typo, though.

PR #1152, opened 2020-05-25
feature/GoodWeEssCapacityConfigure

venu-sagar:
Added Ess capacity as a configurable field.


io.openems.edge.goodwe.et/src/io/openems/edge/goodwe/et/ess/Config.java
```diff
@@ -27,6 +27,9 @@
 
 	@AttributeDefinition(name = "Modbus-ID", description = "ID of Modbus brige.")
 	String modbus_id() default "modbus0";
+	
+	@AttributeDefinition(name = "Ess-Capacity", description = "capacity of the battery in Wh")
```
sfeilmeier, line 31: Call it "Capacity", we are already within an "Ess". Start description with capital letter.

io.openems.edge.goodwe.et/src/io/openems/edge/goodwe/et/ess/Config.java
```diff
@@ -27,6 +27,9 @@
...
+	int essCapacity() default 9000;
```
sfeilmeier, line 32: -> "capacity"
For readability reasons it's often a good idea to write "9_000".

sfeilmeier:
We use Git-Flow, so the branch name should have been something like "feature/GoodWeEssCapacityConfigure" (see e.g. https://blog.axosoft.com/gitflow/)

io.openems.edge.goodwe.et/src/io/openems/edge/goodwe/et/ess/Config.java
venu-sagar: Ok, I will correct it. 
I purposefully set it because there is a possibility to enter inverter capacity whoever is initialising it. If a person is referencing the Sems portal for values, then it is mentioned as just capacity in the portal, and it refers to Inverter capacity, not the battery.

io.openems.edge.goodwe.et/src/io/openems/edge/goodwe/et/ess/Config.java
venu-sagar: Noted and corrected.

sfeilmeier:
I was referring to the name of the branch, not the title of the Pull-Request. That was fine. Sorry for the confusion.

PR #1173, opened 2020-07-01
Controller Api MeteoControl

ahummelsberger:
I added a little controller that collects and sends specific channel values to Meteo Control via their VCOM v2.0 API.
In the Config you can choose a pvInverter (or just a symmetric production meter), a grid meter and an ess.
You can set a measurement interval and a transmition interval, according to the Meteo Control specifications.

The Controller queries the measurements via timedata and then sends them as xml data to the meteo control server.

Example:
When you set Measurement Interval to 900 and Transmition Interval to 60, every 60 mins (1h)  the controller queries timedata from the past 60 mins with a resolution of 900 secs (15 mins) puts them into the required XML body and sends it to the server.

This code has been tested with the OpenEMS simulators.


ahummelsberger:
> Thank you for the very interesting contribution. Before I go into a detailed review, I have two major doubts:
> 
> 1. Why is this a "Controller"? It is not doing any actual control, but just collecting and sending data, right? Maybe `io.openems.edge.meteocontrol` would be the right namespace and it should use EventHandlers to plug into the Cycle instead of Controller `run()` method. Note: Backend.Api.Controller is a Controller, because it actually has the ability to take remote control commands via the Backend.
> 2. Collecting and sending all channel data will potentially take a lot of time, so this should be done in a separate thread. Otherwise it will block the entire Cycle, if there are a lot of Channels or there is a problem with the remote server or web connection.
> 
> Can you refactor the code accordingly or do you need more details/information?

Actually I just used a bnd template that looked to fit the most.
Sure in this case this doesn't need to be a controller.

I thought because Controllers implement the run() method there wouldn't be a blocking issue.
But I understand it now.

I can make a refactoring in 2 - 3 weeks.

sfeilmeier:
Ok. Just for clarification and context: the Controller `run()` method is call synchronuously in every Cycle. This is required to have Controllers with descending priorities being executed one after each other:

https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.core/src/io/openems/edge/core/cycle/CycleWorker.java#L106

sfeilmeier:
Closing because of inactivity.

@ahummelsberger if you are still working on this feature, we can reopen this pull-request at any time.

PR #1177, opened 2020-07-10
Implemented (Still working) on the resend historic data.

venu-sagar:
Implemented (Still working) on the resending of historic data.

- Backend API - Implemented my main logic here in the method resendHistoricData.


io.openems.edge.controller.api.backend/src/io/openems/edge/controller/api/backend/BackendApi.java
```diff
@@ -149,6 +170,17 @@ protected void deactivate() {
...
 	@Override
 	public void run() throws OpenemsNamedException {
+
+		BooleanReadChannel connectionStatus = this.channel(ChannelId.BACKEND_CONNECTED);
+
+		log.info("Backend Connected Channel: " + connectionStatus.getNextValue().get());
+
+		if (connectionStatus.getNextValue().get() != null) {
+			if (connectionStatus.getNextValue().get() == true) {
+				this.resendHistoricData();
```
sfeilmeier, line 180: If you do it like this, the whole collecting and sending process is happening synchronously with the Cycle. This is a bit dangerous, because it can block the Cycle. This is why for the normal collection process I use the separate `BackendWorker`

sfeilmeier:
Development on this topic will be continued in another branch.

PR #1185, opened 2020-07-24
Improve RRD4j implementation

sfeilmeier:
- fix local timezone handling
- use rrd4j db pool
- always close() RrdDb files; otherwise the pool fills up and no more files can be opened
- apply coding best practices
  - apply Checkstyle
  - split interface and impl

### DIFF
--- a/io.openems.edge.timedata.rrd4j/src/io/openems/edge/timedata/rrd4j/RecordWorker.java
+++ b/io.openems.edge.timedata.rrd4j/src/io/openems/edge/timedata/rrd4j/RecordWorker.java
@@ -1,7 +1,8 @@
 package io.openems.edge.timedata.rrd4j;
 
+import java.io.IOException;
+import java.time.Instant;
 import java.time.LocalDateTime;
-import java.time.ZoneOffset;
 import java.time.temporal.ChronoUnit;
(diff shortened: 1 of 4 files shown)


io.openems.edge.timedata.rrd4j/src/io/openems/edge/timedata/rrd4j/RecordWorker.java
```diff
@@ -75,11 +78,10 @@ public void collectData() {
 
 		// Same second as last run? -> RRD4j can only handle one sample per second per
 		// database. Timestamps are all stored "truncated to seconds".
-		if (recordTimestamp.isEqual(this.lastRecordedTimestamp)) {
+		if (timestamp.equals(this.lastTimestamp)) {
```
ahummelsberger, line 81: lastTimestamp is never reassigned, so this would always be false.

sfeilmeier: Thanks for pointing out such an error again... fixed here: https://github.com/OpenEMS/openems/commit/7e758783ec9acbabefeec3a8a8d339471393369b

PR #1190, opened 2020-07-29
Fix cookie HTTP header field name being case sensitive

fgarlin:
HTTP header field names shouldn't be case sensitive as per [specification](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2).

This is not the most elegant solution, but it will fix errors with APIs that force using lowercase characters. If someone can think of more elegant approaches, feel free to comment.

### DIFF
--- a/io.openems.common/src/io/openems/common/utils/JsonUtils.java
+++ b/io.openems.common/src/io/openems/common/utils/JsonUtils.java
@@ -365,6 +365,14 @@ public class JsonUtils {
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


sfeilmeier:
Thank you @fgarlin for pointing that out and contributing! I refactored your code to allow all variants of 'cookie' and added thorough JUnit tests.

fgarlin:
Thanks, that's definitely a cleaner solution!

PR #1208, opened 2020-08-18
Implement OPC UA Api Controller

sfeilmeier:
Designed as per requirement of the MEMAP project: 
- https://openems.github.io/openems.io/openems/latest/randd.html#_memap
- https://www.fortiss.org/forschung/projekte/detail/memap


sfeilmeier:
Reading of Channel values in UaExpert works now:

grafik

PR #1211, opened 2020-08-24
UI: subscribe/general optimization + get edge id in modalview

fabianfnc:
- allow modal components to get edge id via setcurrentcomponent method
- fix of evcs number of phases calculation
- fix of not unsubscribing from evcs cluster
- general optimization of ngOnInit in widgets


sebastianasen (approved):
Well done.

PR #1214, opened 2020-08-28
Changed LocalDateTime to ZonedDateTime to avoid time zone problems in prediction service.

venu-sagar:
Changed the LocalDateTime implementation to ZonedDateTime for adjusting to the UTC time zone.

Added a configurable component to show the prediction values (Prediction and Consumption) for debugging.

Removed a bug which was causing while retrieving the energy values from RRD4J.


io.openems.edge.controller.ess.predictivedelaycharge/src/io/openems/edge/controller/ess/predictivedelaycharge/AbstractPredictiveDelayCharge.java
```diff
@@ -119,12 +123,15 @@ public Integer getCalculatedPower(ManagedSymmetricEss ess, ProductionHourlyPredi
...
 		 */
 		if (now.getHour() == 0 && !this.isTargetHourCalculated) {
-			HourlyPrediction productionHourlyPredition = productionHourlyPredictor.get24hPrediction();
-			Integer[] hourlyProduction = productionHourlyPredition.getValues();
+			Integer[] hourlyProduction = productionHourlyPredictor.get24hPrediction().getValues();
 			Integer[] hourlyConsumption = consumptionHourlyPredictor.get24hPrediction().getValues();
 
 			// Start hour of the predicted values
-			LocalDateTime predictionStartHour = productionHourlyPredition.getStart();
+			ZonedDateTime predictionStartHour = productionHourlyPredictor.get24hPrediction().getStart();
```
sfeilmeier, line 130: Can we assume, that `get24hPrediction()` will always be a fast method? I assume, that e.g. if we were using some ANN, calling this method might be more expensive - and then we should not call it twice here. Maybe line 122 was not a bad idea than.

sfeilmeier (approved):

PR #1222, opened 2020-09-07
Byd Commercial C130 Start Stop Implementation

huseyinsaht:


io.openems.edge.common/src/io/openems/edge/common/component/OpenemsComponent.java
```diff
@@ -257,6 +257,20 @@ public default boolean hasFaults() {
...
+	/**
+	 * Does this OpenEMS Component report any Warnings?
+	 * 
+	 * <p>
+	 * Evaluates all {@link StateChannel}s and returns true if any Channel with
+	 * {@link Level#WARNING} is set.
+	 * 
+	 * @return true if there is a Warning.
+	 */
+	public default boolean hasWarnings() {
```
sfeilmeier: The definition of this method is unclear. Does `hasWarnings` mean, that there are exactly WARNING channels set - or does it mean, that WARNING or FAULT channels are set? In that regard your Javadoc text does not fit with the actual implemented logic using `Level.isAtLeast()`.

As this method is not used anyway, I am dropping it from the pull-request.

sfeilmeier (approved):
Well done!

PR #1235, opened 2020-09-21
Added state channels for all possible mistakes in the Dip-Switch settings

sebastianasen:
Added the state channels and set them according to the read data of the KEBA hardware settings (Dip-Switches).

Examples:
grafik
grafik


sfeilmeier (approved):
Erfüllt den Zweck...

PR #1236, opened 2020-09-21
fixdigitaloutput modal.component.ts returns a boolean value intead of…

clehne:
Bugfix: 
The UI fixdigitaloutput component now sends a boolean value instead of a string. 

Reason: 
The edge fixdigitaloutput component has a boolean property "isOn". When this property is changed using the UI, then the edge property configuration was changed also (from "boolean" to "string"). This completely breaks the influx connection, because it could not write to the database anymore. It will fail with logoutput like this:
`Unable to write to InfluxDB: partial write: field type conflict: input field "ctrlIoFixDigitalOutput0/_PropertyIsOn" on measurement "data" is type string, already exists as type integer `


fabianfnc:
is fixed with https://github.com/OpenEMS/openems/pull/1256

PR #1237, opened 2020-09-21
RevPiDigitalIoDevice modified

clehne:
- added DEBUG_OUT channels to see DataOut write channel content
- channels are now updated depending on the standard cycle process state
- write channels now behave like write channels from other DigitalOut modules, so that they can be used with standard IO controllers

### DIFF
--- a/io.openems.edge.io.revpi/src/io/openems/edge/io/revpi/RevPiDigitalIoDevice.java
+++ b/io.openems.edge.io.revpi/src/io/openems/edge/io/revpi/RevPiDigitalIoDevice.java
@@ -1,6 +1,7 @@
 package io.openems.edge.io.revpi;
 
 import java.io.IOException;
+import java.util.Objects;
 import java.util.Optional;
 
 import org.clehne.revpi.dataio.DataInOut;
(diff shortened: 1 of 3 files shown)


io.openems.edge.io.revpi/src/io/openems/edge/io/revpi/RevPiDigitalIoDevice.java
```diff
@@ -164,44 +215,68 @@ public void handleEvent(Event event) {
...
+			} else {
+				b.append("0");
+			}
+		} else {
+			b.append("-");
+		}
+	}
+
 	@Override
 	public String debugLog() {
```
sfeilmeier, line 261: In the KMTronic relay board implementation, we use a format like "xxx--xx" which i find easier to read for non-techies than "1110011". See https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.io.kmtronic/src/io/openems/edge/io/kmtronic/KmtronicRelayOutput.java#L116 - but it's up to you.

sfeilmeier (approved):
Thanks for the improvements. You may or may not consider my changes and then "squash & merge" yourself.

Regards, Stefan

sfeilmeier:
Closing because of inactivity.

@clehne if you are still working on this feature, we can reopen this pull-request at any time.

PR #1256, opened 2020-10-16
Add/Edit Component + Widget: Value Parsing Quickfix

fabianfnc:


sfeilmeier:
@fabianfnc , @sebastianasen

I reset type related changes to develop, as the approach was not ideal. The only standard way to get properly typed values out of Angular FormControls is by setting the `type` property on the `ion-input` field.

I do this in a generic, automatic way for Component Install and Update - i.e. there should have never been a problem there. See relevant code:

- Generate the schema for fields on Edge, including the field type
https://github.com/OpenEMS/openems/blob/develop/io.openems.common/src/io/openems/common/types/EdgeConfig.java#L544
- Use the schema with Formly
https://github.com/OpenEMS/openems/blob/develop/ui/src/app/edge/settings/component/update/update.component.ts#L61
- Apply the Schema on the ion-input field
https://github.com/OpenEMS/openems/blob/develop/ui/src/app/edge/settings/component/shared/input.html#L2

It might be an option for the future to just use Formly and this strong typing also for the UI widgets. Till then the easiest way will be to just update all the ion-input fields manually, i.e.
```
<ion-input class="regularFont" [value]="formGroup.value.peakShavingPower"
    formControlName="peakShavingPower" style="text-align: end;">
</ion-input>
```
https://github.com/OpenEMS/openems/blob/develop/ui/src/app/edge/live/peakshaving/asymmetric/modal/modal.component.html#L77-L79
to 
```
<ion-input class="regularFont" formControlName="peakShavingPower" style="text-align: end;" type="number">
</ion-input>
```
(by the way `[value]="formGroup.value.peakShavingPower"` is never required; this is what Angular Forms does for us automatically; see https://angular.io/guide/reactive-forms#adding-a-basic-form-control: "Using the template binding syntax, the form control is now registered to the name input element in the template. The form control and DOM element communicate with each other: _the view reflects changes in the model, and the model reflects changes in the view_.")

@fabianfnc: Can you do those changes? You know all the places where this would have to be change the best.

sfeilmeier (changes requested):
See comment above

fabianfnc:
@sfeilmeier Widgets passen jetzt.
Add/Edit Components muss noch angepasst werden, gut testbar mit dem asymmetric peakshaving controller

sfeilmeier (approved):

Issue #1197, opened 2020-08-06
EvcsController unable to set certain channel values from child evcs component


marcelheim:
### Bug Report or Feature Request (mark with an `x`)
```
- [x] bug report -> please search issues before submitting
- [ ] feature request
```

### Bug description or desired functionality.

Version: 2020.16.0 (Release)
Afftected component: io.openems.edge.controller.evcs
Source: [EvcsController.java](../blob/master/io.openems.edge.controller.evcs/src/io/openems/edge/controller/evcs/EvcsController.java) after commit openems@641d048
#### Description
Channels SET_ENERGY_LIMIT, SET_CHARGE_POWER_REQUEST, SET_CHARGE_POWER_LIMIT aren't set properly by methods setEnergyLimit(), setChargePowerRequest() and setChargePowerLimit() in used evcs component
 
#### Example code from [EvcsController.java](../blob/master/io.openems.edge.controller.evcs/src/io/openems/edge/controller/evcs/EvcsController.java)
```java
if (!config.enabledCharging()) {
	evcs.setChargePowerLimit(0);
	return;
}
```

#### Possible fix
Call internal set funtion
```java
if (!config.enabledCharging()) {
	evcs._setChargePowerLimit(0);
	return;
}
```

sebastianasen:
Hallo @marcelheim,
thank you for creating this issue.

Do you use the EvcsController in combination with an EvcsCluster?
Only here, i found the bug that an old requested charge power is still preset.

Because of that, the cluster is setting also the setChargePowerLimit (to the old value).
This will be fixed in the next pull request wich i will mention in here.

---
For your information:
The _setChargePowerLimit will set the "value" of the Channel "SetChargePowerLimit". This value is only used for displaying the current target we sent.

The setChargePowerLimit will set per defenition the "write value" of the Channel. 
In every Chargingstation i will react on the WriteValue and will reset it afterwards. Only that Value will be forwarded to the chargingstation.

PR #1263, opened 2020-10-22
Further development and rework of the EVCS (Electric Vehicle Charging Station) components

sebastianasen:
- Integrate a ramp filter, used primary by EVCS's, instead of using a PID Filter for the charge power:
  - Sent charge power will increase steadily, by a configured rate if the singleton component EVCS Power is activated

- Wrote new tests for EvcsController and EvcsCluster

- Substantial changes in the EvcsCluster:
  - Changed code for minimum guaranteed power
  - Changed the handling of the maximum Power of an EVCS. It will use that value of the maximum for calculation, but the sent 
     charge power will be like before
     (Some cars are reacting like this:    setPower: 22 kW -> chargedPower: 11kW,  setPower: 11kW -> chargedPower 7 kW, ... )
  -  Added logic, to reduce the value calculated from the "maximum allowed discharge power" of the ess, to avoid certain jumps to  
     zero, if the essSoc gets lower. In the future, this should be handled by each battery itself.

- General changes in the energy calculation of OCPP EVCS's:
   - Depending on the charging station, it returns only the energy value of this charging session or the total session.
      Now we are calculating the always the other value, depending on the given

- Fixed Issue #1197 (Send also a chargePowerRequest of 0 if charging is disabled)

- Small bug fixes


io.openems.edge.evcs.api/src/io/openems/edge/evcs/api/ManagedEvcs.java
```diff
@@ -51,6 +55,37 @@
...
+		 * Applies the PID filter and then sets a fixed Active Power.
+		 * 
+		 * <ul>
+		 * <li>Interface: ManagedEvcs
+		 * <li>Writable
+		 * <li>Type: Integer
+		 * <li>Unit: W
+		 * </ul>
+		 */
+		SET_CHARGE_POWER_LIMIT_WITH_PID(new IntegerDoc() //
```
sfeilmeier, line 68: This is not actually a PID filter, so it's a bit misleading.

sfeilmeier (approved):
Well done. It's not perfect, but a big improvement to 'develop'. Thanks!

PR #1265, opened 2020-10-23
add french translation to UI

denis-jidovtseff-jc:


fabianfnc (changes requested):
please add FR language to browserLangToLangTag method in
src/app/shared/service/service.ts

after that change its good to go

fabianfnc (approved):

PR #1266, opened 2020-10-24
FENECON Mini: add write-mode for remote-control

sfeilmeier:


sfeilmeier:
Works well on one test-system, but I was not able to properly control another test system.

PR #1269, opened 2020-10-28
Implemented QueryHistoryEnergyPerPEriod in rrd4j

ahummelsberger:


sfeilmeier (approved):
Thank you Andreas! I am wondering if it would be possible to have a less performance-intensive implementation, as this one is causing a lot of individual requests. But it is certainly good to go for now. Thanks!

ahummelsberger:
> Thank you Andreas! I am wondering if it would be possible to have a less performance-intensive implementation, as this one is causing a lot of individual requests. But it is certainly good to go for now. Thanks!

You're welcome!
I first tried to use a simple fetch statement. But there are problems with the required resolution of the data.
I might find a more efficient way.

PR #1272, opened 2020-10-29
Make default Api ports configurable

denis-jasselette-jc:
It would be nice to be able to override those hard-coded ports. Especially because opening port 502 requires me to be connected as root, which is not ideal (https://www.staldal.nu/tech/2007/10/31/why-can-only-root-listen-to-ports-below-1024/).

I included a change to the naming of the existing parameter openems.data.dir to conform to the namespace conventions: io.openems.*. I am open to change it back if we don't want to break backward compatibility.

### DIFF
--- a/io.openems.common/src/io/openems/common/OpenemsConstants.java
+++ b/io.openems.common/src/io/openems/common/OpenemsConstants.java
@@ -108,6 +108,8 @@ public class OpenemsConstants {
 	public final static String PROPERTY_LAST_CHANGE_AT = "_lastChangeAt";
 
 	public final static String OPENEMS_DATA_DIR_KEY = "io.openems.data.dir";
+	public final static String API_REST_PORT_KEY = "io.openems.api.rest.port";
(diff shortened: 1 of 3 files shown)


sfeilmeier:
Thank you for the pull-request. I am not sure if this addition is adding more confusion. If we add it, we would have at least to document it properly, I suppose, because it only applies on first run, when no Modbus/TCP and REST/Api services have been configured.

I understand the problem with port 502 on Linux systems though. The `DefaultConfigurationWorker` is anyway more meant to support provisioning of new and updating of existing systems.

Do you think we could just disable the automated setup of those services instead?

denis-jasselette-jc:
You are probably right about this over-complicating things. However, I think that using port 502 is not an ideal implicit behavior.

What is the use case for which this was introduced anyway? If one requires to use an API, isn't it straightforward to configure the corresponding component? Was it intended to streamline the setup of the UI or something? Sorry if the answer is obvious, I am still learning about OpenEMS. Unless there are good reasons why this is useful, I would rather, as you said, disable the automated setup and let the user explicitly configure the components.

PR #1278, opened 2020-11-02
Add JUnit tests for Modbus protocol definitions

sfeilmeier:


clehne:
In https://github.com/OpenEMS/openems/pull/1278/commits/325dc9f04b23f80d5755785524e022c397a19f36 was a change in the activation method in io.openems.common/resources/templates/device-modbus/$srcDir$/$basePackageDir$/MyModbusDevice.java

In line 63-67 there is a return call, if super.activate() returns true. This doesn't make any sense for me, because, the component is still activated, but then this.config == null. Are you sure, that this is the wished behaviour?

sfeilmeier:
If you only look at these lines you are right.
```
		if(super.activate(context, config.id(), config.alias(), config.enabled(), config.modbusUnitId(), this.cm, "Modbus",
				config.modbus_id())) {
			return;
		}
		this.config = config;
```
https://github.com/OpenEMS/openems/blob/develop/io.openems.common/resources/templates/device-modbus/$srcDir$/$basePackageDir$/MyModbusDevice.java#L64-L68

But OSGi is doing some more things in background. super.activate() only returns true if it was updating the configuration of the Component: 
https://github.com/OpenEMS/openems/blob/a699fc5f1e50ba78a826f27c0e5a08dfd8ef519e/io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/api/AbstractOpenemsModbusComponent.java#L102-L103 

When that happens, OSGi is destroying the Component immediately (calling `deactivate()`) and recreating it (calling `activate()`) again. Yes, arguable it would be not wrong to assign config before the super-call; I just wanted to show/avoid, that any real business logic gets executed here

clehne:
Thank you. I missed that OSGI behaviour. Never too late to learn something.

PR #1286, opened 2020-11-05
kaco92_and_sum_update

pooran-c:
1. kaco 92
 - added a activate channel in config.
 - added the check for 92 model reading
2. _sum
 - added the allowed charge and discharge power on sumImpl
 - and Reactive power channel


io.openems.edge.ess.cluster/src/io/openems/edge/ess/cluster/EssCluster.java
```diff
@@ -156,6 +161,20 @@ private void calculateChannelValues() {
 			}
 		}
 
+		// Setting the allowed charge and discharge power
+		for (String essId : this.config.ess_ids()) {
+			ManagedSymmetricEss ess;
+			try {
+				ess = this.componentManager.getComponent(essId);
```
sfeilmeier: Careful: If ess is not a `ManagedSymmetricEss` this will give a ClassCastException. I'd rather add it to the above for-loop.

sfeilmeier (approved):
Very well done! I just added one little remark to be careful for.

PR #1289, opened 2020-11-09
Controller for a SG-Ready Heat Pump

sebastianasen:
Implemented a controller for all heat pumps that are using the SG-Ready standard i.e. have the SG-Ready-Label.
See also https://www.waermepumpe.de/normen-technik/sg-ready/

The heat pump is controlled via two Relay contacts, according to the grid active power and the user settings.


sfeilmeier (approved):
Very well done!

PR #1298, opened 2020-11-14
OpenEMS Backend as Docker container with InfluxDB for TimeDB preconfi…

zoernert:
Adds Dockerfile to allow running a container just with backend and preconfigured influxdb
Quick Run/Test via:
`docker run --name openems-backend --net='host' -d stromdao/openems-backend`

TODO: Might be added to cradlew to allow building docker image from source. Requires to copy `openems-backend.jar` into `/tools/docker/openems-backend/build`


tools/docker/openems-backend/.atom-build.yml
```diff
@@ -0,0 +1 @@
+cmd: "docker build -t stromdao/openems-backend .;docker push stromdao/openems-backend"
```
sfeilmeier, line 1: This seems to be stromdao specific. Can we remove this file and instead add the command as a hint to the readme - like:

## Build
`docker build -t stromdao/openems-backend .; docker push stromdao/openems-backend`

sfeilmeier (approved):
Thanks for the contribution. I wanted to do these little changes myself, but didn't have write permissions. See my comments - than you can squash & merge it.

PR #1335, opened 2020-12-16
Ems timeout from sinexcel

pooran-c:
-  Added the EMS time out.
-  Added the BMS Time out.
- Added the Docs


sfeilmeier (changes requested):
I fixed the Java style variable namings.

Three things I do not like:
1. Why do we need to set this register on every cycle? This adds one - mostly unnecessary - modbus request on every cycle. Wouldn't it be sufficient to compare and only set if is not correct?
2. Only logging an error to the console just hides the error away if you have 100s of systems.
3. Whenever you log something, use the logInfo/logWarning/logError methods, as they also print the Component-ID. Otherwise you can, in a multi-inverter scenario, never identify which instance threw the error

sfeilmeier:
@pooran-c Is this Pull-Request still valid or can we close it?

PR #1336, opened 2020-12-16
on/off grid state machine for the new Box

pooran-c:
On/ Off-grid state machine for the new box.


sfeilmeier:
Superseded by #1637

Issue #1351, opened 2021-01-05
[API] Enable multiple-at-once channels readings as a JSON Array


martingruening:
When you frequently read a lot of channel over the REST API you may flood the server with a lot of HTTP GET requests in a short time frame. Not only that you easily exceed connection limits, this is also expensive in terms of resources needed to handle.

**Improvement request:**
Instead implement RegExp matching in the channel address to get multiple values with a single request (i.e. "_sum/Active.*Power"). In those cases return a JSON array instead of single JSON object.

While the single requests takes a little bit longer than the request for an exact match, this can reduce the number of required API requests drastically.

martingruening:
Implemented this, will include in the next PR

Example - Request:
`GET http://kube:8084/rest/channel/.*/Active.*Power`

Response:
```
[
    {
        "address": "pvInverter0/ActivePower",
        "type": "INTEGER",
        "accessMode": "RO",
        "text": "Negative values for Consumption; positive for Production",
        "unit": "W",
        "value": 90
    },
    {
        "address": "meter0/ActiveProductionPower",
        "type": "INTEGER",
        "accessMode": "RO",
        "text": "",
        "unit": "W",
        "value": 465
    },
    {
        "address": "meter0/ActivePower",
        "type": "INTEGER",
        "accessMode": "RO",
        "text": "Negative values for Consumption; positive for Production",
        "unit": "W",
        "value": 465
    },
    {
        "address": "meter0/ActiveConsumptionPower",
        "type": "INTEGER",
        "accessMode": "RO",
        "text": "",
        "unit": "W",
        "value": 0
    }
]
```

Issue #1350, opened 2021-01-05
[API] Make connection limit configurable


martingruening:
Make connection limit for REST API configurable, default is hardcoded at 5 (which is to low if you really use the API)
Hit the limit already when I retrieved only  a handful of channels every 30 seconds. Should be configurable.

martingruening:
Code change done, will merge as part of next PR

Issue #1348, opened 2020-12-30
[API] Insert Unit into channel output


martingruening:
Feature Request:
When reading a channel through the REST API also provide the unit of the channel.

Example - instead of
```
{
    "value": 457,
    "type": "integer",
    "writable": false
}
```

send
```
{
    "value": 457,
    "unit": "W",
    "type": "integer",
    "writable": false
}
```

martingruening:
You can assign this to me. Done the easy modfication already. PR will follow bundled with other improvements in RestHandler.

sfeilmeier:
A quick thought on this: if I would implement it, I would try to reuse the Channel representation in `EdgeConfig`. EdgeConfig is generally used to represent the entire configuration of an OpenEMS Edge and is used in OpenEMS UI to build the interface. See:

https://github.com/OpenEMS/openems/blob/develop/io.openems.common/src/io/openems/common/types/EdgeConfig.java#L236

martingruening:
Good suggestion. Will do that. Shall I keep the existing "writable" property in the API code? This would make sure that we don't break code depending on the API, but then the writable property seems to be redundant with the "accessMode" property of the config.

sfeilmeier:
I believe dropping the "writable" property is ok.

Issue #1347, opened 2020-12-29
[API] Auth error in HTTP request to API should trigger a 401 response code (not 400)


martingruening:
If the authentication fails during an HTTP request to the API it should return a 401 (Unauthorized)
[https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401)

Why is that important?
First: it is the correct response for this situation.

Second: the popular jetty HTTP client library will first issue a request WITHOUT credentials (even when configured with credentials) and only if it receives a 401 response will redo the request with credentials. That means as is you can't query the API with a jetty client (like used in the OpenHAB http binding). The reason is then quite hard to find (took me some hours).

martingruening:
Already fixed, this can be assigned to me. I will submit a PR for RestHandler that includes this change.

sfeilmeier:
Thank you! Right now I do have some other priorities, but will be back on track soon.

PR #1352, opened 2021-01-05
Several improvements in the REST API

martingruening:
- Made connection limit configurable
- Made channel output in GET requests more comprehensive and inline with EdgeConfig
- Deliver the approriate response code for authentication errors
- Enable multiple-at-once channel readings as a JSON Array

### DIFF
--- a/io.openems.edge.controller.api.rest/readme.adoc
+++ b/io.openems.edge.controller.api.rest/readme.adoc
@@ -27,9 +27,52 @@ Use a HTTP request with method `GET` to read the current value of a Channel.
 *Example:* To read the current state of charge of the battery, send a GET request to http://x:user@localhost:8084/rest/channel/_sum/EssSoC. It returns a response like:
 ```
 {
-  "value": 50
+  "type":"INTEGER",
+  "accessMode":"RO",
+  "text":"",
+  "unit":"%",
+  "value":50
 }
 ``` 
+
+The `GET` api also understands regular expressions. Send a GET request to http://x:user@localhost:8084/rest/channel/.*/Active.*Power to read all `ActivePower` and `ReactivePower` channels of all components. It returns a response like:
+
+```
+[
+  {
+    "address":"pvInverter0/ActivePower",
+    "type":"INTEGER",
+    "accessMode":"RO",
+    "text":"Negative values for Consumption; positive for Production",
+    "unit":"W",
+    "value":90
+  },
+  {
+    "address":"meter0/ActiveProductionPower",
+    "type":"INTEGER",
+    "accessMode":"RO",
+    "text":"",
+    "unit":"W",
+    "value":465
+  },
+  {
+    "address":"meter0/ActivePower",
+    "type":"INTEGER",
+    "accessMode":"RO",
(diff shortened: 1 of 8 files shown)


martingruening:
Solves issues: #1347 #1348 #1350 #1351

martingruening:
Corrected scale factors for Voltage, Current and Frequency in Sunny Home Manager 2.0 Meter implementation

sfeilmeier (approved):
Thanks Martin, these are some very good improvements. I have no comments or improvement suggestions on the code quality.

Just want to point out two things:
1. I personally like Regex a lot, but I also feel that many people have problems understanding them. Because of this I am introducing a `matchWildcard` method in my "Implement a Prediction architecture" branch (#1327), that uses well known 'asterisk' approach. It's by far not as flexible as Regex, but I wonder if it could not be an option here. https://github.com/OpenEMS/openems/blob/724c20af2c37a98893961394e9402b9c11cc0be8/io.openems.common/src/io/openems/common/utils/StringUtils.java#L46
2. Have you heard about GraphQL? It is a query language for REST calls that follows the same idea of yours - to reduce calls by joining them. Do you think going that way would be a better approach? (https://graphql.org/)

martingruening:
Thanks for the review and your remarks.

Regarding RegEx: I believe that anybody that can call REST APIs and parse JSON can understand RegExp to a certain degree. And if you only know that `.*` is a wildcard, that can be enough to get a multiple-in-one query quickly going. So I am quite happy with the RegExp approach - it gives power to people who can create complex RegExp and give simplicity to those who just know the simplest RegExp wildcard string.

I have never heard of GraphQL. My initial thought was that it might be overengineering for my use case, but I might be wrong.  I recently started to build an OpenEMS binding for OpenHAB. Initially I wanted to use the REST API to frequently poll the channels, but now I'm investigating the JSON-RPC Websocket as a better way for integration.

sfeilmeier:
Ok - yes, I agree. It's good to have your opinion on that. I might also rethink the approach in the other branch. GraphQL sounded to be a very generic approach on solving these kind of problems. But I am also not sure how widely used it is and if it is worth the effort.

Ping me if you need support on the JSON-RPC topic. There is some documentation here: https://openems.github.io/openems.io/openems/latest/component-communication/index.html#_subscribe_to_channels

You have to send a JSON-RPC "SubscribeChannelsRequest" and from then on you will receive "CurrentDataNotifications". See the UI for an example:

- Send the request:
https://github.com/OpenEMS/openems/blob/develop/ui/src/app/shared/edge/edge.ts#L149-L154

- Definition of the request:
https://github.com/OpenEMS/openems/blob/e19f80d01148e336eec6cfc4985198e98960f69a/ui/src/app/shared/jsonrpc/request/subscribeChannelsRequest.ts#L10-L18

- Handling of the CurrentDataNotifications:
https://github.com/OpenEMS/openems/blob/develop/ui/src/app/shared/edge/edge.ts#L175

- Definition of the notification:
https://github.com/OpenEMS/openems/blob/e19f80d01148e336eec6cfc4985198e98960f69a/ui/src/app/shared/jsonrpc/notification/currentDataNotification.ts#L8-L14

PR #1377, opened 2021-01-28
Remove default configurations for JSON/REST and Modbus/TCP Api

sfeilmeier:


sfeilmeier:
This is a followup of #1272

PR #1383, opened 2021-02-01
wrong datatypes and scale factors @ BMW modbus implementation changed

JohannesHundschell:
- Register 1006 MAXIMUM_OPERATING_CURRENT (sign)
- Register 1010 DISCHARGE_MAX_CURRENT (sign)
- Register 1029 CURRENT (sign and scale factor)
- Register 1030 AVERAGE_TEMPERATURE (sign)
- Register 1031 MIN_CELL_TEMPERATURE (sign)
- Register 1032 MAX_CELL_TEMPERATURE (sign)
- Register 1037 INSULATION_RESISTANCE (scale factor)
- Register 1041 MAXIMUM_LIMIT_DYNAMIC_CURRENT_HIGH_RES (sign)
- Register 1042 MINIMUM_LIMIT_DYNAMIC_CURRENT_HIGH_RES (sign)

PR #1387, opened 2021-02-10
Missed "throws exception"

pooran-c:
Added the throws exception.


sfeilmeier:
Fixes faults introduced in #1383

sfeilmeier (approved):

PR #1391, opened 2021-02-17
Feature/wago analog input

Nanunanuk:
Hi,
we added the functionality to read in analog signals with a wago fieldbus coupler.
Version is running but throws an error at startup. Something like wrong starting adress 512. I am not so familiar with the Modbus registers.
Another thing that needs to be changed is that each analog module will need a different calibration factor. That can be implemented in the config file for the different modules. But what if two modules of the same type with different factors are included?

Nils Reiners (Fraunhofer ISE)


sfeilmeier:
Hi Nils,

Thank your for the contribution. I just reset some Eclipse IDE files to develop so the [Files Changed](https://github.com/OpenEMS/openems/pull/1391/files) shows only the actual changes.

Can you post the exact exception message about wrong starting adress 512? This might point us in the right direction. I remember that some registers start at 0 and some start at 512 for WAGO fieldbus.

The calibration factor is an interesting problem. Due to the dynamic implementation of the fieldbus it is indeed difficult to do. What is the use-case for calibration factor? Is it used e.g. to derive an actual temperature value from an analog output? In that case maybe it is not even the task of the WAGO fieldbus component, but rather of a higher level component that uses this value.

Regards, Stefan

PR #1393, opened 2021-02-18
Patch05 evcs cluster

clehne:
The event handler of the EvcsClusterSelfConsumption is not called due to the missing event interface. 
With this patch Channels are updated now, but I haven't checked full functionality of the controller.


sebastianasen (approved):
Thanks for your improvements.

PR #1395, opened 2021-02-19
Implement MQTT Communication

sfeilmeier:
This could eventually replace/supersede the current WebSocket-based communication between UI, Backend and Edge.

First priority is to replace communication between Edge and Backend.


sfeilmeier:
Note: see if we can join efforts with https://github.com/ConsolinnoEnergy/openems/tree/feature/MQTT/io.openems.edge.bridge.mqtt by @DerStoecki

martingruening:
Any approximation when this will be merged? I am happy to test/improve it.

sfeilmeier:
Hi Martin, Thanks for the reminder. I still need to see if and how this approach can be merged with what Consolinno did. Yet... maybe it would still be good to just try to finalize one implementation and see if we need to break/refactor it later.

The initial reason for implementing this MQTT controller was replacing the way Edge communicates with Backend, but that was postponed for now, because this proved to not be the bottleneck that I thought it would be.

martingruening:
I see your point. But from an ecoysystem point of view it way easier to integrate data from OpenEMS through MQTT (pretty standard everywhere) than with a custom JSON-RPC/Websocket implementation. So my suggestion is to merge one of the implementations. If Consolinno will maintain the code, choose that one. My integration effort into OpenHAB is stopped until the MQTT Publish capability is implemented in OpenEMS.

martingruening:
I think it is a pitty to leave your MQTT v5 compliant unused and untested. Shall I go ahead and test your code with my local deployment based on the Mosquitto MQTT broker? I would invest time for testing and improving the code if I could be sure that it will be merged based on a new PR from my side. Then we could still improve it from there based on the contribution of @DerStoecki (i.e. adding missing features).

sfeilmeier:
Yes... I am aware and agree. I have it on my todo list, but I just did not have time to have a look at the other implementation and/or eventually merge my implementation. Any feedback from your side will help of course... I really hope I can do it within the next few days.

martingruening:
Awesome! Thank you very much. I will try build this on the weekend and test.

PR #1399, opened 2021-02-23
Improve OpenEMS documentation

sfeilmeier:
Integrate some documentation that was created within the EASY-RES research project


sfeilmeier:
Integrate some documentation that was created within the EASY-RES research project

PR #1402, opened 2021-02-27
Update RRD4j

sfeilmeier:
- Add 'PersistencePriority' to Channels
  - "VERY_HIGH" for Sum-Channels
  - "HIGH" for Nature-Channels
  - "MEDIUM" for local Channels that might be important - like Relay-Status of WAGO Fieldbus
  - ...
- RRD4j only persists Channels with at least the configured priority
- Write RRD4j databases very seldomly, approx. every 5 minutes
- Reduce the size of RRD4j databases
- Migrate last 30 days of old databases

To manually delete RRD4j databases of Channels that do not have sufficient persistence priority, execute in the RRD4j directory:
```
find . -type f -size +700k -exec rm -rf {} \;
```


sfeilmeier:
RRD4j by default only persists channels that have a Persistence Priority of at least MEDIUM. Only these values will be available for local UI, prediction and (later) for resending to the Backend.

@wgerbl, @sebastianasen: Can you please review the definition of Persistence Priority of the Channels?

sebastianasen (approved):
Well done. 
Small doubt: Is it a problem (for now) that the battery Channels like cellVoltages are not saved locally?

sfeilmeier:
@sebastianasen: thanks. It is necessary to reduce the number of locally stored Channels in order to reduce the amount of writes to the hard disk - and Cell Voltages and Cell Temperatures are responsible for a lot of writes. From my perspective it should be sufficient to - by default - only persist Min-/Max Cell-Voltages and -Temperatures. If it is really required to keep all data locally for debugging, one can always set the `persistencePriority` config in RRD4j to `VERY_LOW`, which is the default setting.

- https://github.com/OpenEMS/openems/blob/feature/rrd4j-less-writes/io.openems.edge.timedata.rrd4j/src/io/openems/edge/timedata/rrd4j/Config.java#L22-L23
- https://github.com/OpenEMS/openems/blob/feature/rrd4j-less-writes/io.openems.edge.common/src/io/openems/edge/common/channel/internal/AbstractDoc.java#L65

sebastianasen:
@sfeilmeier: Ok, thaks.

PR #1411, opened 2021-03-10
Feature/dynamic charge

venu-sagar:
controller to handle dynamic prices based charging and discharging of the battery.


sfeilmeier:
This improvement was added by  #1637

PR #1422, opened 2021-03-25
Feature/prediction solcast

hogmoff:
Implements a Predictor for Provider Solcast


pooran-c (approved):
- Edited changes in the BND file
- order of the variable changed in the config file
- Added the empty test folder.

PR #1424, opened 2021-03-26
UI fix: take newest @angular-devkit/ version

fabianfnc:


sfeilmeier:
@fabianfnc, @dadariga: This does not solve the build problems with UI. I believe it is some issue with npm... I was successful with 'yarn install' instead. And updated UI dependencies in #1426. Please give it a try.

PR #1425, opened 2021-03-29
FENECON Home: add Battery Protection

sfeilmeier:
Fenecon Home Battery:
- add Battery-Protection

GoodWe:
- simplify state-machine
- avoid switching between "inbetween" and "full"/"empty" modes
- add state-machine graph for documentation

Generic HybridEss
- add DC power and energy calculation
- add Battery and PV power to debugLog


sfeilmeier:
@huseyinsaht FYI. Nothing to do right now from your side.

PR #1433, opened 2021-04-04
Timedata - MS SQL - First working draft

Laksh-kit:
Hello, 

Similar to Influx DB's functionality, I have tried developing a functionlaity for Microsoft SQL. The components were tested and it seems to work. The build is definitely not complete, and has a lot of hardcoded things, which are specific to KIT Work environement. I have already planned to develop a general working functionality similar to Influx DB. Currently I face few issues with the code in the following 

[https://github.com/Laksh-kit/openems/blob/feature/timedata-mssql/io.openems.shared.mssql/src/io/openems/shared/mssql/MssqlConnector.java](url)

function: **write** (Commented out)

I am not sure how to use OpenemsException classs in the above file. Could you please give me an insight into what i am doing wrong. 

Thanks and regards
Laksh


io.openems.shared.mssql/src/io/openems/shared/mssql/MssqlConnector.java
```diff
@@ -0,0 +1,173 @@
...
+		}
+	}
+	
+	/**CODE TO BE CHECKED
+	public void write(String query) throws OpenemsException {
+		if (this.isReadOnly) {
+			log.info("Read-Only-Mode is activated. Not writing measurement: " +
+					StringUtils.toShortString(query, 100));
+		}
+		try {
```
sfeilmeier, line 134:
```
this.executor.execute(() -> {
	try {
		this.getConnection().createStatement().executeUpdate(query);
		this.getConnection().commit();
	} catch (SQLException e) {
		this.log.warn("Unable to write point: " + e.getMessage());
		try {
			this.getConnection().rollback();
		} catch (SQLException e1) {
			e1.printStackTrace();
		}
	}
});
```

sfeilmeier: Hi Laksh. If you are not used to Java Lamda functions this might have been a bit tricky to understand. The part inside the execute()-method call `() -> {...}` is actually called asynchronously by the executor. Because of this you cannot simply throw the Exception the way you did.

All in all your solution looks good, but is maybe to complicated for the beginning. The structure with a 'shared' bundle for InfluxDB is only there to make the interface available for Backend and Edge. From what I understood this might not (yet) be required in your case, so the architecture could be done simpler. 

Anyhow, the above commented code should work for you. Only downside is, that you will not get an Exception thrown on write errors which you could act on. Right now you would just get a warning on the console log.

Regards,
Stefan

Laksh-kit: Hi Stefan, 
Thanks for the quick review. I have checked it, and the code works as per your suggestion. I am new to JAVA, and yes Lambda functions are not quiet clear to me yet. I appreciate your help. 

Also, yes you are right, currently my requirment is only to write the values from the components to the DB. But in future I also plan in integrating Forecast algorithms (which will carry out the forecasting and write data to DB), which as per my understanding is a part of the backend. Along with that I believe there would be instances where my algorithms (also as per my understanding, a part of the backend) would have to directly interact with the DB (both read and write). In such cases, I think connection elements in a shared file would be necessary. Please correct me if I am wrong. 

Thanks and regards
Laksh

io.openems.edge.timedata.mssql/src/io/openems/edge/timedata/mssql/TimedataMssqlImpl.java
```diff
@@ -44,44 +63,50 @@ public TimedataMssqlImpl() {
...
+	@Reference
+	protected ComponentManager componentManager;
+	
 	@Activate
 	void activate(ComponentContext context, Config config) {
 		super.activate(context, config.id(), config.alias(), config.enabled());
+		this.mssqlConnector = new MssqlConnector(config.server(), config.dbname(), config.port(), 
+				config.username(), config.password(), config.isPasswordProtected(), config.isReadOnly());
+		
 		this.config = config;
```
Laksh-kit, line 76: @sfeilmeier

Hi Stefan,  
At this point of our discussion, I would also need your suggestion, with PreparedStatements. I have planned that I write down the Insert Strings for Prepared statements in another file, and simple call them and activate the prepared statements too during activation of the connections. As per my understanding, I would be compling the statements only once during the whole life cycle of the code, i.e. during activation, and then every data insert would be carried out using the PreparedStatements initiated here. 

Do you think it is a good idea, or could it be somehow made more simpler?

Thanks and regards
Laksh

io.openems.shared.mssql/src/io/openems/shared/mssql/MssqlConnector.java
sfeilmeier: Hi Laksh,
OpenEMS really tries to solve most of the control part locally, so that a Backend should not strictly be required for any control logic. Interaction with a `Timedata` service for reading and writing locally is absolutely possible. So keeping everything in the "OpenEMS Edge"-namespace for now would be sufficient and keep things simple. But your approach is not wrong, of course.

Also again, OpenEMS tries to be as modular and 'pluggable'  as possible. See for example the simple implementation of a 'Persistence-Model-Predictor', that simply 'predicts' the next day with the values from last day. Technically in OpenEMS there is:
- The `Predictor`-Api defines what a Predictor-Service needs to provide, i.e. a `get24HoursPrediction()` method (see https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.predictor.api/src/io/openems/edge/predictor/api/oneday/Predictor24Hours.java)
- The `Timedata`-Api defines what a Timedata-Service needs to provide, e.g. a `queryHistoricData()` method. (see https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.timedata.api/src/io/openems/edge/timedata/api/Timedata.java and https://github.com/OpenEMS/openems/blob/e4ef05bf41f0378c75ece9e9bf435e0b3483281b/io.openems.common/src/io/openems/common/timedata/CommonTimedataService.java)
- `Predictor.PersistenceModel` uses a `Timedata` Service to query the values for the last 24 hours and uses them to serve a Prediction. (see https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.predictor.persistencemodel/src/io/openems/edge/predictor/persistencemodel/PersistenceModelPredictorImpl.java#L75)

This modular structure allows, e.g.
- to use your MSSQL-Timedata-Service immediately for this simple prediction
- to write advanced predection algorithms that use any timedata service, e.g. to feed a neural network
- or you could at any time implement a custom predictor, that uses a webservice (see #1422)

io.openems.edge.timedata.mssql/src/io/openems/edge/timedata/mssql/TimedataMssqlImpl.java
sfeilmeier: Have you seen the Postgres implementation in the Metadata.Odoo bundle? There I am also using PreparedStatements: https://github.com/OpenEMS/openems/blob/develop/io.openems.backend.metadata.odoo/src/io/openems/backend/metadata/odoo/postgres/task/InsertEdgeConfigUpdate.java

If you only have one long-living connection, you should be able to proceed as you described.

PR #1435, opened 2021-04-06
Improvements to GoodWe Battery-Inverter

sfeilmeier:
GoodWe:
- simplify state-machine
- avoid switching between "inbetween" and "full"/"empty" modes
- add state-machine graph for documentation

Generic HybridEss
- add DC power and energy calculation
- add Battery and PV power to debugLog


sfeilmeier:
Moved this code to #1425

PR #1447, opened 2021-04-19
UI: FlatWidget for grid

lukasrgr:
No changes in Style
Screenshot 2021-05-05 164533


ui/src/app/edge/live/common/grid/grid.component.html
```diff
@@ -0,0 +1,10 @@
+<oe-flat-widget img="assets/img/grid.png" [title]="'General.grid' | translate" (click)="presentModal()">
+    <oe-flat-widget-line [name]="'General.gridBuyAdvanced'| translate" [value]="gridBuyAdvancedChannel"
```
sfeilmeier, line 2: Why **advanced**?

ui/src/app/edge/live/flat/abstract-flat-widget.ts
```diff
@@ -28,6 +28,7 @@ export abstract class AbstractFlatWidget implements OnInit, OnDestroy {
         @Inject(Websocket) protected websocket: Websocket,
         @Inject(ActivatedRoute) protected route: ActivatedRoute,
         @Inject(Service) protected service: Service,
+        @Inject(Utils) public utils: Utils,
```
sfeilmeier, line 31: `Utils` are only static methods, so it should not be required to `Inject` it. Injection is only required for services that keep a context/state.

ui/src/app/edge/live/flat/flat-widget-line/abstract-flat-widget-line.ts
```diff
@@ -38,19 +36,19 @@ export abstract class AbstractFlatWidgetLine implements OnDestroy {
...
     }
 
-    protected setValue(value: any) {
-        this.displayValue = this.converter(value);
+    protected setValue(value: any, showNotWhenValueEquals0?: boolean) {
+        this.displayValue = this.converter(value, showNotWhenValueEquals0);
     }
 
-    protected subscribe(channelAddress: ChannelAddress) {
+    protected subscribe(channelAddress: ChannelAddress, showNotWhenValueEquals0?: boolean) {
```
sfeilmeier, line 43: I kind of understand why you need this weird `showNotWhenValueEquals0`, but I don't like it too much.

It should really be solved by using two different static Converters - one that shows zeros and that other one doesn't. Like this you do not need to spill this logic all over the place.

Second, but less prefered option: if you really need it, I would always keep it 'positive' to avoid 'double negative' (doppelte Verneinung) because it makes it very hard to understand the actual logic. I like a bit more the `showWhen0` that you use elsewhere.

sfeilmeier (changes requested):

ui/src/app/edge/live/common/grid/grid.component.html
lukasrgr: removed it out of the name

ui/src/app/edge/live/common/grid/grid.component.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  public gridBuyAdvancedChannel: number;
+  public gridSellAdvancedChannel: number;
+  public gridMode: number;
+
+  protected getChannelAddresses(): ChannelAddress[] {
+    let channelAddresses: ChannelAddress[] = [
+      GridComponent.GRID_ACTIVE_POWER, GridComponent.GRID_MODE, new ChannelAddress('_sum', 'GridActivePowerL1'),
+      new ChannelAddress('_sum', 'GridActivePowerL2'),
+      new ChannelAddress('_sum', 'GridActivePowerL3')
```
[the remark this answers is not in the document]
lukasrgr: done

ui/src/app/edge/live/common/grid/grid.component.ts
```diff
@@ -0,0 +1,45 @@
...
+    ]
+    return channelAddresses;
+  }
+  protected onCurrentData(currentData: CurrentData) {
+    this.gridMode = currentData.allComponents[GridComponent.GRID_MODE.toString()];
+    let gridActivePower = currentData.allComponents[GridComponent.GRID_ACTIVE_POWER.toString()];
+    this.gridBuyAdvancedChannel = gridActivePower;
+    this.gridSellAdvancedChannel = gridActivePower * -1;
```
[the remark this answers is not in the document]
lukasrgr: done

ui/src/app/edge/live/flat/flat-widget-line/abstract-flat-widget-line.ts
lukasrgr: I only implmented this input() param with 'double negative', because if called in html it becomes 'false' without giving value. But you're completely right, i didn't like it too. But to have 2 converters for nearly the same converter, is not effective. 
And if i think about all the converters, which will be implemented in the future, i wasn't sure. 

Anyway for this example i don't need this second converter by now

ui/src/app/edge/live/flat/abstract-flat-widget.ts
sfeilmeier: Static methods should be used in a static way and not with an instance + inject. There should actually never exist an instance of the `Utils` class.

ui/src/app/edge/live/flat/flat-widget-line/abstract-flat-widget-line.ts
sfeilmeier: Having two converters with nearly same logic is no problem in general. Especially if it avoids spoiling the other code, like in this example.

See the Modbus `ElementToChannelConverter` in OpenEMS Edge, where we use a similar technique. It is used to convert from a value read via Modbus interface to a Channel value, e.g. adjusting the value to the proper unit. https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/api/ElementToChannelConverter.java

sfeilmeier (approved):

PR #1449, opened 2021-04-19
TimeslotPeakshaving + HighLoadTimeslot: improve parsing of LocalDate and LocalTime

sfeilmeier:
Do not break if leading zeros are missing. See https://stackoverflow.com/a/57394132/4137113


wgerbl:
Sieht gut aus!

PR #1451, opened 2021-04-20
Update GoodWe Battery Parameters 

huseyinsaht:


sfeilmeier:
This improvement will be further enhanced and embedded in a separate Patch

PR #1464, opened 2021-05-03
Limit usabale capacity

pooran-c:
- Includes limiting usable capacity controller.
- Includes unit test cases for the controller
- Also includes extra channels for the modbusTCPapi controller


io.openems.edge.controller.ess.limitusablecapacity/src/io/openems/edge/controller/ess/limitusablecapacity/LimitUsableCapacityControllerImpl.java
```diff
@@ -0,0 +1,181 @@
...
+
+	@Override
+	public void run() throws OpenemsNamedException {
+		ManagedSymmetricEss ess = this.componentManager.getComponent(this.config.ess_id());
+
+		// Set to normal state and return if SoC is not available
+		Value<Integer> socOpt = ess.getSoc();
+		if (!socOpt.isDefined()) {
+			this.state = State.NO_LIMIT;
+			return;
```
sfeilmeier, line 78: If you return here, your Channels are never written, i.e. the `STATE_MACHINE` Channel can never be `NO_LIMIT` and also if the SoC of a ESS disappears (e.g. cable was removed), `ALLOWED_CHARGE_POWER` and `ALLOWED_DISCHARGE_POWER` will still show the old values; even though the limit is not applied.

io.openems.edge.controller.ess.limitusablecapacity/src/io/openems/edge/controller/ess/limitusablecapacity/LimitUsableCapacityControllerImpl.java
```diff
@@ -0,0 +1,181 @@
...
+				&& this.config.stopDischargeSoc() > this.config.allowDischargeSoc()
+				&& this.config.allowDischargeSoc() > this.config.allowChargeSoc()
+				&& this.config.allowChargeSoc() > this.config.stopChargeSoc()) {
+			throw new OpenemsException("Please re-check the configuration, invalid values present in the Soc columns");
+		}
+	}
+
+	@Override
+	public void run() throws OpenemsNamedException {
+		ManagedSymmetricEss ess = this.componentManager.getComponent(this.config.ess_id());
```
sfeilmeier, line 72: It would be preferrable to use `@Reference` for getting the `ManagedSymmetricEss`.

I discussed also with @JohannesHundschell, that it would be good to allow having this Controller applied on multiple ESS at the same time. This avoids having to configure it for each and every ESS in a big system separately. Can you implement that or do you need support?

io.openems.edge.controller.ess.limitusablecapacity/readme.adoc
```diff
@@ -0,0 +1,16 @@
...
+== On lower boundary
+
+- Allow discharging till "Stop-Discharge-Soc"
+- If ESS SoC reduces further till "Force-Charge-Soc", start to force-charge the ESS with 20 % of ESS `MaxApparentPower` till "Stop-Discharge-Soc"
+- Afterwards block any discharging below "Allow-Discharge-Soc"
+
+== On upper boundary
+
+- Allow charging till "Stop-Charge-Soc"
+- Afterwards block any charging above "Allow Charge Soc"
```
sfeilmeier, line 14: Can you validate if I got the description right?

io.openems.edge.controller.ess.limitusablecapacity/src/io/openems/edge/controller/ess/limitusablecapacity/LimitUsableCapacityControllerImpl.java
```diff
@@ -0,0 +1,181 @@
...
+		int soc = socOpt.get();
+
+		// initialize allowedCharge
+		Integer allowedCharge = null;
+
+		// initialize allowedDischarge
+		Integer allowedDischarge = null;
+
+		boolean stateChanged;
+		do {
```
sfeilmeier, line 89: I would have preferred to use our default State-Machine implementation; but in general this is fine.

sfeilmeier (changes requested):

io.openems.edge.controller.ess.limitusablecapacity/src/io/openems/edge/controller/ess/limitusablecapacity/LimitUsableCapacityControllerImpl.java
sfeilmeier: Here is an example for `@Reference` with multiple Component-IDs:

https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.meter.virtual/src/io/openems/edge/meter/virtual/symmetric/subtract/VirtualSubtractMeter.java#L65-L69

pooran-c: Changed with ManagedSymmetricEss reference.

io.openems.edge.controller.ess.limitusablecapacity/src/io/openems/edge/controller/ess/limitusablecapacity/LimitUsableCapacityControllerImpl.java
pooran-c: Updated this.

io.openems.edge.controller.ess.limitusablecapacity/readme.adoc
pooran-c: Yes, Only small change, start charging with 20 % of maxapparent power or static power value.

PR #1487, opened 2021-05-28
SoC estimation improvement in simulator.

venu-sagar:
Improved the Soc calculation by using the double value instead of float and estimating the soc through the current energy value in the battery.


sfeilmeier (approved):
Well done. Thanks!

PR #1503, opened 2021-06-04
Updated the MQTT component to optionally filter its messages by topic paths

JohnVidler:
The current implementation of MQTT can turn in to a bit of a fire hose of data for situations where you only want some of the fields possible. This patch adds a filter function which inhibits certain topics from actually sending to the broker, instead they silently complete, doing nothing.

A single new configuration string has been added, into which a semicolon-delimited set of topic filters can be written allowing only topics which match these specifications to pass.

The topic filters use the same syntax as normal MQTT client subscriptions, so can use the `+` and `#` specifiers to indicate single, or multi-level paths.

An example path might be `edge/fems33/channel/meter+/+` which would pull all metrics in meter0, meter1, etc. paths, but exclude any other topics.

THIS NEEDS TESTING - Please can someone with an otherwise working setup please check that this works, as I'm currently unable to, as I don't have access to our test hardware.


da-Kai:
implemented with #3321

PR #1526, opened 2021-06-21
Export Channels to Excel

sfeilmeier:
Adds a link in UI Profile to download an Excel file with all Channels and their current value. Also adds information about the 'source' of a Channel, e.g. the modbus register address.

grafik

Example export:
[ChannelExport-femsXXX-scheduler0.xlsx](https://github.com/OpenEMS/openems/files/6686152/ChannelExport-femsXXX-scheduler0.xlsx)

Example with Modbus Sources
grafik


sebastianasen (approved):
Really good improvement!

PR #1541, opened 2021-06-22
Implement go-e Charger Home

hogmoff:
Implements a go-e Charger Home 

There is still sometimes a strange behaviour if plugging the car. The state is permanently changing from loading to stopping. Don't know where the issue comes from.


nicoketzer:
When exactly does this happen? When plugging in an empty e-car or a full one? or when unplugging?

Possibly I can also generate the bug with my setup via RasPi under the same conditions?

Do you use the ModBUS TCP API or the HTTP API?

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/Config.java
```diff
@@ -0,0 +1,35 @@
...
+
+@ObjectClassDefinition(name = "EVCS go-e Charger Home", //
+		description = "Implements the go-e Charger Home electric vehicle charging station.")
+@interface Config {
+
+	@AttributeDefinition(name = "Component-ID", description = "Unique ID of this Component")
+	String id() default "evcs0";
+
+	@AttributeDefinition(name = "Alias", description = "Human-readable name of this Component; defaults to Component-ID")
+	String alias() default "";
```
nicoketzer, line 14: Maybe set default to "Go-e Charger [COUNT OF CHARGERS]"

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeAPI.java
```diff
@@ -0,0 +1,259 @@
...
+	}
+	
+	/** Sets the Current in Ampere for go-e
+	 * 
+	 * See https://github.com/goecharger
+	 * 
+	 * @return JsonObject with new settings
+	 * @throws OpenemsNamedException on error
+	 */
+	public JsonObject setCurrent(int current) {
```
nicoketzer, line 121: Maybe you could add a check if the entered current is greater than or equal to the minimum current and less than or equal to the maximum current. (I don't know but maybe you have built in that the setCurrent function can only be called with values between min and max. Then the recheck would be again pointless)

Code in Java for smth like that

```
public static int constrain_value(int value, int min, int max) { 
   if(min > max){
      //Switch Min and Max if Min is bigger than max
      return constrain_value(value, max, min);
   }else{
      //Normal Return Path
      if(min == max){
         //If Min and Max are the Same the Value must also be the Same so return min or max 
         return min;
      }else{
         if(value >= min){
            if(value <= max){
               return value;
            }else{
               return max;
            }  
         }else{
            return min;
         }
      }
   }   
}
```

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeAPI.java
```diff
@@ -0,0 +1,259 @@
...
+				throw new OpenemsException(
+						"Error while reading from go-e API. Response code: " + status + ". " + body);
+			}
+		} catch (OpenemsNamedException | IOException e) {
+			throw new OpenemsException(
+					"Unable to read from go-e API. " + e.getClass().getSimpleName() + ": " + e.getMessage());
+		}
+	}
+	
+}
```
nicoketzer, line 259: A function that reads out the current (total) power demand from OpenEMS and then adjusts the charging power of the charging station depending on how high it is would also be interesting.

In addition, a PV surplus charging function would also be cool. 

Intended would be that when the battery is charged or fed into the grid that this surplus energy (from 3.3 kW surplus possible with Go-E Home Chargers) is passed on to the charging box (eg at 3.3kW surplus a charging current of 6A, at <3.3kW switch off and at >3.3kW the current is calculated via the power in kW : 230V : 3 (since on L1 L2 L3)).

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeChargerHomeImpl.java
```diff
@@ -0,0 +1,340 @@
...
+	private int GetMaxHardwarePower(int cableCurrent) {
+		int MaxPower = 0;
+		int phases = 3;
+		if (cableCurrent < this.MaxCurrent && cableCurrent > 0) {
+			MaxPower = phases*cableCurrent*230/1000;
+		}
+		else {
+			MaxPower = phases*this.MaxCurrent*230/1000;
+		}
+		return MaxPower;
```
nicoketzer, line 192: Here you could include that is considered how much power the inverter can. Because that would then be the maximum power that can be used without having to draw power from the grid (e.g. GoodWe GW 10 --> 10kW max so the maximum power without having to draw from the grid would be 10kW).

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeChargerHomeImpl.java
```diff
@@ -0,0 +1,340 @@
...
+		if (MinPower > MinPowerHw || MinPower < 0) {
+			return MinPowerHw;
+		}
+		else {
+			return MinPower;
+		}
+	}
+		
+	private int GetMaxPower(int phases) {
+		return phases*this.MaxCurrent*230/1000;
```
nicoketzer, line 213: Should also be limited by the Go-E Charger Hardware:
11kW Go-e Home Charger: 16A --> 230V * 16A * 3 = 11.040 W 
22kW Go-e Home Charger: 32A --> 230V * 32A * 3 = 22.080 W

So for the 11kW Charger the calculated max Power would be 22.080W (22kW) because the default "maxCurrent" is 32000mA also the real maxCurrent for a 11kW Charger is 16A

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeChargerHomeImpl.java
```diff
@@ -0,0 +1,340 @@
...
+	 * discarded. The value is also depending on the DIP-switch settings and the
+	 * used cable of the charging station.
+	 */
+	private void setPower() {
+		
+		WriteChannel<Integer> energyLimitChannel = this.channel(ManagedEvcs.ChannelId.SET_ENERGY_LIMIT);
+		int energyLimit = energyLimitChannel.getNextValue().orElse(0);
+
+		// Check energy limit
+		if (energyLimit == 0 || energyLimit > this.getEnergySession().orElse(0)) {
```
nicoketzer, line 230: A Option to set "Charge until Battery has xy%" would also be nice

nicoketzer (commented):
Technically well implemented. 
Perhaps in a few places still a little polish necessary but otherwise fine (see comments).

sfeilmeier:
@sebastianasen Could you have a look at this?

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeChargerHomeImpl.java
```diff
@@ -0,0 +1,340 @@
...
+		case 24: // 0b00011000: Phase 1+2 is active
+			return 2;
+		case 56: // 0b00111000: Phase1-3 are active
+			return 3;
+		default:
+			return 0;
+		}
+	}
+	
+	private int GetMaxHardwarePower(int cableCurrent) {
```
sebastianasen, line 183: Sorry - It is not perfectly declared what the minimum Hardware Power and the maximum hardware power is and for what it is used.
Maybe we must change that generally, but for now, the minimum and maximum hardware power includes the power that can be set to the charging station in that situation.

This Channel must be set according to the minimum current, the used phases and the default voltage.
It is mainly used for the UI and the evcs Controller:

### Charging with one phase:
grafik

### Charging with three phases:
grafik

Normally we would write 6A on the left side and 32A on the right side - that would be much more easier.
But as the complete system and every Controller is working with power values and it is easier to understand that we have e.g. 12 kW from the PV and are charging these 12 kW in the car and so on ... we decided this way.

		int minHwCurrent = this.parent.config.minHwCurrent() / 1000;
		this.parent.channel(Evcs.ChannelId.MINIMUM_HARDWARE_POWER)
		.setNextValue(230 /* Voltage*/ * minHwCurrent /* min current*/ * phases);

This can be set when we read the relevant data.

In your case, the customer would be able to set 22kW even if the car is using one phase. 
Target of 22kW -> about 95A cannot be set to the EVCS and is confusing the customer in the Monitoring

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeChargerHomeImpl.java
```diff
@@ -0,0 +1,340 @@
...
+		if (cableCurrent < this.MaxCurrent && cableCurrent > 0) {
+			MaxPower = phases*cableCurrent*230/1000;
+		}
+		else {
+			MaxPower = phases*this.MaxCurrent*230/1000;
+		}
+		return MaxPower;
+	}
+	
+	private int GetMinHardwarePower(int phases, int cableCurrent) {
```
sebastianasen, line 195: Please start also in here with small letters for variables and method names

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeChargerHomeImpl.java
```diff
@@ -0,0 +1,340 @@
...
+		int MinPower = 0;
+		int MinPowerHw = phases*this.MinCurrent*230/1000;
+		if (cableCurrent < this.MinCurrent && cableCurrent > 0) {
```
sebastianasen, line 198: Take always the greatest of both.
this.MinCurrent can be set to more than the 6A coming from the charger.
For example when we activate a Renault Zoe mode in the UI we changed the configured minimum to 10A because the Zoe is not charging bellow

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeChargerHomeImpl.java
```diff
@@ -0,0 +1,340 @@
...
+		}
+		if (MinPower > MinPowerHw || MinPower < 0) {
+			return MinPowerHw;
+		}
+		else {
+			return MinPower;
+		}
+	}
+		
+	private int GetMaxPower(int phases) {
```
sebastianasen, line 212: As i wrote above by calling this method. This Channel is not set by the hardware limits - so please remove it.

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeChargerHomeImpl.java
sebastianasen: This method is anyways not needed, but if we are not getting the real maximum current 11 or 22 kW charger from the charger itself, we have only the possibility to get this value out of the configuration

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeChargerHomeImpl.java
sebastianasen: Yes of course, but this should be added in the nature and used in the controller.

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeChargerHomeImpl.java
```diff
@@ -0,0 +1,340 @@
...
+						.orElse(0)) { /* charging under MINIMUM_HARDWARE_POWER isn't possible */
+					power = 0;
+					this.goeapi.setActive(false);
+				}
+				else {
+					this.goeapi.setActive(true);
+				}
+				
+				Value<Integer> phases = this.getPhases();
+				Integer current = power * 1000 / phases.orElse(3) /* e.g. 3 phases */ / 230; /* voltage */
```
sebastianasen, line 247: The power we get should be already in W, so "*1000" is not needed.

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeAPI.java
sebastianasen: This is already done in when the method is called - could be done in here as well.

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeAPI.java
```diff
@@ -0,0 +1,259 @@
...
+			
+		try {
+			if (current != this.parent.activeCurrent) {
+				JsonObject json = new JsonObject();
+				if (!debugMode) {
+					Integer CurrentAmpere = current / 1000;
+					String URL = "http://" + this.IP + "/mqtt?payload=amp=" + Integer.toString(CurrentAmpere);
+					json = this.sendRequest(URL, "PUT");
+					this.parent.activeCurrent = current;
+					return json;
```
sebastianasen, line 131: Please make sure that the results are set if there is no communication error happening. 

this.parent._setSetChargePowerLimit(power);
(This is setting the normal value of that channel and not the write value, to store the last power setpoint)

This would be an interesting part, what we could really log, if the debug mode is activated
this.parent.debugLog(result.toString());

This could be also done after you call the method, when the json you return includes a success message or you return a boolean.

sebastianasen (changes requested):
Thank you @hogmoff for the implementation.

I noted my change requests.
Please ask if something is not clear.

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeAPI.java
nicoketzer: To 1:
In my instalation my whole Home Consumes over the Backup Line (continus) from the Goodwe GW 10 Inverter. So i can max. draw around 3.33kW per Phase (10kW in total). So for me it would be great if the Charger automaticly in-/decrease the Charging Power so that the 3.33kW maximum Power per Phase is not exeded. 

E.g. If i have nothing running my Car should Charge with full 10kW but if i cook or smth. like that and draw 2kW on L1 than it should regulate the Charger from 3.33kW per Phase to the maximum of 1.3kW per Phase(so that the 3.33 kW per Phase will not be reached)[i wrote 1.3kW per Phase because the Charger can not async charging per Phase(my knowlege point) so i must limit L1 L2 and L3 when only L1 whould be needed]

To 2:
That' great!

To 3:
The 3.3kW is the minimum Charging Power for a Charger with three Phase's connected.
(Minimum Current per Phase is 6A so 230V * 6A * 3 ≈ 3.3kW)

If you only connected your Charger with one Phase than the minimum Xharging Power is 1.1kW (230V*6A)

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeChargerHomeImpl.java
nicoketzer: Im not sure on that but i think somewhere deep deep in the API is a value (i will lookup that!)

But if not is it possible to make a Select-Field so that the User itself (the user must know witch charger he got) can Select his Type when he adds the Go-E charger?

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeAPI.java
```diff
@@ -0,0 +1,259 @@
...
+	}
+	
+	/** Limit MaxEnergy for go-e
+	 * 
+	 * See https://github.com/goecharger
+	 * 
+	 * @return JsonObject with new settings
+	 * @throws OpenemsNamedException on error
+	 */
+	public boolean limitMaxEnergy(boolean limit) {
```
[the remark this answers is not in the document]
nicoketzer: I can help out here.

In your go-e charger you can define a Value of Energy after that the Charger automaticly stop's charging.

E.g. you can say "After 30kWh stop charging"

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeAPI.java
sebastianasen: Ok thanks for explanation.
in 3. your wrong - I think the 3.3 was only for the calculation in point 1. 
230 * 6 * 3 = 4.140    With one phase: 1.380

But the main point for me is that the implementation of the charging station has nothing to do with any storage, pv or e.g. your heat pump.

The EVCS itself is only responsible for reading Data and provide commands to set the charge power.

Even the storage system itself is not doing anything as a standalone Component.

For the entire logic, we are using separate controllers, that expecting e.g. a EVCS and a ESS - no matter what evcs and ess configured.

As long as the given Components provides the complete Nature Channels and their functionality.
https://github.com/OpenEMS/openems/blob/f57602689f46a31e47e779068a5f742c624ec03a/io.openems.edge.controller.evcs/src/io/openems/edge/controller/evcs/EvcsController.java#L252

The evcs controller is calculating the current power, depending on the current user inputs and the current storag & meter informations

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeAPI.java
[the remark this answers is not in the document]
sebastianasen: Ok, thanks. We have the posibillity not everywhere, so the maximum Energy feature was done by us. We checkt if the energy limit was reached and set a charge request of zero in that case.

Do you know if there is any problem if we set zero as a target. Not every evcs can handle it and is confused if we do this.

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeChargerHomeImpl.java
sebastianasen: We are pre configuring such things for customers, but of course it would be great to have the direct infromation from the charger

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeChargerHomeImpl.java
```diff
@@ -0,0 +1,340 @@
...
+		switch(Phasen) {
+		case 8: // 0b00001000: Phase 1 is active
+			return 1;
+		case 24: // 0b00011000: Phase 1+2 is active
+			return 2;
+		case 56: // 0b00111000: Phase1-3 are active
+			return 3;
+		default:
+			return 0;
+		}
```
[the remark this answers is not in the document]
nicoketzer: Yes and no.
If you dont know how many Phases are connected you can not calc (in most cases) the Power that is given out.
This is the case because the Go-e Charger mostly set's his output Power ofer a Current Value so if you set 6A it makes a diverence if you have P = 230V*6A*1 or P = 230V*6A*2 or P = 230V*6A*3

And if you mean if there's a difference betw. L1 and L2 or L2 and L3 or L3 and L1 than i can say from my point of knowlege that this doesn't matter the charger only say's "One Phase, Two Phase's or Three Phase's"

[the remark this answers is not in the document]
sebastianasen: Yes, the last point cleared my question, thanks

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeChargerHomeImpl.java
nicoketzer: The minimum Hardware Power is 6A*230V * [Count of connected Phases]

The maximum Power is e.g. 11kW or 22kW (depends on the Model of the Charger) but the thing is a 11kW Charger can also set up to 32A but than the charger automaticly reduces to the max. of 16A

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeChargerHomeImpl.java
nicoketzer: That's true but i meen he wantet to define the absolut minimum power that the charger can output

-> For sure it's a good idea to make this changeable

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeAPI.java
nicoketzer: Yea if you calc the Input-Value's you totaly right its about 4,1kW 
but i said 3,3kW because at my charger if i set Current to 6A only around 5.3A are flowing and 212V (says the Charger itself)
In the real world this values arent true. My Car shows e.g. 3,7kW when the charger say's 3,3kW. This comes from the diffrence the Charger shows in Voltage (Measured Voltage at Charger is ≈ 226V and Charger say's thers only 212V)


Ok thats absolutly clear now but as you sayed the Evcs - Controller does this (is this right or am i wrong?) is it than possible to say this Controller "hey when you over 10kW shut down the charger"?

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeChargerHomeImpl.java
nicoketzer: I just looked up the Api Doc at GH and found that

Screenshot_20210623_184030_com android chrome

If my understanding is right the cbl parameter gives Informations about how much Current is the max Power

hogmoff:
> When exactly does this happen? When plugging in an empty e-car or a full one? or when unplugging?
> 
> Possibly I can also generate the bug with my setup via RasPi under the same conditions?
> 
> Also maybe you must set a wait interval for readout/send command and the actual action happening and ready to give back as status. I did have the same Problem when i wrote 10A Current via the Rest API and then directly read out the Current via the "/status". Sometimes it gave me the actual Value and sometimes the go-e charer was to slow and i got back the old Current-Value. A little Delay of about 500ms fixed this for me

This happens only in Combination with the Controller for electrical vehicle station in automatic mode and a non full and plugged e-car.

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/Config.java
hogmoff: This is really unusual in the Openems Project. In any other EVCS the default value is an empty string

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeChargerHomeImpl.java
hogmoff: yes, that's right. My cable allows 32A and status for the parameter is "cbl": "32".
If you use a 22kW Charger with a 16A Cable the maximum Hardware-Power is 11kW. So i took the maximum current from configuration and limit the value with the used cable.

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeAPI.java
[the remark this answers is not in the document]
hogmoff: the value of zero is allowed as target, no problems with that thing here

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeChargerHomeImpl.java
hogmoff: Sure, but the unit for the Charger-current is mA. So i've to multiply with 1000

nicoketzer:
> > When exactly does this happen? When plugging in an empty e-car or a full one? or when unplugging?
> > Possibly I can also generate the bug with my setup via RasPi under the same conditions?
> > Also maybe you must set a wait interval for readout/send command and the actual action happening and ready to give back as status. I did have the same Problem when i wrote 10A Current via the Rest API and then directly read out the Current via the "/status". Sometimes it gave me the actual Value and sometimes the go-e charer was to slow and i got back the old Current-Value. A little Delay of about 500ms fixed this for me
> 
> This happens only in Combination with the Controller for electrical vehicle station in automatic mode and a non full and plugged e-car.

Is Only The State "broken" or is the car also not charging?

hogmoff:
> > > When exactly does this happen? When plugging in an empty e-car or a full one? or when unplugging?
> > > Possibly I can also generate the bug with my setup via RasPi under the same conditions?
> > > Also maybe you must set a wait interval for readout/send command and the actual action happening and ready to give back as status. I did have the same Problem when i wrote 10A Current via the Rest API and then directly read out the Current via the "/status". Sometimes it gave me the actual Value and sometimes the go-e charer was to slow and i got back the old Current-Value. A little Delay of about 500ms fixed this for me
> > 
> > 
> > This happens only in Combination with the Controller for electrical vehicle station in automatic mode and a non full and plugged e-car.
> 
> Is Only The State "broken" or is the car also not charging?

the car is charging for a few seconds and after the next status the activation is unchecked. If i set the controller to manuel mode than all works fine.

io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeChargerHomeImpl.java
hogmoff: okay, then i've to set the minimum always to this.minCurrent. Otherwise with a 32A cable the minCurrent is the same as the maxCurrent

hogmoff:
> > > > When exactly does this happen? When plugging in an empty e-car or a full one? or when unplugging?
> > > > Possibly I can also generate the bug with my setup via RasPi under the same conditions?
> > > > Also maybe you must set a wait interval for readout/send command and the actual action happening and ready to give back as status. I did have the same Problem when i wrote 10A Current via the Rest API and then directly read out the Current via the "/status". Sometimes it gave me the actual Value and sometimes the go-e charer was to slow and i got back the old Current-Value. A little Delay of about 500ms fixed this for me
> > > 
> > > 
> > > This happens only in Combination with the Controller for electrical vehicle station in automatic mode and a non full and plugged e-car.
> > 
> > 
> > Is Only The State "broken" or is the car also not charging?
> 
> the car is charging for a few seconds and after the next status the activation is unchecked. If i set the controller to manuel mode than all works fine.

i found the issue although in the goeapi.getStatus() Function. The active state (alw) is set in some time values to 0 without any car plugged. One step further to fix this.

hogmoff:
I've the same issue with starting and stopping with the current commit. But now I can show the problem with a screenshot in Chronograf. Configuration is with Controller in automatic mode. The Charger starts after plugging the car and after 10cycles the status from car is "fully loaded" whats not the case. Then in the next cycle the status is again in charging mode and so on.
grafik
I'm really confused because if i change to manual mode then everything works fine. This was the case at the end at the screenshot after appr. 1min.
Any ideas whats wrong here? Maybe it's relevant, the car is an ID.3

hogmoff:
I think that i've found the problem in https://github.com/hogmoff/openems/blob/feature/goe/io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeChargerHomeImpl.java

If the SET_CHARGE_POWER_LIMIT is lower than the MINIMUM_HARDWARE_POWER the charger will be deactivated. That's the case when no more Power is available (more consumption, less production) in automatic mode. Now it's clear, that this behaviour is not in manuell mode.

grafik

if (power < minimumHardwarePowerChannel.value()
            .orElse(0)) { /* charging under MINIMUM_HARDWARE_POWER isn't possible */
          power = 0;
          this.goeapi.setActive(false);

Can I set the MINIMUM_HARDWARE_POWER instead of deactivate the charger? What is the best way to handle this?

sebastianasen:
> 
> 
> I think that i've found the problem in https://github.com/hogmoff/openems/blob/feature/goe/io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeChargerHomeImpl.java
> 
> If the SET_CHARGE_POWER_LIMIT is lower than the MINIMUM_HARDWARE_POWER the charger will be deactivated. That's the case when no more Power is available (more consumption, less production) in automatic mode. Now it's clear, that this behaviour is not in manuell mode.
> 
> grafik
> 
> if (power < minimumHardwarePowerChannel.value()
> .orElse(0)) { /* charging under MINIMUM_HARDWARE_POWER isn't possible */
> power = 0;
> this.goeapi.setActive(false);
> 
> Can I set the MINIMUM_HARDWARE_POWER instead of deactivate the charger? What is the best way to handle this?

I do not get why you want to set the minimum hardware power in here.
The minimum Hardware Power will be something like  1.380 W or 4.140 W depending on the phases that are used.
Each for 6A as a minimum.
A lower current setpoint cannot be taken by such a AC charger.

An additional feature would be that we can still charge the car for a configurable time if the available power suddenly drops.
We didn't touched it for now, because the user has already so many things to configure and we saw that many customers are confused for the first time, what every configuration means.

hogmoff:
> > I think that i've found the problem in https://github.com/hogmoff/openems/blob/feature/goe/io.openems.edge.evcs.goe.chargerhome/src/io/openems/edge/evcs/goe/chargerhome/GoeChargerHomeImpl.java
> > If the SET_CHARGE_POWER_LIMIT is lower than the MINIMUM_HARDWARE_POWER the charger will be deactivated. That's the case when no more Power is available (more consumption, less production) in automatic mode. Now it's clear, that this behaviour is not in manuell mode.
> > grafik
> > if (power < minimumHardwarePowerChannel.value()
> > .orElse(0)) { /* charging under MINIMUM_HARDWARE_POWER isn't possible */
> > power = 0;
> > this.goeapi.setActive(false);
> > Can I set the MINIMUM_HARDWARE_POWER instead of deactivate the charger? What is the best way to handle this?
> 
> I do not get why you want to set the minimum hardware power in here.
> The minimum Hardware Power will be something like 1.380 W or 4.140 W depending on the phases that are used.
> Each for 6A as a minimum.
> A lower current setpoint cannot be taken by such a AC charger.
> 
> An additional feature would be that we can still charge the car for a configurable time if the available power suddenly drops.
> We didn't touched it for now, because the user has already so many things to configure and we saw that many customers are confused for the first time, what every configuration means.

Now i have it, thank you for your explanation. Than the strange behavour i mentioned before comes from the controller and not from the goe-api.
From my point of view the code is now ready.

sebastianasen (commented):
@sfeilmeier Please have also a look on this

sebastianasen:
@sfeilmeier  Please have a look for a final review

sfeilmeier (approved):
Thanks a lot @hogmoff for the superb work!

PR #1560, opened 2021-07-11
Fix UI grid component wrong display sell buy

clehne:
OpenEMS-UI grid component always shows BuyPower.
This patch divides between sell and buy power. On sellPower sellPower is shown and buy power shows "-". This is the same behavior as storage component does.
buyPower will be shown appropriately


sfeilmeier:
@lukasrgr: Can you have a look at this after your holiday please? I assume it's already fixed in our internal fork.

PR #1561, opened 2021-07-11
Feature documentation add sma tripower8 10

clehne:
Sunny Tripower PV inverter driver is also working fine on types 8.0 and 10.0. Tested on both devices.


sfeilmeier (approved):
Thanks!

PR #1573, opened 2021-07-21
Similar day prediction model.

pooran-c:
Implementation of similar day model,
- This model takes a "channel's address and the number of weeks" as input from config to make predictions for one day.
   - Average of the last four similar day values is the predicted data.
- this pull request also has a unit test case.


io.openems.edge.predictor.similardaymodel/readme.adoc
```diff
@@ -0,0 +1,3 @@
+= io.openems.edge.predictor.similardaymodel
+
```
sfeilmeier, line 2: Fix title + add some descriptive information like "Similar day model uses the same weekday of the last 4 weeks - where 4 is configurable - to predict the power" ... "This approach is specifically useful to predict Consumption." "Accuracy of this approach has been scientifically verified within the EMSIG project https://openems.io/research/emsig/"...

io.openems.edge.predictor.similardaymodel/src/io/openems/edge/predictor/similardaymodel/Config.java
```diff
@@ -0,0 +1,28 @@
...
+	@AttributeDefinition(name = "Component-ID", description = "Unique ID of this Component")
+	String id() default "predictor0";
+
+	@AttributeDefinition(name = "Alias", description = "Human-readable name of this Component; defaults to Component-ID")
+	String alias() default "";
+
+	@AttributeDefinition(name = "Is enabled?", description = "Is this Component enabled?")
+	boolean enabled() default true;
+	
+	@AttributeDefinition(name = "How many weaks?", description = "No of weaks the data is needed?")
```
sfeilmeier, line 20: Typo 'weeks'

io.openems.edge.predictor.similardaymodel/src/io/openems/edge/predictor/similardaymodel/Config.java
```diff
@@ -0,0 +1,28 @@
+package io.openems.edge.predictor.similardaymodel;
+
+import org.osgi.service.metatype.annotations.AttributeDefinition;
+import org.osgi.service.metatype.annotations.ObjectClassDefinition;
+
+@ObjectClassDefinition(//
+		name = "Predictor Similarday-Model", //
+		description = "Implements Similar day predictor model")
```
sfeilmeier, line 8: Use the same writing troughout the code and docs, e.g. "Similar-Day"

sfeilmeier (changes requested):

sfeilmeier:
The implementation looks very good! We need to find a way though to prioritize this predictor to Same-as-yesterday, if both are configured - but still fallback to same-as-yesterday e.g. if there is not sufficient data (i.e. OpenEMS running for less than 4 weeks).

sfeilmeier (approved):

Issue #1569, opened 2021-07-14
Simulator DataSource: Single Direct does not work anymore.


Fyusel:
```
- [X] bug report
- [ ] feature request
```

This issue occurs since version 2021.10.0 (same for 2021.11.0 and 2021.12.0)
Last working version: 2021.9.0

When creating the "Simulator DataSource: Single Direct" component the following error occurs:

```
2021-07-14 08:09:08,180 [re.Cycle] ERROR [.direct.SingleDatasourceDirect] bundle io.openems.edge.simulator:1.0.0.202107130634 (131)[Simulator.Datasource.Single.Direct(30)] : The activate method has thrown an exception
Jul 14 08:09:08 raspberrypi java[478]: java.lang.IllegalArgumentException: Converter for value [[Ljava.lang.String;@ccf336] of type [class [Ljava.lang.String;] to type [INTEGER] is not implemented.
Jul 14 08:09:08 raspberrypi java[478]:         at io.openems.edge.common.type.TypeUtils.getAsType(TypeUtils.java:318)
Jul 14 08:09:08 raspberrypi java[478]:         at io.openems.edge.common.component.AbstractOpenemsComponent.addChannelsForProperties(AbstractOpenemsComponent.java:310)
Jul 14 08:09:08 raspberrypi java[478]:         at io.openems.edge.common.component.AbstractOpenemsComponent.addChannelsForProperties(AbstractOpenemsComponent.java:260)
Jul 14 08:09:08 raspberrypi java[478]:         at io.openems.edge.common.component.AbstractOpenemsComponent.updateContext(AbstractOpenemsComponent.java:225)
Jul 14 08:09:08 raspberrypi java[478]:         at io.openems.edge.common.component.AbstractOpenemsComponent.activate(AbstractOpenemsComponent.java:140)
Jul 14 08:09:08 raspberrypi java[478]:         at io.openems.edge.simulator.datasource.api.AbstractCsvDatasource.activate(AbstractCsvDatasource.java:39)
Jul 14 08:09:08 raspberrypi java[478]:         at io.openems.edge.simulator.datasource.single.direct.SingleDatasourceDirect.activate(SingleDatasourceDirect.java:43)
Jul 14 08:09:08 raspberrypi java[478]:         at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

PR #1586, opened 2021-08-05
Array-Properties: handle datatype properly

sfeilmeier:
- Tries to extract 'one' value of the array
- Falls back to 'STRING' if there are more two or more values
- Handle 'IllegalArgumentException' instead of fail


sfeilmeier:
Fixes #1569

PR #1588, opened 2021-08-06
Implementation of an installation assistant

raphael-piller:
This pull request includes the implementation of an installation assistant.

The assistant has the following noteworthy functions:
- It is organized in different views, which can look like this:
installer_view
- Through the different views data gets collected
- In the corresponding views, the data gets used, for example to configure the system as specified

This pull request also introduces the possibility for the user to create an account, which then can
be used to login and use the installation assistent.


sfeilmeier:
This video shows the procedure of the installation assistant: https://youtu.be/GduyiJHvqik?t=654

PR #1613, opened 2021-09-06
Modbus/TCP Api Controller: improve error handling

sfeilmeier:
Unhandled runtime Exceptions, e.g. `IllegalArgumentException` can cause the j2mod library to crash. This pull-request adds error handling and only throws allowed and handled Exceptions. Additionally the Exception stacktrace is printed to console and a `ProcessImageFault` channel is set.


wgerbl:
Looks good as we discussed

PR #1644, opened 2021-10-22
InfluxDB: write data with milliseconds precision

sfeilmeier:
The timestamp is rounded to the global configured Cycle-Time


sfeilmeier:
Tested successfully by Con Seifert - see https://community.openems.io/t/arbeiten-mit-influxdb/637/12

PR #1686, opened 2021-12-20
Switch to log4j2

OlafSeng:
Hi Stefan,

as discussed in the [community](https://community.openems.io/t/schwachstelle-in-log4j-und-openems/707/5), I have created a merge request regarding my Log4J2 changes.
The idea is to use a `PaxOsgi`appender, as described [here](https://ops4j1.jira.com/wiki/spaces/paxlogging/pages/499613746/Configuring+Log4J2).

I was able to test the changes successfully for the OpenEMS edge running on my local machine. However, the configuration file in 'pax/logging.config' is not updated when I start the backend locally.
Unfortunately, I was not able to find out why, but the problem already existed before applying my changes, so it might be an issue with the development branch.

Originally, I have developed these changes for a slightly older version of OpenEMS which still runs on Java 8. On this version, both OpenEMS Backend and Edge are working fine.

Regards,

Olaf

### DIFF
--- a/cnf/pom.xml
+++ b/cnf/pom.xml
@@ -245,12 +245,12 @@
 		<dependency>
 			<groupId>org.ops4j.pax.logging</groupId>
 			<artifactId>pax-logging-api</artifactId>
-			<version>2.0.11</version>
+			<version>2.0.13</version>
 		</dependency>
 		<dependency>
 			<groupId>org.ops4j.pax.logging</groupId>
-			<artifactId>pax-logging-log4j1</artifactId>
-			<version>2.0.11</version>
+			<artifactId>pax-logging-log4j2</artifactId>
+			<version>2.0.13</version>
 		</dependency>
 		<dependency>
 			<groupId>org.osgi</groupId>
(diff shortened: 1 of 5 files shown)


sfeilmeier:
Hi Olaf,

thank you very much for your contribution! I fixed updating of the configuration files for Edge and Backend. It works reliably here on my machine. 

Could you give it a final test so that we can merge it?

Regards,
Stefan

OlafSeng:
Hi Stefan,

thank you for fixing the configuration. Now it works reliably on my machine as well, so I would say we can merge the changes.

Funny thing is, at first it didn't work. But after I switched from '16.0.1.j9-adpt' to '17.0.1.12.1-amzn' and completely rebuilt the application, everything is fine.

Regards,

Olaf

sfeilmeier (approved):

sfeilmeier:
Thanks for checking. Strange... sometimes the build behaviour is weird. I usually rely on my `prepare-commit.sh` script for the final check. (https://github.com/OpenEMS/openems/blob/develop/tools/prepare-commit.sh) 

Thanks again for the contribution!

PR #1747, opened 2022-03-02
Feature/revpi-full-io-support

mishbieg:
This brings in the latest librevpi-dio-java 0.2.0 which has support for accessing the Revpi hardware beyond boolean values. Further, the existing Revpi Digital IO Implementation is moved into a subfolder as a prestep to adding other Revpi IO modules. The plan is to add a common module that all IO modules can use, which contains common code. In later steps the individual IO modules are added and support for multiple instances will be added.


mishbieg:
@sfeilmeier please have a look and let me know if you are OK with my idea of reorganising the Revpi DIO module before adding more IO modules. Luckily, C.Lene already released a new version of the library, so my changes should no longer be needed. We can start adding more Revpi modules right away.

mishbieg:
Updated the branch an created new pull request

PR #1769, opened 2022-03-25
reworked Config and Influx to be compatible with InfluxDBv2

simonfey:
@sfeilmeier I reworked the Config for InfluxDB. This allows users to connect their Backend to InfluxCloud (with https). You can still connect to classic instances by entering an appropriate URL with protocol, ip & port and setting username:password as apikey.

### DIFF
--- a/io.openems.backend.timedata.influx/src/io/openems/backend/timedata/influx/Config.java
+++ b/io.openems.backend.timedata.influx/src/io/openems/backend/timedata/influx/Config.java
@@ -8,23 +8,20 @@ import org.osgi.service.metatype.annotations.ObjectClassDefinition;
 		description = "Configures the InfluxDB timedata provider")
 @interface Config {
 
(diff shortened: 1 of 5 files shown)


sfeilmeier:
Thanks @simonfey! I applied some more clean-ups and documentation hints and migrated also the Edge component.

zoernert:
@simonfey - Coul you provide a sample config for v1.8 - I tried to get it from the hints - but failed. 

Result:
`ERROR [backend.timedata.influx.Influx] [Timedata.InfluxDB] Unable to write to InfluxDB. NotFoundException: null`

However my configuration should be right (bucket = database, ApiKey = username:password, ...) - and this testing server worked before. So I expect that the URL (in my case I have  http://localhost:8086 ) should be right.

sfeilmeier:
It works for me with the following settings:
- url="http://localhost:8086"
- org="-"
- apiKey="user:password"
- bucket="db/default"
- measurement="data"

I believe your `bucket` is wrong. Must be `{database}/{retention-policy}`

PR #1772, opened 2022-03-28
Update MeterMicrocareSdm630Impl.java

DerWahreKlinki:
Changed Modbus addresses for reactive power. Used to point at the registers of apparent power


sfeilmeier (approved):
Perfekte Arbeit. 👍 Vielen Dank!

PR #1782, opened 2022-04-04
fixed filter for influxdbv2 query to use edge instead of fems keyword

simonfey:
Tested with OpenEMS backend.
This filter does not work, since edges by default use the influxdb field "edge" instead of "fems"


sfeilmeier (approved):

sfeilmeier:
Sorry, missed this when backporting from FEMS.

PR #1801, opened 2022-04-14
Consider inverter losses for HybridEss

simonfey:
Hi @sfeilmeier. This PR is related to https://github.com/OpenEMS/openems/pull/1800
I noticed that the AC Limit calculation neglects any losses on the inverter. 
This has a high impact on the actual charge rate of the battery (for me this was about 50% off).
This change will consider the losses on the inverter and provides a more precise AC limit.
In case of a calculated EssDischargePower from an unmanaged battery the losses would be 0 and not make any difference.


simonfey:
I noticed that the calculation is not reliable since the current on DC side ramps up quite quickly since the AC side may have a delay of several seconds. This calculation is a lot more reliable when I build a moving average. Consider this PR as draft.

simonfey:
I did some further experimentation in this and did not like the solution. I managed to achieve similar results by adjusting the risk parameter.

PR #1822, opened 2022-05-10
Fix Gitpod and switch to openems-odoo version 14

sfeilmeier:


sfeilmeier:
Note: updated docs are available at https://openems.github.io/openems.io/openems/latest/simulation/gitpod.html

PR #1894, opened 2022-07-13
Enhance a generic Sunspec meter to also support Elgris meters

tsicking:
The Sunspec common starting block of SolarEdge is different from the one of Elgris meters. Therefore I created this patch.


io.openems.edge.meter.sunspec/src/io/openems/edge/meter/sunspec/SunSpecMeters.java
```diff
@@ -0,0 +1,17 @@
+package io.openems.edge.meter.sunspec;
+
+public enum SunSpecMeters {
+	SOLAR_EDGE("SolarEdge", 2), //
+	ELGRIS_ZERO_EXPORT("Elgris Zero Export", 1), //
+	ELGRIS_SMART_METER("Elgris Smart Meter", 1), //
+	TQ_EM_420("TQ EM420", 1);
```
sfeilmeier, line 7: This is similar to my critics in #1991. We should really create individual Meter components per hardware manufacturer/device. This makes it much more visible to OpenEMS users, which devices are supported and also allows for more specific adjustments per device down the line. This might not be needed so much for meters, but it's our experience e.g. with OCPP charging stations, etc.

See the implementation of SolarEdge Grid-Meter as an example (-> https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.solaredge/src/io/openems/edge/solaredge/gridmeter/SolarEdgeGridMeter.java).

tsicking: I have created such components.

sfeilmeier:
@tsicking: I know this has been pending for a very long time, but what is the reason you closed this? (we need to find a way so that I am not the only person reviewing PRs... it's just way too much work for me to handle)

tsicking:
I thought this is quite outdated, as it is still based on the old `SymmetricMeter` and `AsymmetricMeter` interfaces, and the repository from which this PR was created is not the one we currently use. 
We are going to make a new PR of the Elgris meter soon.

PR #1895, opened 2022-07-16
Add Meter B+G E-Tech DRT428M2 Modbus RTU

mishbieg:
This change adds the DRT428M2 Modbus RTU meter from B+G E-Tech.
The documentation of the register setup is also addes as PDF.

Change-Id: Ia21244687fb8fec8d2be317ceb1146cf16a5c8a7


pooran-c:
1. Please change the channel Id class as an interface. 
    
channelIdInterface

2. Remove unwanted channels.

3. Please rename the implementation file
implementationFile

4. Styling, please classes follow camel case

mishbieg:
> 1. Please change the channel Id class as an interface.
> 
> 
> channelIdInterface
> 
>     2. Remove unwanted channels.
> 
>     3. Please rename the implementation file
>        implementationFile
> 
>     4. Styling, please classes follow camel case

Thanks for the review. I think I fixed everything you mentioned. Technically I think the meter can be used for production and consumption. While I had no trouble verifying that consumption is messured in both Grid and Consumption mode, I'm unsure how I can verify Production? Do you have any tips on what I need to simulate so that I can messure and verify production? For my project, currently it is enough if it meassures consumption. Just thinking of others, that want to use the meter. If there are code changes needed to add more support, I guess we can do that in another pull request.

pooran-c:
Take a look at this discussion thread, 
https://community.openems.io/t/simulate-pv-production-and-grid-consumption/821
I think this is similar to your question.

pooran-c:
Could just add Active power here in debuglog() method in "MeterBgeTechDrt428M2Impl.java", and if needed you add the channels in the "ctrlDebugLog" controller.

example :
 	@Override
	public String debugLog() {
		return "L:" + this.getActivePower().asString();
	}

mishbieg:
> Take a look at this discussion thread, https://community.openems.io/t/simulate-pv-production-and-grid-consumption/821 I think this is similar to your question.

Thanks, I will look into that. I guess the grid acting meter simulator should help.

pooran-c (approved):
Approving the pull request.

mishbieg:
Thanks 👍

PR #1944, opened 2022-10-06
New bundle for Alfen EVCS

tsicking:


io.openems.edge.evcs.alfen/src/io/openems/edge/evcs/alfen/Alfen.java
```diff
@@ -0,0 +1,158 @@
...
+import io.openems.common.exceptions.OpenemsError.OpenemsNamedException;
+import io.openems.common.types.OpenemsType;
+import io.openems.edge.common.channel.Doc;
+import io.openems.edge.common.channel.WriteChannel;
+import io.openems.edge.common.component.OpenemsComponent;
+
+
+public interface Alfen extends OpenemsComponent {
+
+	public enum ChannelId implements io.openems.edge.common.channel.ChannelId {
```
clehne, line 16: The EVCS interface has changed in the last couple of weeks. We need to check if this still compiles with the new interface

io.openems.edge.evcs.alfen/src/io/openems/edge/evcs/alfen/AlfenImpl.java
```diff
@@ -0,0 +1,340 @@
...
+	}
+
+	@Activate
+	void activate(ComponentContext context, Config config) throws OpenemsException {
+		if (super.activate(context, config.id(), config.alias(), config.enabled(), config.modbusUnitId(), this.cm,
+				"Modbus", config.modbus_id())) {
+			return;
+		}
+
+		this._setMaximumHardwarePower(22000);
```
clehne, line 93: This values should be configurable. We should handle the configuration same as Keba implementation does - by using mA for min and max range of chargepoint. Note that this has to do with the UI Renault Zoe option, which reconfigures the chargepoint by setting a different min current (in mA)

clehne (changes requested):
see my comments

tsicking:
The bundle needs to be updated to fit with the new Evcs interface. An adapted pull request will come soon.

PR #1961, opened 2022-11-02
ElementToChannelConverter KEEP_POSITIVE fixed

tsicking:
In the previous version, the ElementToChannelConverter KEEP_POSITIVE mapped a value to a converter instead of a new value.


sfeilmeier (approved):
True. Thanks for finding it and fixing it! 👍

PR #1964, opened 2022-11-04
FEMS Backports 2022.11

sfeilmeier:
* Implementation of Weidmüller fieldbus coupler UR20-FBC-MOD-TCP-V2 
  * Configuration is dynamically read from Modbus and Channels are created accordingly in the same format as for the existing WAGO fieldbus coupler.
* Implementation of Meter Ziehl EFR4001IP
* GoodWe Grid-Meter: implement ModbusSlave, improve code, HIGH modbus priority
* Modbus-Bridge Cleanup ** Prepare for bigger changes in Modbus-Bridge implementation 
  * Rename Thread inside AbstractWorker to `thread` and make it protected 
  * Drop priority ONCE, because it has no feasible use the way it is implemented currently. Use LOW instead. It makes sure the value is actually going to be read, even if e.g. the connection cannot be established from the beginning 
  * Simplify TasksManager and move unused parts to Kostal Piko implementation
* UI: Wait for App till translations are loaded
* Structural changes in EVCS 
  * AbstractManagedEvcs
  * Write handler no longer needed in every Evcs Component 
  * Abstract Methods to ensure the minimum required information 
  * Calculating the flexible hardware power limits depending on the hardware limits and current used phases 
  * Handles the current ChargeState (Additionally to a normal State "CHARGING", it is aware if the limit is increasing, decreasing, normal or not charging)
  * This State is important because an evcs and the car itself need from 5 to 40 seconds (round about), till a limit is taken 
* Changed:
  * Phases stored as Enum
  * MaximumHardwarePower set by the component changed to FixedMaximumHardwarePower
  * MaximumHardwarePower still present (for boundaries in UI), but calculated automatically mentioned in AbstractManagedEvcs 
  * EvcsCluster ClusterState depending on ChargeStates of each evcs. 
  * Blocking new distribution if there was a regulation before and one or more evcss still increasing or decreasing. 
  * Limit for each evcs is set with a filter
  * Fallback solution if the grid meter is above the limit 
  * Evcs Comonents
  * Extending AbstractManagedEvcs & changes according to that
* Fix TypeUtils multiply
* Improvements in App-Center


io.openems.edge.evcs.api/src/io/openems/edge/evcs/api/Evcs.java
```diff
@@ -70,25 +80,81 @@ public enum ChannelId implements io.openems.edge.common.channel.ChannelId {
...
+		 * 
 		 * <ul>
 		 * <li>Interface: Evcs
 		 * <li>Readable
 		 * <li>Type: Integer
 		 * </ul>
 		 */
-		PHASES(Doc.of(OpenemsType.INTEGER) //
+		PHASES(Doc.of(Phases.values()) //
+				.debounce(5) //
```
cvabc, line 93: Is there any reason to add the `decounce(5)`?

clehne: On some charging station models the phases are not switched consistently. Instead within the very first seconds they are switch several times, e.g. from 3 to 1 to 3 to 1. I have seen that often.

cvabc: Thank you!

For EVCS with fixed phases, usually it's simply set during initialization:
https://github.com/OpenEMS/openems/blob/c78334d8810d2a291511a9fa3e1c6072c03da84a/io.openems.edge.evcs.hardybarth/src/io/openems/edge/evcs/hardybarth/EvcsHardyBarthImpl.java#L79
which "works" for 3 phase EVCS, since by default this channel is 3 phases.

However, we found that for our case of only 1 phase, the `this._setPhases(Phases.ONE_PHASE);` does not work.
Does it mean we need to repeat the `this._setPhases(Phases.ONE_PHASE);` for 5 times?

PR #1976, opened 2022-11-10
Modbus Bridge: optimize execution of Tasks

sfeilmeier:
The idea is to try to execute all Write-Tasks as early as possible (directly after the EXECUTE_WRITE event) and all Read-Tasks as late as possible to have values available exactly when they are needed (i.e. at the BEFORE_PROCESS_IMAGE event). A WaitHandler is introduced, that is responsible for calculating a 'wait' time before execution of Modbus ReadTasks. It uses an internal queue of waiting times that would have been possible in previous Cycles. The minimum value of this queue is taken as next waiting time.

See discussion at OpenEMS Community: https://community.openems.io/t/zeitlicher-ablauf-der-modbus-tasks/734

PR #1977, opened 2022-11-10
Possibility to set the Modbus priority in the Open-EMS-UI for KM-Tronic Relay Board

nicoketzer:
In some cases it is necessary to set the priority of individual relay boards to High. Since a realization by a separate build for the required system usually comes with problems this is a general implementation.

Useful or necessary is this if e.g. over a switching operation of a relay a start or stop operation is triggered or if the switched relay is supposed to execute other important activities.

With this implementation it is possible to have several relay boards in one OpenEMS-Edge which have different Modbus priorities. 

Default value is set to "LOW" as before to not influence existing systems.

This was tested locally and on an own live instance of Openems in connection with a KM-Tronic 4 channel relay board (RTU) and other devices on the bus.


sfeilmeier:
Hi Nico,

Thanks for the effort. I understand that the underlying use-case for this improvement is for example a system where multiple devices are connected to one Modbus/RTU interface. Example: `battery1`, `battery2`, `battery3`, `battery4`, `io0` all connected to `modbus0`. All `battery*` components have multiple LOW and HIGH priority read tasks and one write task (See e.g. [Soltaro Single Battery]( https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.battery.soltaro/src/io/openems/edge/battery/soltaro/single/versionb/SingleRackVersionBImpl.java#L239)) The `io0` component currently only has one LOW priority read task and multiple write tasks (See e.g. [KMTronic Relay Board]( https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.io.kmtronic/src/io/openems/edge/io/kmtronic/four/KmtronicRelay4PortImpl.java#L70)). In the example none of the batteries are currently reachable for communication because they would have to be started by an output signal via KMTronic.

Currently the ModbusWorker implementation would first execute on every cycle:
- All high priority tasks of all devices
- One low priority task of one device
- All write tasks of all devices

After detecting, that the batteries are 'defective', it will execute on every cycle:
- One high priority tasks of all devices
- One low priority task of one device
- All write tasks of all devices

The read task for the `io0` would - even with non-defective components - be executed only once in a while. E.g. if there are 120 LOW priority read tasks over all devices, this task gets executed only once every 120 cycle. Now that the read and write tasks of the four defective batteries are executed on every cycle, they will delay the entire process. This leads to a situation where `io0` data is hardly ever read at all (e.g. once in 10 minutes).

https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/api/worker/ModbusWorker.java#L71-L88) for implementation details.

Your pull request improves the problem a little bit, as it allows setting the `io0` priority to HIGH, which will eventually cause the `io0` data to be read maybe once per minute. But it does unfortunately not solve the underlying problem.

Additionally this HIGH priority would only be needed while the batteries are not available. But as your suggestion is a fixed configuration of the Component this cannot be automated properly to 'switch back to LOW priority' if it is not anymore required.

My alternative pull request (#1976) tries to solve the underlying problem instead, by actually disabling all reads and writes to a defective component for a period of time. (See https://github.com/OpenEMS/openems/blob/feature/modbus-task-execution/io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/api/worker/DefectiveComponents.java#L13). The pull request also adds some other improvements, like timing of read tasks in properly working systems as discussed in OpenEMS Community (https://community.openems.io/t/zeitlicher-ablauf-der-modbus-tasks/734)

Till now I only tested my algorithm in software JUnit tests (see https://github.com/OpenEMS/openems/blob/feature/modbus-task-execution/io.openems.edge.bridge.modbus/test/io/openems/edge/bridge/modbus/api/worker/ModbusWorkerTest.java#L127) and have not been able to test it in hardware. 

But as in my opinion this would be the more generic solution to the problem, I would rather try to focus on that. I would very much appreciate if you could support in testing the other pull request with actual hardware.

Regards,
Stefan

nicoketzer:
Hello Stefan,
thanks first of all for your detailed answer, my answer has been delayed a bit now.

I agree in principle that the problem you describe would be solved by the implementation I presented, but only the symptom is fought and the problem is not fixed at the root.

The reason why I personally implemented this is that most of my heating system is controlled by openems and also by my own FEMS (FEMS from a Fenecon Home + external OpenEMS that was adapted to my needs).

Since I want to make sure that the read and write tasks for the KM-Tronic relay boards of my heating control are done "at all costs" this implementation was created which I now want to share with the community. In addition, it was not an option for me to bring out only an OpenEMS version that classifies all tasks as "HIGH". This has the background that also some elements in my house (e.g. outdoor lighting, ...) are controlled by another KM-Tronic-Relay-Board and these are " not important " to me when it comes to a readout or write speed ( whether the light switches off 10s later or earlier and whether the state arrives 10s later or directly in OpenEMS is in the "Outdoor Light" case completely irrelevant ). 

Nevertheless I like to support you with the test of your pull request which wants to solve the "Modbus problem" fundamentally. For this purpose my FEMS is available at any time (we can discuss details directly).

Regards,
Nico

PR #1991, opened 2022-11-24
Feature/carlo gavazzi em24

tsicking:
Package renamed and added EM24.


io.openems.edge.meter.carlo.gavazzi/src/io/openems/edge/meter/carlo/gavazzi/MeterCarloGavazziEmSeriesImpl.java
```diff
@@ -32,25 +31,25 @@
 
 @Designate(ocd = Config.class, factory = true)
 @Component(//
-		name = "Meter.CarloGavazzi.EM300", //
+		name = "Meter.CarloGavazzi.EM3X0", //
```
sfeilmeier, line 34: This change is critical, because it breaks every system that currently has an EM300 configured.

As the protocols anyway differ slightly, should we rather avoid the configuration setting for "Model" and have two different Components?

I searched a bit through the Carlo Gavazzi protocols, but unfortunately there does not seem to be an identification register like SOCOMEC has.

tsicking: I created a seperate component for the EM24 and a general abstract component, which is extended by both the EM300 and EM24.

clehne:
Outdated. Will be replaced by a new PR.

PR #2020, opened 2022-12-28
EdgeWebsocket and UiWebsocket does not start reliable

opernikus-common:
EdgeWebsocket and UiWebsocket does not start reliable if FileMetadata is used. Due to asynchronous Event Handling and the given FileMetadata implementation Metadata.Events.AFTER_IS_INITIALIZED may not be received by the socket bundles OR it may be received multiple times. FileMetadata may be very fast (Testenv with few users only) and Odoo Metadata may be too slow (huge number of users). The socket servers are now started either by Event or within the activate() method, whatever comes first.

### DIFF
--- a/io.openems.backend.common/src/io/openems/backend/common/metadata/AbstractMetadata.java
+++ b/io.openems.backend.common/src/io/openems/backend/common/metadata/AbstractMetadata.java
@@ -23,8 +23,10 @@ public abstract class AbstractMetadata extends AbstractOpenemsBackendComponent i
 	 * Make sure to call this method once initialized!.
 	 */
 	protected void setInitialized() {
-		this.isInitialized.set(true);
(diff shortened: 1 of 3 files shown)


sfeilmeier:
@da-Kai We have discussed this once after you had changed this to events. Can you please review + merge this PR? Thank you!

sfeilmeier:
Discussed this with @michaelgrill. The proposed solution in https://github.com/opernikus-common/openems/pull/1 does not work, because it drops the actual 'isInitialized' feature. Still the problem could/should be solved using OSGi features. There is the "Condition Service Specification" since OSGi 8, which we could use for this: https://docs.osgi.org/specification/osgi.core/8.0.0/service.condition.html

clehne:
Closed because no longer up-to-date. See https://github.com/OpenEMS/openems/pull/2401/ for more info.

PR #2034, opened 2023-01-17
Bug: CalculateAverage now supports all numbers instead of integer

JoopAue:
Hi team,

Thank you for the wonderful project! 

During development I ran into, what seems to me, a bug in the CalculateAverage class.

---

- CalculateAverage::addValue resulted in ClassCastExceptions when the channel was not of type Integer
- This happened because the parameter of the method was Channel<Integer>

- Now supports any implementation of Number

---
Let me know if there is anything I can improve


sfeilmeier (approved):
Thank you very much for the contribution & bugfix. No further comments from my side - great work!

PR #2041, opened 2023-01-30
Enhance influx bundle to use either FluxQL or InfluxQL language

opernikus-common:
FluxQL raises performance issues on systems with high load. Especially the loading of the history view is significantly slower with FLUXQL as it is with InfluxQL. The new java influx library supports both ways to access an influx DB. This pull requests adds functionality to configure which query language should be used. It
* adds a configuration option to backend/timedata/influx/Config.java (with configuration backward compatibility)
* adds a configuration option to edge/timedata/influxdb/Config.java (with configuration backward compatibility)
* io.openems/shared/influxdb 
** is splitted into an Interface InfluxConnector
** a Common and a FluxQL implementation
** added a InfluxQL implementation
Notes:
* provide a getLatestValue implementation also
* FluxQL should have only very minor changes in InfluxConnectorFluxQL:queryChannelValuesMap. The hardcoded DB name "db/default" is now configurable. Also the data fetch range is increased from "-5m" to "-16m" (fix UI history display issue for systems which update DB channels only every 15min).

### DIFF
--- a/io.openems.backend.timedata.influx/src/io/openems/backend/timedata/influx/Config.java
+++ b/io.openems.backend.timedata.influx/src/io/openems/backend/timedata/influx/Config.java
@@ -13,15 +13,18 @@ import org.osgi.service.metatype.annotations.ObjectClassDefinition;
 
 	@AttributeDefinition(name = "Org", description = "The Organisation; '-' for InfluxDB v1")
 	String org() default "-";
-	
+
 	@AttributeDefinition(name = "ApiKey", description = "The ApiKey; 'username:password' for InfluxDB v1")
 	String apiKey();
 
 	@AttributeDefinition(name = "Bucket", description = "The bucket name; 'database/retentionPolicy' for InfluxDB v1")
 	String bucket();
-	
+
 	@AttributeDefinition(name = "Measurement", description = "The InfluxDB measurement")
 	String measurement() default "data";
+
+	@AttributeDefinition(name = "Use FLUX Queries?", description = "If enabled FLUX query language is used, else InfluxQL will be used")
+	boolean useFluxQueries() default true;
 
 	@AttributeDefinition(name = "Read-Only mode", description = "Activates the read-only mode. Then no data is written to InfluxDB.")
 	boolean isReadOnly() default false;
--- a/io.openems.shared.influxdb/src/io/openems/shared/influxdb/InfluxConnectorFluxQL.java
+++ b/io.openems.shared.influxdb/src/io/openems/shared/influxdb/InfluxConnectorFluxQL.java
@@ -1,0 +1,414 @@
+package io.openems.shared.influxdb;
+
+import java.net.URI;
+import java.time.ZonedDateTime;
+import java.time.temporal.ChronoUnit;
+import java.time.temporal.TemporalAdjusters;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.SortedMap;
+import java.util.TreeMap;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import com.google.gson.JsonElement;
+import com.google.gson.JsonNull;
+import com.google.gson.JsonPrimitive;
+import com.influxdb.query.FluxRecord;
+import com.influxdb.query.FluxTable;
+import com.influxdb.query.dsl.Flux;
+import com.influxdb.query.dsl.functions.restriction.Restrictions;
+
+import io.openems.common.OpenemsOEM;
+import io.openems.common.exceptions.OpenemsError.OpenemsNamedException;
+import io.openems.common.exceptions.OpenemsException;
+import io.openems.common.timedata.Resolution;
+import io.openems.common.types.ChannelAddress;
+import io.openems.common.utils.JsonUtils;
+
+public class InfluxConnectorFluxQL extends InfluxConnectorCommon implements InfluxConnector {
+
+	/**
+	 * The Constructor.
+	 *
+	 * @param url          URL of the InfluxDB-Server (http://ip:port)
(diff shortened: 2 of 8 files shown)


sfeilmeier:
@opernikus-common: we found the same issues in our InfluxDB and fixed it in a very similar way locally... just did not yet have time to push it upstream. Will try to see if we can merge it.

sfeilmeier (commented):
It's unfortunate that we implemented the same thing twice in different ways. I have only small comments here but would appreciate if you could have a look at my alternative implementation #2044.

Internally we did a lot more changes in Backend Timedata services in order to prepare for future performance optimizations. Introduction of TimedataManager so that we can create multiple Timedata databases: one optimized for UI with continuous aggregates and another one for data warehouse applications. #2045 is working in production here.

io.openems.shared.influxdb/src/io/openems/shared/influxdb/InfluxConnectorFluxQL.java
```diff
@@ -0,0 +1,414 @@
...
+						result.put(channelAddress, JsonUtils.getAsJsonElement(record.getValue()));
+					}
+				}
+			}
+
+		} catch (OpenemsException e) {
+			this.log.error(e.getMessage());
+			e.printStackTrace();
+
+		}
```
DerStoecki, line 278:
```suggestion
  try {
            channelAddresses.forEach(address -> {
                var builder = new StringBuilder() //
                        .append("from(bucket: \"").append(this.bucket).append("\")") //
                        .append("|> range(start: -16m)") //
                        .append("|> filter(fn: (r) => ") //
                        .append("r._measurement == \"").append(InfluxConnector.MEASUREMENT).append("\" ");
                if (influxEdgeId.isPresent()) {
                    builder.append("and r." + OpenemsOEM.INFLUXDB_TAG + " == \"" + influxEdgeId.get() + "\" ");
                }
                builder //
                        .append("and r._field == \"").append(address.toString()).append("\")") //
                        .append("|> last()");
                var query = builder.toString();

                // Execute query
                List<FluxTable> queryResult = null;
                try {
                    queryResult = this.executeQuery(query);


                    queryResult.forEach(table -> {
                        table.getRecords().forEach(record -> result.put(address, JsonUtils.getAsJsonElement(record.getValue())));
                    });
                } catch (OpenemsException e) {
                    throw new RuntimeException(e);
                }

            });
        } catch (RuntimeException e) {
            this.log.error(e.getMessage());
            e.printStackTrace();
            //throw new OpenemsException(e.getCause()); <- if you rather want to throw 
            //the exception as an OpenEmsException
        }
```

io.openems.shared.influxdb/src/io/openems/shared/influxdb/InfluxConnectorFluxQL.java
```diff
@@ -0,0 +1,414 @@
...
+	}
+
+	/**
+	 * Queries the latest available channel values.
+	 *
+	 * @param influxEdgeId     the unique, numeric Edge-ID; or Empty to query all
+	 *                         Edges
+	 * @param channelAddresses the {@link ChannelAddress}es
+	 * @return a map of {@link ChannelAddress}es and values
+	 * @throws OpenemsException on error
```
DerStoecki, line 242: If you are not throwing this -> remove throws

io.openems.shared.influxdb/src/io/openems/shared/influxdb/InfluxConnectorFluxQL.java
```diff
@@ -0,0 +1,414 @@
...
+	 */
+	public Map<ChannelAddress, JsonElement> queryChannelValuesMap(Optional<Integer> influxEdgeId,
```
DerStoecki, line 244: i can only see 1 use and its within the class itself -> therefor should be private

DerStoecki (commented):
I did not test the implementation, but at first glance it looks good.

There are some changes i would suggest:
1. as already mentioned -> usage of Enums instead of booleans and create InfluxConnectors via a Factory-> pass through the enum and other config files and the factory should create the instance
2. the rest are mainly for(var foo : bar) to java.stream suggestions
3. there are some duplicated code parts that can be abstracted into connectorcommon

io.openems.backend.timedata.influx/src/io/openems/backend/timedata/influx/Config.java
```diff
@@ -13,16 +13,19 @@
...
 
 	@AttributeDefinition(name = "Bucket", description = "The bucket name; 'database/retentionPolicy' for InfluxDB v1")
 	String bucket();
-	
+
 	@AttributeDefinition(name = "Measurement", description = "The InfluxDB measurement")
 	String measurement() default "data";
 
+	@AttributeDefinition(name = "Use FLUX Queries?", description = "If enabled FLUX query language is used, else InfluxQL will be used")
+	boolean useFluxQueries() default true;
```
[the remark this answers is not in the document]
clehne: Good hint! Thanks.

clehne:
> It's unfortunate that we implemented the same thing twice in different ways. I have only small comments here but would appreciate if you could have a look at my alternative implementation #2044.
> 
> Internally we did a lot more changes in Backend Timedata services in order to prepare for future performance optimizations. Introduction of TimedataManager so that we can create multiple Timedata databases: one optimized for UI with continuous aggregates and another one for data warehouse applications. #2045 is working in production here.

Thanks for sharing your version. Had a very short review of #2044. Code quality looks better than in our version. I think we can integrate it easily. Will try it in our testenvironment and then switch to #2044 soon. If everything works fine I will close #2041.

opernikus-common:
PR closed. see PR #2044 for a more advanced version.

sfeilmeier:
@DerStoecki, @clehne Could we then settle on #2045. Please review there. I sent invitations fo you for OpenEMS repository. Thanks!

PR #2042, opened 2023-01-30
Proposal for OpenEMS Contribution Guidelines

opernikus-common:
This is a proposal to anchor contribution guidelines into OpenEMS. 

This contribution guidelines are based on [PR 1006](https://github.com/OpenEMS/openems/pull/1006) from @fabianfnc. 
The document has been expanded to include suggestions from @sfeilmeier as well as some thoughts from the consolinno and opernikus team. It adds a subpage into the antora OpenEMS documentation tree.


sfeilmeier (approved):
Thank you! I am merging this as a first version, which we can continuously build on. Specifically all the bullet points should be explained with examples. But it's better to have something than nothing for now.

PR #2047, opened 2023-02-02
proposal version for a new EVCS API (v2)

opernikus-common:
The current EVCS nature is not suitable for larger chargeparks. It includes unnessecary or unclear information (e.g. MaxPower, MaxHwPower, FixedMaxPower ) and it is missing important information (e.g. PhaseRotation, PowerObject, Constraints).

This pull request opens a proposal for a change on the EVCS nature. It puts a new version of the evcs api  (io.openems.edge.evcs.v2.api) beside the old implementation. It should not be merged into the development branch. It should be taken as a reference for a discussion on a future EVCS nature.

Therefore we also like to introduce multiple EvcsPower objects (analog to EssPower), each responsible for a single electrical cabinet (done in another PR). It makes it a lot easier to limit Currents on a given power line segment and build controller hierarchies.

For more a discussion on this topic see commmunity post [Proposal: New EVCS Api](https://community.openems.io/t/proposal-new-evcs-api/1308)

### DIFF
--- a/io.openems.edge.evcs.api/src/io/openems/edge/evcs/v2/api/Evcs.java
+++ b/io.openems.edge.evcs.api/src/io/openems/edge/evcs/v2/api/Evcs.java
@@ -1,0 +1,846 @@
+package io.openems.edge.evcs.v2.api;
+
+import java.util.function.Consumer;
+
+import io.openems.common.channel.AccessMode;
+import io.openems.common.channel.Level;
+import io.openems.common.channel.PersistencePriority;
+import io.openems.common.channel.Unit;
+import io.openems.common.types.OpenemsType;
+import io.openems.edge.common.channel.Channel;
+import io.openems.edge.common.channel.Doc;
+import io.openems.edge.common.channel.EnumReadChannel;
+import io.openems.edge.common.channel.IntegerReadChannel;
+import io.openems.edge.common.channel.LongReadChannel;
+import io.openems.edge.common.channel.StateChannel;
+import io.openems.edge.common.channel.value.Value;
+import io.openems.edge.common.component.OpenemsComponent;
+import io.openems.edge.common.modbusslave.ModbusSlaveNatureTable;
+import io.openems.edge.common.modbusslave.ModbusType;
+import io.openems.edge.common.type.TypeUtils;
+
+public interface Evcs extends OpenemsComponent {
+
+    public static final Integer DEFAULT_MAXIMUM_HARDWARE_POWER = 22_080; // W
+    public static final Integer DEFAULT_MINIMUM_HARDWARE_POWER = 4_140; // W
+    public static final Integer DEFAULT_MAXIMUM_HARDWARE_CURRENT = 32_000; // mA
+    public static final Integer DEFAULT_MINIMUM_HARDWARE_CURRENT = 6_000; // mA
+    public static final Integer DEFAULT_VOLTAGE = 230; // V
+    public static final int DEFAULT_POWER_RECISION = 230;
+
+    public enum ChannelId implements io.openems.edge.common.channel.ChannelId {
+
+	/**
+	 * Status.
+	 *
+	 * <p>
+	 * The Status of the EVCS charging station.
+	 *
+	 * <ul>
+	 * <li>Interface: Evcs
+	 * <li>Readable
+	 * <li>Type: Status @see {@link Status}
+	 * </ul>
+	 */
+	STATUS(Doc.of(Status.values()) //
+		.accessMode(AccessMode.READ_ONLY) //
+		.persistencePriority(PersistencePriority.HIGH)), //
+
+	/**
+	 * Charge Power.
+	 *
+	 * <p>
+	 * <ul>
+	 * <li>Interface: Evcs
+	 * <li>Readable
+	 * <li>Type: Integer
+	 * <li>Unit: W
+	 * </ul>
+	 */
+	CHARGE_POWER(Doc.of(OpenemsType.INTEGER) //
+		.unit(Unit.WATT).accessMode(AccessMode.READ_ONLY) //
+		.persistencePriority(PersistencePriority.HIGH)), //
+
+	/**
+	 * Minimum Power this hardware needs to run.
+	 * 
+	 * <p>
+	 * The power is always given for all three phases. If MinimumPower = 4140W
+	 * (3*1380W) and a vehicle charges on 2 phases than the MinimumPower the EVCS
+	 * uses is 2706W (2*1380W).
+	 *
+	 * <ul>
+	 * <li>Interface: Evcs
+	 * <li>Readable
+	 * <li>Type: Integer
+	 * <li>Unit: W
+	 * </ul>
+	 */
+	MINIMUM_POWER(Doc.of(OpenemsType.INTEGER) //
+		.unit(Unit.WATT) //
+		.accessMode(AccessMode.READ_ONLY) //
+		.persistencePriority(PersistencePriority.HIGH)), //
+
+	/**
+	 * Maximum Power defined by software.
+	 *
+	 * <p>
+	 * The power is always given for all three phases. If MaximumPower = 22080W
+	 * (3*73600W) and a vehicle charges on 2 phases than the MaximumPower for this
+	 * EVCS is 14720W (2*7360W).
+	 * 
+	 * <ul>
+	 * <li>Interface: Evcs
+	 * <li>Readable
+	 * <li>Type: Integer
+	 * <li>Unit: W
+	 * </ul>
+	 */
+	MAXIMUM_POWER(Doc.of(OpenemsType.INTEGER) //
+		.unit(Unit.WATT) //
+		.accessMode(AccessMode.READ_ONLY) //
+		.persistencePriority(PersistencePriority.HIGH)), //
+
+	/**
+	 * Charging Type.
+	 *
+	 * <p>
+	 * Type of charging.
+	 *
+	 * <ul>
+	 * <li>Interface: Evcs
+	 * <li>Readable
+	 * <li>Type: ChargingType @see {@link ChargingType}
+	 * </ul>
+	 */
+	CHARGING_TYPE(Doc.of(ChargingType.values()) //
+		.accessMode(AccessMode.READ_ONLY) //
+		.persistencePriority(PersistencePriority.HIGH)), //
+
+	/**
+	 * Current.
+	 *
+	 * <p>
+	 * The current for all three phases in mA.
+	 * 
+	 * <p>
+	 * See Evcs.initializeCurrentSumChannels() also.
+	 *
+	 * <ul>
+	 * <li>Interface: Evcs
+	 * <li>Readable
+	 * <li>Type: Integer
+	 * <li>Unit: Milliampere
+	 * </ul>
+	 */
+	CURRENT(Doc.of(OpenemsType.INTEGER) //
+		.unit(Unit.MILLIAMPERE) //
+		.accessMode(AccessMode.READ_ONLY) //
+		.persistencePriority(PersistencePriority.HIGH)),
+
+	/**
+	 * CurrentL1.
+	 *
+	 * <p>
+	 * The current for phase L1.
+	 * 
+	 * <p>
+	 * In charge parks the phases of chargepoints are rotated when installed. Thus a
+	 * chargepoint measures current on L1 but it may actually be connected to L2 or
+	 * L3.
+	 * 
+	 * <p>
+	 * CurrentL1 needs to reflect the real phase L1, thus the driver is responsible
+	 * to rotating the phases before updating this channel. Thus the controller
+	 * using this channel can always trust that CURRENT_L1 reflects the current on
+	 * Phase 1.
+	 * 
+	 * <p>
+	 * chargepoint driver developer may use the following helper method:
+	 * this.getPhaseRotation().getFirstPhase()
+	 * 
+	 * <ul>
+	 * <li>Interface: Evcs
+	 * <li>Readable
+	 * <li>Type: Integer
+	 * <li>Unit: Milliampere
+	 * </ul>
+	 */
+	CURRENT_L1(Doc.of(OpenemsType.INTEGER) //
+		.unit(Unit.MILLIAMPERE) //
+		.accessMode(AccessMode.READ_ONLY) //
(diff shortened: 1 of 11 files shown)


io.openems.edge.evcs.api/src/io/openems/edge/evcs/v2/api/ManagedEvcs.java
```diff
@@ -0,0 +1,531 @@
...
+    }
+
+    /**
+     * Gets the priority of the EVCS in [ms]. See {@link ChannelId#PRIORITY}.
+     *
+     * @return the Channel {@link Value}
+     */
+    public default Value<Integer> getPriority() {
+	return this.getPriorityChannel().value();
+    }
```
TimStoecker, line 293: Is the unit [ms] a possible typo?

io.openems.edge.evcs.api/src/io/openems/edge/evcs/v2/api/PhaseRotation.java
```diff
@@ -0,0 +1,71 @@
...
+     * L1 is connect to L2,...
+     */
+    L2_L3_L1(1, "L2_L3_L1", Evcs.ChannelId.CURRENT_L2, Evcs.ChannelId.CURRENT_L3, Evcs.ChannelId.CURRENT_L1), //
+    /**
+     * EVCS which use a rotated hardware connection configuration.
+     * 
+     * <p>
+     * L1 is connect to L3,...
+     */
+    L3_L1_L2(1, "L3_L1_L2", Evcs.ChannelId.CURRENT_L3, Evcs.ChannelId.CURRENT_L1, Evcs.ChannelId.CURRENT_L2);
```
TimStoecker, line 27: Is the value of 1 everywhere a typo? It should be what L1 is physically connected to.
```suggestion
public enum PhaseRotation implements OptionsEnum {
    /**
     * EVCS which use standard hardware connection configuration.
     * 
     * <p>
     * L1 is connect to L1,...
     */
    L1_L2_L3(1, "L1_L2_L3", Evcs.ChannelId.CURRENT_L1, Evcs.ChannelId.CURRENT_L2, Evcs.ChannelId.CURRENT_L3), //
    /**
     * EVCS which use a rotated hardware connection configuration.
     * 
     * <p>
     * L1 is connect to L2,...
     */
    L2_L3_L1(2, "L2_L3_L1", Evcs.ChannelId.CURRENT_L2, Evcs.ChannelId.CURRENT_L3, Evcs.ChannelId.CURRENT_L1), //
    /**
     * EVCS which use a rotated hardware connection configuration.
     * 
     * <p>
     * L1 is connect to L3,...
     */
    L3_L1_L2(3, "L3_L1_L2", Evcs.ChannelId.CURRENT_L3, Evcs.ChannelId.CURRENT_L1, Evcs.ChannelId.CURRENT_L2);
```

io.openems.edge.evcs.api/src/io/openems/edge/evcs/v2/api/Priority.java
```diff
@@ -0,0 +1,47 @@
...
+    LOW(1, "Low"), //
+    /**
+     * EVCS is used for cars with average standing times.
+     */
+    MEDIUM(2, "Medium"), //
+    /**
+     * EVCS should be charged as fast as possible.
+     */
+    HIGH(3, "High") //
+    ;
```
TimStoecker, line 22: I would change MEDIUM to REGULAR or STANDARD or DEFAULT. Since Medium is the default one / the one everyone uses when no priority is required it should be something more readable. Medium seems confusing and SGReady already use REGULAR as a default for consistency.
```suggestion
public enum Priority implements OptionsEnum {
    /**
     * priority is unknown.
     */
    UNDEFINED(-1, "Undefined"), //
    /**
     * EVCS is used for cars with long standing times.
     */
    LOW(1, "Low"), //
    /**
     * EVCS is used for cars with average standing times.
     */
    REGULAR(2, "Regular"), //
    /**
     * EVCS should be charged as fast as possible.
     */
    HIGH(3, "High") //
    ;
```

io.openems.edge.evcs.api/src/io/openems/edge/evcs/v2/api/SocEvcs.java
```diff
@@ -0,0 +1,98 @@
...
+		 * <li>Readable
+		 * <li>Type: Integer
+		 * <li>Unit: Percent
+		 * </ul>
+		 */
+		SOC(Doc.of(OpenemsType.INTEGER) //
+				.unit(Unit.PERCENT) //
+				.accessMode(AccessMode.READ_ONLY) //
+				.persistencePriority(PersistencePriority.HIGH)); //
+
```
TimStoecker, line 34: Is it possible to add something like REMAINING_ENERGY with the unit Wh as a Channel? Just the percentage seems unlikely to be useful.
This channel would indicate how much energy is needed to charge the Battery to 100% and could be used for optimizing the charge duration.
Also a Channel for CURRENT_CAPACITY in Wh would be useful to put in the frontend. A user could therefor do a remaining distance calculation.
As an alternative we could use MAXIMUM_CAPACITY for a basis of calculation like in SymmetricESS.

TimStoecker (changes requested):
I looked over the files and added some comments and suggestions.
Overall it looks really good but I would like to add some details (especially in SocEvcs). 
See Comments below.

io.openems.edge.evcs.api/src/io/openems/edge/evcs/v2/api/Evcs.java
```diff
@@ -0,0 +1,846 @@
...
+
+    /**
+     * Initializes Channel listeners to sum current for L1 + L2 + L3 and add the
+     * value to the current channel.
+     *
+     * @param evcs the {@link Evcs}
+     */
+    public static void initializeCurrentSumChannels(Evcs evcs) {
+	final Consumer<Value<Integer>> currentSum = ignore -> {
+	    evcs._setCurrent(TypeUtils.sum(//
```
wimmerp, line 343: that should not be the sum. if you have a three phase vehicle and the controller will set each phase to "6A", than the sum would be 18A. But internal the charger will have only "6A". that would be a bit confusing, wouldn't it?

io.openems.edge.evcs.api/src/io/openems/edge/evcs/v2/api/EvcsPower.java
```diff
@@ -0,0 +1,20 @@
+package io.openems.edge.evcs.v2.api;
+
+import io.openems.edge.common.filter.RampFilter;
+
+public interface EvcsPower {
```
wimmerp, line 5: very nice to use a Power model!

io.openems.edge.evcs.api/src/io/openems/edge/evcs/v2/api/SocEvcs.java
wimmerp: we should use similar namings to the iso, than it is often easier to understand
https://www.typhoon-hil.com/documentation/typhoon-hil-software-manual/References/iso15118_protocol.html#iso15118_protocol__section_e5d_thl_ctb
energy_capacity

and than you can calculate the rest out of it. 
evtl call it ISO15118EVCS?

wimmerp (commented):
very good. just the "Sumcurrent" might be a challenge?!

io.openems.edge.evcs.api/src/io/openems/edge/evcs/v2/api/EvcsPower.java
sfeilmeier: Ideally I would like to merge the existing EssPower - that does power distribution for energy storage system clusters - with the EvcsPower. Because we eventually need to distribute 'power' (and separately 'energy') among all managed and unmanaged devices.

The EssPower internally uses a linear equation system with linear constraints to model the possibilities and enable power distribution. We had a plan to migrate this linear equation system to OjAlgo, which allows quadratic optimization as well, so that we could write smart(er) target functions for the power distribution than our current manual 'Optimizers' (https://github.com/OpenEMS/openems/tree/develop/io.openems.edge.ess.core/src/io/openems/edge/ess/core/power/optimizers). 

Still... this is a huge, separate topic and should not block the definition of this API.

wimmerp: hi @sfeilmeier, the middle plan was also to use linear equations for the "chargecontrolling", because we really loved that approach. yes quadratic would be amazing, than you could 'punish' but still allow a bit 'overcurrent' and get a better result. i did that in my theses, but the 'runtime' was at that time very long. but if it gets an fast 'solution' that would be fantastic.

sebastianasen:
First of all, I appreciate the collaboration between you guys and the outputs.

I have still a few opened questions:

- I know it is a little bit confusing, that there are multiple information about maximum and minimum evcs Power (MaxPower, MaxHwPower, FixedMaxPower) as you mentioned. This was made like this, because we wanted to have an easy way to show the current possible settings understandable and stable to the customer in the UI. 

grafik

Surely, for charge parks it is not as important as for normal households, but the Nature should be prepared for both, isn't it. It would be great if the UI part has still the possibility, to show the customer the possible minimum and maximum for his charger (depending also on the phases used).
If i have only a dynamic maximum, calculated by the EvcsPower, the Inputs for the user are confusing and not stable (e.g. User Input 10 kW, dynamic maximum 7 kW. Same for the Minimum)

grafik

Internally the Power constraints will anyway act in this ranges and could limit the charging station but still have this limits present and as initial constraints set.

Maybe it would be clearer if the channels would be named different. (Open for suggestions)
**MaximumHardwarePower** (Let's say UiMaximumLimit): The AbstractManagedEvcsComponent or Evcs.addCalculatePowerLimitListeners(this); calculating this value by just setting the real hardware limits e.g. KEBA Dip-Switch Settings [FixedMaximumHardwarePower] and the current phases separately. 
**FixedMaximumHardwarePower**: Set (normally) once, mentioned before
**MaximumPower**: Defined during run time (e.g. Car is charging less than requested) - This could be also the output of the EvcsPower maximum
In this approach, a new evcs must only set the FixedMaximumHardwarePower and the phases.


- Detailed Phase information
 Please feel free to add the channels already in the current evcs api if that gets us ahead. I agree with the PhaseRotation as an important information and the current values as well. 
Even if the three enum values are fine, it could make sense to add the other three possibilities like the "invert" config in a normal meter, if the installer is doing wrong things.

- Priority is not used in the Evcs Nature for now (I assume this the plan)

io.openems.edge.evcs.api/src/io/openems/edge/evcs/v2/api/Evcs.java
opernikus-common: Good point. Replaced the method with an initializer to set current = max (currentL1, currentL2, currentL3). Why max? Chargepoints should have same current on all used phases. But with max() we are always on the safe side.

io.openems.edge.evcs.api/src/io/openems/edge/evcs/v2/api/ManagedEvcs.java
```diff
@@ -0,0 +1,531 @@
...
+	 * <ul>
+	 * <li>Interface: ManagedEvcs
+	 * <li>Read Only
+	 * <li>Type: Integer
+	 * <li>Unit: Milliseconds
+	 * </ul>
+	 */
+
+	TIME_PRECISION(Doc.of(OpenemsType.INTEGER) //
+		.unit(Unit.MILLISECONDS) //
```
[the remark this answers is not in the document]
opernikus-common: Changed Unit from ms to s.

io.openems.edge.evcs.api/src/io/openems/edge/evcs/v2/api/ManagedEvcs.java
opernikus-common: Yes, thanks. Fixed.

io.openems.edge.evcs.api/src/io/openems/edge/evcs/v2/api/PhaseRotation.java
opernikus-common: Absolutely great. Good catch. This might have saved hours of debugging.

io.openems.edge.evcs.api/src/io/openems/edge/evcs/v2/api/SocEvcs.java
opernikus-common: @TimStoecker, @wimmerp, both good points. Also I like the idea of naming it appropriate to ISO15118. I personally would prefer V2gEvcs or somethink like that. Right now I am not aware of a chargepoint providing this information. 
I would leave the interface like it is for now.
AFAIK we will have a bidirectional chargepoint on the hackaton.  We could decide it there.

opernikus-common:
@sebastianasen, 
I am well aware of the solar optimized charge widget. It is very impressive feature of OpenEMS and personally I like it very much. We need to take care that this will work with the new nature!

I removed the Channels nevertheless, because 
1) the XXMin/MaxPower channel name always confuses developers and one need to dive deep into the widget to understand the meaning
2) curently in my opinion this Channels are controller specific and they have nothing to do with the EVCS.

Also I am not lucky with the Widget reconfiguring the EVCSs current (Zoe option). A new driver developer does not know that and that sometimes broke a productive system of us (when a ZOE is charged for the first time). 

Unfortunately right now, I do not have a solution for the solar optimized charge feature, but before we switch to the new API we should take care that the API includes all relevant Channels. 

So we may add your suggestions to the EVCS. But without having looked at the widget closely, I would prefer to put the Channels into the ControllerEvcs. Maybe we can have a talk about this, when we know a little more about the EvcsPower object.

io.openems.edge.evcs.api/src/io/openems/edge/evcs/v2/api/Evcs.java
```diff
@@ -0,0 +1,846 @@
...
+    public default EnumReadChannel getPhaseRotationChannel() {
+	return this.channel(ChannelId.PHASE_ROTATION);
+    }
+
+    /**
+     * Gets the current PhaseRotation. See {@link ChannelId#PHASE_ROTATION}.
+     *
+     * @return the Channel {@link Value}
+     */
+    public default Phases getPhaseRotation() {
```
tsicking, line 701: Should be of type PhaseRotation, not Phases

PR #2056, opened 2023-02-08
Made Modbus ID configurable

SebastianRinger:
The previous code had a hardcoded Modbus ID of 1. We have a Kaco NX3 for testing, which has a Modbus ID of 3. Somehow, the Kaco responds to Modbus TCP calls on Modbus ID 1, so a hardcoded Modbus ID of 1 works with TCP. But on Modbus RTU you need the correct Modbus ID of 3 to get an answer. So the previous code with a hardcoded Modbus ID of 1 does not work for a  Modbus RTU connection.

Simple fix: make the Modbus ID configurabel.


io.openems.edge.pvinverter.kaco.blueplanet/src/io/openems/edge/pvinverter/kaco/blueplanet/Config.java
```diff
@@ -19,6 +19,9 @@
 	@AttributeDefinition(name = "Modbus-ID", description = "ID of Modbus bridge.")
 	String modbus_id() default "modbus0";
 
+	@AttributeDefinition(name = "Modbus Unit-ID", description = "The Unit-ID of the Modbus device.")
+	int modbusUnitId() default 3;
```
sfeilmeier, line 23: We should set a default value of 1 to stay backwards compatible with Modbus/TCP. Maybe add a hint here or in the readme that it has to be set explicitely for Modbus/RTU connections.

io.openems.edge.pvinverter.kaco.blueplanet/src/io/openems/edge/pvinverter/kaco/blueplanet/KacoBlueplanet.java
```diff
@@ -72,7 +72,9 @@ public class KacoBlueplanet extends AbstractSunSpecPvInverter implements SunSpec
 	// .put(DefaultSunSpecModel.S_160, Priority.LOW) // from 40792
 	// .put(SunSpecModel.S_64204, Priority.LOW) // from 40842
 
-	private static final int UNIT_ID = 1;
+	// We had a Kaco nx3 for testing, which had a modbus ID of 3. A static modbus ID of 1 worked for TCP (no idea why),
+	// but not for RTU. To get this plugin working with the Kaco nx3 on a RTU connection, the modbus ID must be configurable.
+	//private static final int UNIT_ID = 1;
```
sfeilmeier, line 77: Once this PR is finished, these comments are not anymore required.

sfeilmeier (changes requested):
Hi Sebastian.

Thanks for the contribution. 

Unit-ID 1 is very common as a default for Modbus/TCP. 

Please add a note in the readme. that this implementation is then compatible with the NX3 as well.  

Our white-list is here: https://fenecon.de/fenecon-fems/fems-app-kaco-pv-wechselrichter/. We will add those to the readme then as well.

io.openems.edge.pvinverter.kaco.blueplanet/src/io/openems/edge/pvinverter/kaco/blueplanet/Config.java
SebastianRinger: Just to be clear, the NX3 has a Modbus ID of 3 for both TCP and RTU. On TCP, it responds to Modbus calls on ID 3 *and* ID 1.  
Is the Modbus ID default value 1 needed for other Kaco devices? I only tested with the NX3, so I have no knowledge on how other Kaco devices operate.

sfeilmeier: I did not test with different Unit-IDs than 1. Anyhow it would (could?) be a breaking change if we now just change it to 3, which sounds dangerous. 

From my point of view:
Either we should get a note from KACO directly whether '3' is always the default for their SunSpec Modbus devices and '1' just happens to work. Or we should document it in the Config.java and/or in the readme, that typically this parameter has to be set to '3'.

SebastianRinger: I agree with your point of view.  
I will go with setting the default to 1, and documenting it in the config & readme ("If 1 does not work, try 3").

sfeilmeier (approved):
Perfect. I adjusted the readme a bit and added our FENECON compatibiilty list. Thanks again for the contribution! 👍

PR #2057, opened 2023-02-09
Fix pv and meter device documentation

clehne:
documentation format of some pv inverter fixed. Added a constraint to SMA SHM20 (not working together with pv inverter only with battery inverters)

### DIFF
--- a/io.openems.edge.meter.sma.shm20/readme.adoc
+++ b/io.openems.edge.meter.sma.shm20/readme.adoc
@@ -12,11 +12,10 @@ Implementation of the SMA Sunny Home Manager SHM 2.0.
 
 == Additional application notes
 
-
 Experiences from testing this plugin:
 
 * The SHM 2.0 can not be accessed directly. It must be connected to a SMA inverter. The SHM2.0 is read out through the SMA inverter. 
-* It does not work with SMA STP 25000TL-30 ( Device Type: 9285 ). Right now it is unclear why STP 15.000 works and STP25.000 not.  
+* It does not work with SMA STP 25000TL-30 ( Device Type: 9285 ). Right now it is unclear why STP 15.000 works and STP25.000 does not work.  
 
  
    
--- a/io.openems.edge.meter.sma.shm20/readme.adoc
+++ b/io.openems.edge.meter.sma.shm20/readme.adoc
@@ -1,16 +1,26 @@
 = SMA Sunny Home Manager 2.0 Integrated Meter
 
-This implementation uses the integrated energy meter of the SMA Sunny Home Manager 2.0
-Data needs to polled through Modbus from an attached SMA battery inverter as the HM2.0 does not have a local API on the device itself
+Implementation of the SMA Sunny Home Manager SHM 2.0.
 
-Tested on::
-- Sunny Home Manager 2.0 with Sunny Island 4/6/8 
+== Compatibility
 
-Note::
-The combination Sunny Home Manager and Sunny Tripower Inverter does not work. SHM seems to work with battery inverters only.   
+** Sunny Home Manager 2.0 (with Sunny Island 4.4M (Device Type: 9279 )
+** Sunny Home Manager 2.0 (with Sunny Island 8.0H (Device Type: 9334 )
+** Sunny Home Manager 2.0 (with SMA STP 15.000TL-20)
 
-Implemented Natures::
-- SymmetricMeter
-- AsymmetricMeter
+
+
+== Additional application notes
+
+
+Experiences from testing this plugin:
+
+* The SHM 2.0 can not be accessed directly. It must be connected to a SMA inverter. The SHM2.0 is read out through the SMA inverter. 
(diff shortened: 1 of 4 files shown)


sfeilmeier:
Thanks for the effort. What do you think about the readme format I used for the KACO PV-Inverter (Compatibility, etc.):
https://github.com/OpenEMS/openems/pull/2056/files#diff-7ea9b39e1464a0b628672b2587791b35c968acbd4c8e9ed4f4eb0d2010171c93

I think it adds value to the generated Antora Docs and makes them more readable: https://openems.github.io/openems.io/openems/latest/edge/device_service.html#_kaco_blueplanet_pv_inverter. Should we apply this format everywhere?

--

Is this related to the problem in this thread also? Or how can we improve the docs in that regard? I do not have much experience with SMA hardware. https://community.openems.io/t/probleme-mit-sunny-home-manager-2-0/1328

Sn0w3y:
Problems using openEMS with Sunny Home Manager 2.0 and an STP10 in the System. I set the TCP Server in the Sunnyportal on Port 502 to on. After this my STP10 had set the MaxActivePower and i could not set it back without resetting it. The Battery which was connected also has disconnected as the EMS worte wrong registers.

Following Setup was used:

Home Manager 2.0 (not functioning Properly as Register addresses are written if the STP10 is also used in the System)
STP 10 SE (does not work)
STP 15.000TL-20 (works with Modbus TCP without Problems on Port 502)
Goodwe GW10K-ET (no Problem here as it is already tested and used a long time)

Important: If the EMS writes too often (1 Cycle Time = 1 Second) it overflows the flash read/write times so it needs to be changed (desoldered and changed) on the Inverter. So it is damaged.

clehne:
@sfeilmeier, very much like the idea of reformatting the documentation and especially adding the compatibility matrix. 
Over the time we should add this compatibility matrix in every bundle. This will help other to see what systems work and what might not work. 
I would like to change this PR but currently struggling on the SMA STP inverters. Because my experience tells different things than @Sn0w3y reported. We tested a system with three STP 25.000 together with a SHM 2.0 on monday. We were not able to read the grid meter values via the inverter. So I assumed that it only works with Sunny Island as we have such a combination running. Now I wonder that it was working with the STP15.000TL-20. I need to dig deeper into this.

Sn0w3y:
> @sfeilmeier, very much like the idea of reformatting the documentation and especially adding the compatibility matrix. Over the time we should add this compatibility matrix in every bundle. This will help other to see what systems work and what might not work. I would like to change this PR but currently struggling on the SMA STP inverters. Because my experience tells different things than @Sn0w3y reported. We tested a system with three STP 25.000 together with a SHM 2.0 on monday. We were not able to read the grid meter values via the inverter. So I assumed that it only works with Sunny Island as we have such a combination running. Now I wonder that it was working with the STP15.000TL-20. I need to dig deeper into this.

We tried it with:

- Single STP 15.000-20 TL (worked even if Home Manager 2.0 was online and Implemented in the System)
- With: 1x STP10 Hybrid Inverter (actually a SMA-Sungrow coorp.) it did not work as the Problems occured as i already wrote above :)

opernikus-common:
> Important: If the EMS writes too often (1 Cycle Time = 1 Second) it overflows the flash read/write times so it needs to be changed (desoldered and changed) on the Inverter. So it is damaged.

@Sn0w3y,  it has probably to do with Sunspec Model 121 (see Model Definition in [SmaPvInverter ](https://github.com/OpenEMS/openems/blob/3c45bde67b96c19010a24a7306bb997c70928ffb/io.openems.edge.pvinverter.sma/src/io/openems/edge/pvinverter/sma/SmaPvInverter.java#L59)) "Basic Settings" and Model S123 "Immediate Control".
The Sunny Tripower Bundle is cyclically writing to S123:WMaxLimPct  and probably to S123:WMaxLimEna.
In General this should be ok, because the SMA Inverter is prepared for this. It should not happen, that you write to S121 cyclically. I think this is what has happened to your STP10. But I don't know why. Except you have explicitly written to those channels. Did you do so?
Maybe it is best for you to comment out those two models (S121 and S123) for now. All other Models are readonly. 

Nevertheless, we should add a configuration option to run STP in readonly mode to be on the safe side. Will look at this.

Sn0w3y:
> Except you have explicitly written to those channels. Did you do so? 

Of course not ! Would be pretty dumb to do so :)

I just tried to Implement it in my EMS but i guess it would be the best to add a configuration for the STP (Hybrid Inverters) to be safe yes ! Btw. is there any Documentation to search for the Register addresses as SMA does not provide anything for Modbus TCP Addresses to Read or Write ?

clehne:
@Sn0w3y, SMA provides Sunspec. So you may have a look at  https://sunspec.org/specifications/. They have very detailed descriptions of everything. Still I did not understand, why your STP10 gets broken. We have one up and running without any problems. 

@sfeilmeier, I modified the docs as request. You may close this PR. The readonly functionality for STP inverters should be handled in another PR.

da-Kai:
Close as requestsed by @clehne 

> @sfeilmeier, I modified the docs as request. You may close this PR. The readonly functionality for STP inverters should be handled in another PR.

PR #2058, opened 2023-02-09
Two small bugfixes in VirtualMeterAdd

tsicking:
a) Consumption Energy is now calculated correctly (remained null before)
b) Min- and max active powers are automatically derived from active power and should not be calculated by the meter. (The sum of the max powers is not necessarily the max power of the sum anyway)


clehne (approved):
Looks good to me.

sfeilmeier (approved):
As the improvements in #2062 are not going to be merged very soon, I'll add this bugfix now.

PR #2059, opened 2023-02-09
AbstractSunSpecMeter Energy value fix

tsicking:
Energy values were mapped twice in AbstractSunspecMeter, once with the wrong converter.


sfeilmeier (approved):
Thank you for the fix! (what a stupid mistake... 😒)

sfeilmeier:
That would be great. It was anyway good to separate the bugfix from a new feature. These small PRs are really easy to review&merge from my side.

PR #2060, opened 2023-02-10
Energy per phase added to AbstractSunSpecMeter

tsicking:
As discussed in the previous pull request, energy per phase is added to the abstract sunspec meter.


sfeilmeier (approved):
Thank you for the improvement!

PR #2061, opened 2023-02-13
Adds maven wrapper.

hydroid7:
Encapsulates Maven to a dedicated wrapper. 
It is the way to go with Gradle and Maven should follow too.


sfeilmeier:
Thank you for the contribution! 

Could you share some docs (here as a comment or in a readme file) on you to use mvnw to build OpenEMS (Edge/Backend)?

hydroid7:
Sure! 
The maven wrapper is required to run maven on systems where 
- no maven or 
- an older version of maven is installed.

The addition of the wrapper does not change the build system and has the following benefits:
1. No need for installed maven on the system
2. Guarantee that always the right version of maven is used
3. Maven version is stored in Git and therefore updates can be done "in code".

PR #2072, opened 2023-02-22
Backend Alerting: bugfixes and improvements

sfeilmeier:
- Add Executor with one Thread
- Handle Events in Queue
- Check validity again just before sending an alert (when queue is very long)
- Fix typos/naming


da-Kai (approved):
👍

PR #2077, opened 2023-02-23
add docs for intellij

wimmerp:
added some documentation for usage of openems with intellij.


DerStoecki:
NOTE:
I could NOT compile on Version 2022.2.4 with the newest amdatu plugin (which is strange)
however: 
version 2021.2.4 works fine! (Amdatu Version 3.0.1)
It's either a problem with my setup or with the amdatu verison itself / bnd.bnd i have to look into this, but i just wanted to leave this here.

sfeilmeier (approved):
Well done, Paul! :-)

PR #2079, opened 2023-02-24
Implemented Wallbe Evcs based on old Consolinno Version as BETA

TimStoecker:
Implemented the Wallbe EVCS during the Hackathon based on the old Consolinno version as a BETA. Testing is not possible at the moment but will be done at a later date.

### DIFF
--- a/io.openems.edge.evcs.wallbe/src/io/openems/edge/evcs/wallbe/api/Wallbe.java
+++ b/io.openems.edge.evcs.wallbe/src/io/openems/edge/evcs/wallbe/api/Wallbe.java
@@ -1,0 +1,297 @@
+package io.openems.edge.evcs.wallbe.api;
+
+import io.openems.common.channel.AccessMode;
+import io.openems.common.channel.Unit;
+import io.openems.common.exceptions.OpenemsError;
+import io.openems.common.types.OpenemsType;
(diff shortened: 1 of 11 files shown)


io.openems.edge.evcs.wallbe/src/io/openems/edge/evcs/wallbe/api/Wallbe.java
```diff
@@ -0,0 +1,297 @@
...
+
+    /**
+     * Gets the Value of {@link ChannelId#CURRENT_L1}.
+     *
+     * @return the value
+     */
+    default int getCurrentL1() {
+        Channel<Integer> channel = this.getCurrentL1Channel();
+        return channel.value().orElse(channel.getNextValue().orElse(0));
+    }
```
tsicking, line 199: I would omit the `orElse()`'s and change the return type to `Value<Integer>`, like it's done in other interfaces.
Maybe add `public` to make it clearer, and write it in one line `this.getCurrentL1Channel().value()`.
Same for the other getters. 
Those getters that are not needed (if there are any) somewhere, you can also leave out in my opinion.

tsicking (changes requested):
Thanks for implementing. See my review comments for the requested changes.

PR #2087, opened 2023-02-27
Adds Hardware Access to OpenEMS

hydroid7:
As the title states, Pi4J should be integrated into OpenEMS.

OpenEMS Community discussion: https://community.openems.io/t/hardware-abstraction-layer-fur-openems-pi4j-revpi-etc/1349


hydroid7:
The component does not longer use Pi4J.

PR #2097, opened 2023-03-09
OpenEMS UI: New widgets (Live & History) for Thermometer nature.

laupp:


clehne:
The new Thermometer widget lists all component which fulfills the Thermometer nature. This is not a good practice. E.g. the consolinno temperature hw module supports 16 temperatures.  Thus the Thermometer widget would show all 16 temperatures. In most cases only a few temperature sensors are of relevance for showing in the UI. 
We need to find a way how to configure which temperatures are shown within the UI. 
For example the Controller IO Fix digital output widget is only shown, if such a controller exists. It doesn't matter how many relays are there. Maybe we could add a very simple ControllerThermometer and show the temperature only if such a controller is active. But the name ControllerThermometer is bad, because there is no Controller functionality.
Someone an idea how we can do the selection which temperatures to show and which not to show?

ui/src/app/edge/history/history.module.ts
```diff
@@ -143,6 +146,9 @@ import { TimeOfUseTariffDischargeWidgetComponent } from './timeofusetariffdischa
     SymmetricPeakshavingChartComponent,
     SymmetricPeakshavingChartOverviewComponent,
     SymmetricPeakshavingWidgetComponent,
+    ThermometerChartComponent,
+    ThermometerChartOverviewComponent,
+    ThermometerWidgetComponent,
```
lukasrgr, line 151: wrapping these components into modules is recommended. Improves readability and compactness

ui/src/app/edge/live/thermometer/thermometer.component.ts
```diff
@@ -0,0 +1,35 @@
...
+import { AbstractFlatWidget } from 'src/app/shared/genericComponents/flat/abstract-flat-widget';
+import { formatNumber } from '@angular/common';
+
+@Component({
+  selector: 'Thermometer',
+  templateUrl: './thermometer.component.html'
+})
+export class ThermometerComponent extends AbstractFlatWidget {
+
+  public sensors: EdgeConfig.Component[] = null;
```
lukasrgr, line 12: template variables can be protected with angular 14. https://twitter.com/angular/status/1537181658461265920

Strict nullchecks also require this variable to be of a union type "EdgeConfig.Component[] | null"

ui/src/app/edge/live/thermometer/thermometer.component.ts
```diff
@@ -0,0 +1,35 @@
...
+  public readonly CONVERT_TO_Degree = Utils.CONVERT_TO_Degree;
```
lukasrgr, line 13: can you change this Callback into fully Uppercase? This is how we did it for the other Converters as well

ui/src/app/edge/live/thermometer/thermometer.component.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    for (let sensor of this.sensors) {
+      channelAddresses.push(
+        new ChannelAddress(sensor.id, 'Temperature')
+      )
+    }
+    return channelAddresses;
+  }
+
+  protected override onCurrentData(currentData: CurrentData) {
```
lukasrgr, line 32: you do not have to implement this method, if you dont use it

ui/src/app/edge/live/thermometer/thermometer.component.html
```diff
@@ -0,0 +1,11 @@
+<ng-container *ngIf="isInitialized">
+    <oe-flat-widget img="assets/img/thermometer.png" [title]="'General.temperature' | translate">
+
+        <ng-container *ngFor="let sensor of sensors, let i = index">
+            <oe-flat-widget-line [name]="sensor.alias" [channelAddress]="sensor.id + '/Temperature'"
```
lukasrgr, line 5: If you use the [channelAddress]-Input, you do not need to subscribe the channel in getChannelAddresses() in the .ts-file

lukasrgr (changes requested):

ui/src/app/edge/history/thermometer/widget.component.ts
```diff
@@ -0,0 +1,29 @@
...
+    selector: ThermometerWidgetComponent.SELECTOR,
+    templateUrl: './widget.component.html'
+})
+export class ThermometerWidgetComponent implements OnInit {
+
+    @Input() public period: DefaultTypes.HistoryPeriod;
+
+    private static readonly SELECTOR = "thermometerWidget";
+
+    public edge: Edge = null;
```
lukasrgr, line 16: Union type is recommended for strict null checks https://www.typescriptlang.org/docs/handbook/advanced-types.html#optional-parameters-and-properties

lukasrgr (changes requested):

sfeilmeier:
> The new Thermometer widget lists all component which fulfills the Thermometer nature. This is not a good practice. E.g. the consolinno temperature hw module supports 16 temperatures. Thus the Thermometer widget would show all 16 temperatures. In most cases only a few temperature sensors are of relevance for showing in the UI. We need to find a way how to configure which temperatures are shown within the UI. For example the Controller IO Fix digital output widget is only shown, if such a controller exists. It doesn't matter how many relays are there. Maybe we could add a very simple ControllerThermometer and show the temperature only if such a controller is active. But the name ControllerThermometer is bad, because there is no Controller functionality. Someone an idea how we can do the selection which temperatures to show and which not to show?

@clehne 
(1) I have been thinking about introducing an Edge Component that holds UI configuration (e.g. "Core.UI"). This component could hold a JSON configuration that is used to parameterize the UI, e.g. which Widgets should be shown, different ording of the widgets on the screen, etc.. Benefit of using an Edge Component is, that there is already a way to transfer the information via EdgeConfig and it would work both with Edge and Backend. Downside is, that there would be only one config per Edge, so it is not directly supported to have multiple configs e.g. per User. Example:


```
{
   "widgets":[
      {
         "widget":"ThermometerWidget",
         "componentId":"thermometer0",
         "hide":true
      }
   ]
}
```

(2) Of course, the ControllerThermometer idea is more like the existing architecture and should be followed if possible, because it puts the Temperature value in a context (e.g. in the SingleThresholdController, etc.)

(3) Even another idea we discussed at the Hackathon, was the introduction of a Widget that allows defining an SVG (e.g. via Inkscape) and then inserts the Channel-Values (e.g. Temperature values) in placeholders. This could be done similarly and additiionally to idea (1) above and would also make it possible to put Temperature in context (e.g. visualize an actual floor plan or hot water boiler)

```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100">
  <path d="M30,1h40l29,29v40l-29,29h-40l-29-29v-40z" stroke="#000" fill="none"/> 
  <path d="M31,3h38l28,28v38l-28,28h-38l-28-28v-38z" fill="#a23"/> 
  <text x="50" y="68" font-size="48" fill="#FFF" text-anchor="middle">{{ thermometer0/Temperature }}</text>
</svg>
```
(Source is not relevant, just for completeness: https://dev.w3.org/SVG/tools/svgweb/samples/svg-files/410.svg)

(4) Similarly to idea (3), a Config could also parameterize `oe-modal-line`s

```
{
   "lines":[
      {
         "title":"My Temperature",
         "channel":"thermometer0/Temperature"
      }
   ]
}
```

lukasrgr:
@laupp @DerStoecki any update on this branch?

PR #2100, opened 2023-03-12
Improve Shelly Implementation

nicoketzer:
Improve Shelly Implementation by using the Generic HTTP Worker.

As noted in #2098 here is another use of the Generic HTTP Worker.

### What's the Improvement?
At the moment, the Shelly Implementation block´s the "handleEvent()" and so block´s the Main OpenEms-Cycle.
With this implementation of the Shelly, each component has its own worker, so the OpenEMS cycle no longer blocks.

I have tested this implementation for a long time on my energy storage system.

In my system are currently 9 Shelly's integrated, where each has about 200ms response time to the "/status" query. So each cycle was blocked only by these queries for at least 1,8s (as soon as one fails the timeout of 5s per Shelly takes effect --> 45s no new OpenEMS cycle in worst case).

With this implementation, this problem has now been solved. The data still gets into the OpenEMS just as reliably.

The Common Shelly API is no longer needed.

It would be possible to outsource common components to an abstract class, but the effort to pass the triggerNextRun() event of the worker via the main component to the abstract component to the worker would be disproportionate.

As noted in #2098 and #2099 this is Based on the Generic HTTP Worker. If the Main Pull (#2098) get changed, the Changes would also be transferred to this pull.


hydroid7:
Hi @nicoketzer,
could you please take a look at my pull request at #2233? It should provide a simple and yet flexible HTTP bridge.

PR #2111, opened 2023-03-20
Added option for use without an ESS attached

DerWahreKlinki:
One can configure "NONE" for ess-id. In that case SoC is not used for calculating charge power


sfeilmeier (commented):
Thank you @DerWahreKlinki for the suggestion. I changed the implementation and solved it differently after reading through the current source code.

@sebastianasen: Could you have a look please? I split the changes in 'understandable small' parts, see https://github.com/OpenEMS/openems/pull/2111/commits

sebastianasen (approved):
Thank you for the improvements. I also like the approach of the static calculation methods.

sfeilmeier (approved):

PR #2116, opened 2023-03-21
Include keys in components object in edge config

jpdark:
The id is a property of the model that is represented by each object, so including it in that representation is desirable.
Same goes for the channels when included in the components.

This benefits (de)serialization in clients for the edge config api.

### DIFF
--- a/io.openems.common/src/io/openems/common/types/EdgeConfig.java
+++ b/io.openems.common/src/io/openems/common/types/EdgeConfig.java
@@ -517,6 +517,7 @@ public class EdgeConfig {
 				properties.add(property.getKey(), property.getValue());
 			}
 			var result = JsonUtils.buildJsonObject() //
+					.addProperty("id", this.getId()) //
 					.addProperty("alias", this.getAlias()) //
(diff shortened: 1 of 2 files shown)


hydroid7:
Hi there,
the changes look ok to me.

Can you please elaborate a bit why this change is useful?
Do you plan any feature that depend on the `id`s?

Increasing the API surface should be done only after a discussion :)

jpdark:
@hydroid7 Would love to hear what you think. Keen to merge (or close) this PR.

hydroid7 (changes requested):
Hi,
thank you for pushing this 😄 . IMHO the PR is a little bit small now.
Please also document your changes in the docs for other to be able to find it.

jpdark:
Thanks @hydroid7. I'm not 100% sure what you are requesting. My understanding is that you would like:
- a larger PR (🤔 ?)
- additional documentation for `getEdgeConfig` rpc

I hope to address both with 2 additional commits.
Thanks.

hydroid7 (approved):

hydroid7:
Thank you for the docs, it looks good.

PR #2118, opened 2023-03-23
UI: add eslint rule

lukasrgr:
Added Eslint rule for explicit member accessibility. Set it to warn, because there are many errors to fix


sfeilmeier:
Thanks Lukas. Nice improvement. Should we add all/some of the fixes for the warnings to this PR as well?

sfeilmeier (approved):

PR #2122, opened 2023-03-28
Refactor streams to use toList() method

SalmaElKhalfaoui:
With Java 17 `.collect(Collectors.toList())` can be written as `.toList()`


sfeilmeier:
@michaelgrill: in App-Center you currently require modifiable lists. The new `toList()` method produces unmodifiable Lists and breaks JUnit tests. Do you want to refactor or should we keep the old Collectors where required?

io.openems.edge.core/src/io/openems/edge/core/appmanager/dependency/AppManagerAppHelperImpl.java
```diff
@@ -1066,8 +1066,7 @@ private Optional<OpenemsAppInstance> findNeededApp(DependencyDeclaration declara
 		}
 		var appId = config.appId;
 		if (declaration.createPolicy == DependencyDeclaration.CreatePolicy.ALWAYS) {
-			var neededApps = this.getAllInstances().stream().filter(t -> t.appId.equals(appId))
-					.collect(Collectors.toList());
+			var neededApps = this.getAllInstances().stream().filter(t -> t.appId.equals(appId)).collect(Collectors.toList());
```
michaelgrill, line 1069: you can also use `toList()` here or remove the change because its only formatting

michaelgrill (changes requested):

michaelgrill (approved):
👍very good work

sfeilmeier (approved):

PR #2126, opened 2023-03-30
Some Improvements in the UI

Sn0w3y:
I optimized the changelog.component.ts so it is configurable without a need to compile it again from the /src/assets/changelog.json here.

Optimized the Validators in the Registration Form including a Mail Validator Function !


Sn0w3y:
I created more Validators because i guess its a good manner to check the fields according to their usetype. Before you could add: 

In the Mail Field: test@test. and it would be recognized as a mail.

This could lead to Spam Registrations in a bad Use of the Registration Form.

The Password Validator checks if the Password is Strong enough [in Terms of the Length] so that it is not easily guessable by a 3rd Party.

The other Fields are Checked according to their Length aswell as a Name or a Lastname can not be only one Letter.

Regarding the Changelog:

The Changelog has been added so that the Application does not need to be compiled again and again in Times of Updating the Changelog itself.

Regards.

lukasrgr:
@hydroid7 did someone request these changes?

Sn0w3y:
> @hydroid7 did someone request these changes?

Well actually if not needed just delete it. I thought this is a open-Source Project to improve and work together :D 

Just delete it then. I will not bother you again and contribute my "improvements" :)

lukasrgr:
@Sn0w3y didnt mean to scare you off, i will approve of it

lukasrgr (commented):
Approved

sfeilmeier:
@Sn0w3y: thank you for your ideas and active contribution. I'll review it in a minute.

@lukasrgr: thanks for having a look. Still we need to assure that changes don't break anything before we approve.

ui/src/app/registration/modal/modal.component.ts
```diff
@@ -6,6 +6,12 @@ import { RegisterUserRequest } from 'src/app/shared/jsonrpc/request/registerUser
 import { Service, Websocket } from 'src/app/shared/shared';
 import { environment } from 'src/environments';
 
+function emailValidator(control: FormControl): { [s: string]: boolean } | null {
+  const emailRegex = /^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$/; // Regex to check for valid email format
```
sfeilmeier, line 10: Validating an email with regex is unfortunately very complicated - see e.g. this thread on StackOverflow: https://stackoverflow.com/questions/201323/how-can-i-validate-an-email-address-using-a-regular-expression

After briefly reading through the thread I think your implementation is currently incomplete. Can we not reuse the `Validators.email` as before?

ui/src/app/registration/modal/modal.component.ts
```diff
@@ -96,30 +102,31 @@ export class RegistrationModalComponent implements OnInit {
...
-        city: new FormControl("", Validators.required),
-        country: new FormControl("", Validators.required),
-        phone: new FormControl("", Validators.required),
-        email: new FormControl("", [Validators.required, Validators.email]),
-        password: new FormControl("", Validators.required),
-        confirmPassword: new FormControl("", Validators.required),
+        companyName: new FormControl("", [Validators.required, Validators.maxLength(50)]),
+        firstname: new FormControl("", [Validators.required, Validators.maxLength(50)]),
+        lastname: new FormControl("", [Validators.required, Validators.maxLength(50)]),
+        street: new FormControl("", [Validators.required, Validators.maxLength(50)]),
```
sfeilmeier, line 108: I am not sure we should be so strict. There are again a lot of discussions on that topic on the internet (e.g. https://ux.stackexchange.com/questions/55529/what-should-the-character-limits-for-first-last-name-inputs-be)

While I think it is indeed feasible to have a limit, I think 50 characters could be too limiting in a few cases practice. How about 100?

ui/src/app/registration/modal/modal.component.ts
```diff
@@ -96,30 +102,31 @@ export class RegistrationModalComponent implements OnInit {
...
+        zip: new FormControl("", [Validators.required, Validators.pattern('^[0-9]{5}$')]),
```
sfeilmeier, line 109: This requires 5 digits, right? That won't work e.g. for Austria. Can we find a better regex - ideally including the source webpage/discussion?

ui/src/app/registration/modal/modal.component.ts
```diff
@@ -96,30 +102,31 @@ export class RegistrationModalComponent implements OnInit {
...
+        city: new FormControl("", [Validators.required, Validators.maxLength(50)]),
+        country: new FormControl("", [Validators.required, Validators.maxLength(50)]),
+        phone: new FormControl("", [Validators.required, Validators.pattern('^[0-9]+$')]),
```
sfeilmeier, line 112: This won't allow `+49 (0) ...` format, etc.

ui/src/app/registration/modal/modal.component.ts
```diff
@@ -96,30 +102,31 @@ export class RegistrationModalComponent implements OnInit {
...
+        email: new FormControl("", [Validators.required, emailValidator, Validators.maxLength(50)]),
+        password: new FormControl("", [Validators.required, Validators.minLength(8), Validators.maxLength(20)]),
+        confirmPassword: new FormControl("", [Validators.required, Validators.minLength(8), Validators.maxLength(20)]),
+        Installateur: new FormControl("", [Validators.required, Validators.maxLength(50)]),
```
sfeilmeier, line 116: It's not part of your Pull Request, but this should actually be named `installer`...

ui/src/assets/changelog.json
sfeilmeier, line 1: This file actually belongs to #2125. I still feel that we should not do the change, because our current format allows more possibilities than your proposed solution. I had removed the actual FEMS changelog file, so this was not obvious unfortunately.

In FEMS changelog we e.g. use logs like this:
```
      {
        version: '2023.3.6',
        changes: [
          "App Center: öffentliches Release",
          "Link zu " + Changelog.link("FAQ", 'https://fenecon.de/faq/#fems') + " korrigiert",
          Changelog.product(...Product.FEMS_ALL_TIME_OF_USE_TARIFF) + "Verbesserung der Datenanzeige in der Live-Ansicht"
        ]
      },
```

This uses a definition in changelog.constants.ts:

```
    public static readonly FEMS_AWATTAR = new Product('FEMS-App Awattar HOURLY', 'https://fenecon.de/produkte/fems/fems-app-awattar-hourly/');
    public static readonly FEMS_CORRENTLY = new Product('FEMS-App STROMDAO Corrently', 'https://fenecon.de/produkte/fems/fems-app-stromdao-corrently/');
    public static readonly FEMS_TIBBER = new Product('FEMS-App Tibber', 'https://fenecon.de/produkte/fems/fems-app-tibber/');
    public static readonly FEMS_ALL_TIME_OF_USE_TARIFF = [Product.FEMS_AWATTAR, Product.FEMS_CORRENTLY, Product.FEMS_TIBBER];
```

I don't see a clean way of solving this using a plain JSON file instead. Additionally the current approach shows errors immediately on compile time, which is not possible using JSON.

---

From my understanding the benefit of using a separate JSON file, that is loaded dynamically at runtime is, that the OpenEMS UI would load faster as long as the Changelog is not opened - which would be nice to have of course, especially as the Changelog won't get shorter in time... Maybe the technically better solution would be to use the same format as we do now (see the benefits above), but apply lazy loading -> https://angular.io/guide/lazy-loading-ngmodules. What do you think?

sfeilmeier (changes requested):

sfeilmeier:
@Sn0w3y: Again... did not mean to scare you off, just need to make sure to not break anything - e.g. registration of installers from Austria, etc.; your suggestions and improvements are very much welcomed!

hydroid7:
@Sn0w3y To give you a bigger picture, merging small patches is always a big headcache, since we want to ensure that the change does not break the workflow of any other OpenEMS user, als @sfeilmeier also mentioned. 

Validating user input is one of the hardest things around here, so the change is about a few lines, but in reality there is a lot behind it, as you can see from the change requests 👍 

The best summary, why small changes are easy to write and hard to merge, gives Linus Torvalds about "trivial patches" regarding the linux kernel: https://lkml.org/lkml/2004/12/20/255

Best

PR #2137, opened 2023-04-12
KEBA EVCS: cleanup

sfeilmeier:
- Split impl and interface
- Remove unnecessary 'Alias' channel


sfeilmeier:
@sebastianasen: I've been trying to update the KEBA code to latest 'best practices' without changing any logic. There would be more room for improvement, but I belive this is a first good step in the right direction. Please have a look.

sebastianasen (approved):
Thanks for the clean up.
As you said, this does not break anything.
FYI: For the evcs controller I have already done this similar to your changes comming with a new hysteresis functionality. As this has not been tested in live systems, I may push these general changes separately.

PR #2138, opened 2023-04-12
ESS Fix Active Power Controller: enhancements

sfeilmeier:
- Add _Phase_ and _Relationship_ configuration settings
- Improve internal log messages (shows Controller-ID)
- Extend JUnit test
- Drop now obsolete Controller.Asymmetric.FixActivePower


io.openems.edge.controller.ess.fixactivepower/test/io/openems/edge/controller/ess/fixactivepower/EssFixActivePowerImplTest.java
```diff
@@ -48,15 +45,18 @@ public void testOff() throws OpenemsException, Exception {
...
-						.build()) //
-				.next(new TestCase() //
-						.output(ESS_SET_ACTIVE_POWER_EQUALS, null));
+						.setPhase(Phase.ALL) //
+						.setRelationship(Relationship.EQUALS) //
+						.build()); //
 	}
 
 	@Test
 	public void testGetAcPower() throws OpenemsException, Exception {
```
sebastianasen, line 54: A new test for Relationship and a different Phase configuration would be great, but otherwise everything is fine.

sebastianasen (approved):
See my comments below.
We should also consider in future whether it makes sense to update the UI widget according to the two new configuration properties.

PR #2144, opened 2023-04-16
Introduce ElectricityMeter as replacement for SymmetricMeter & AsymmetricMeter

sfeilmeier:
Distinguishing between `SymmetricMeter` and `AsymmetricMeter` and having `AsymmetricMeter extends SymmetricMeter` did never really make sense physically. In fact the _asymmetric_ meter (i.e. the three-phase meter) is the _most normal_ case. Special cases are:
- a _symmetric_ meter, i.e. one that has the same values on all phases; this can be handled entirely by deriving the individual phase values from the sum values, via dividing the power by three.
- a _single-phase_ meter, i.e. one that only has values on one phase. This requires active configuration and then allows setting the other phases to zero. The `SinglePhaseMeter` Nature is kept, because it allows a use-case when UI should really only show one value for these devices.

The special cases can be represented properly by this newly introduced `ElectricityMeter`, which combines most of the Channels and replaces SymmetricMeter and AsymmetricMeter entirely. It also clearly distinguishes _electricity meters_ from _heat meters_ that are slowly coming to OpenEMS.

This PR includes:
- Replace SymmetricMeter and AsymmetricMeter
- Breaking:
  - Renamed Meter.Virtual.Asymmetric.Add -> Meter.Virtual.Add
  - Dropped Meter.Virtual.Symmetric.Add
  - Renamed Meter.Virtual.Symmetric.Subtract -> Meter.Virtual.Subtract
- Simulator Grid Meter Reacting: add Voltage and Current sum and per phase for better testing
- UI fixes
  - show negative current per phase
  - fix translation issues

It also tries to clarify the sometimes confusing representation of positive and negative values for Power and Current. See the Javadoc in ElectricityMeter and the discussion at https://community.openems.io/t/change-energy-channel-assignment-based-on-meter-type/1603/6.

As this PR touches a lot of files, it is kept as small, simple and non-breaking as possible. Follow-up PRs will add more features, like
- Make EVCS behave like a ConsumptionMeter, i.e. `Evcs extends ElectricityMeter` and replace `CHARGE_POWER` with `ACTIVE_POWER`. This will allow us to drop the special handling of Evcs ChargePower in UI and handle the EVCS power as 'Controlled/Managed' in Edge.
- Introduce a `_sum/UnmanagedConsumptionActivePower` for all Consumption that is not actively managed by OpenEMS Edge. This information is required for better prediction accuracy.
- Many device implementations are incomplete for individual phase information. These will be added in individual PRs (as required by our [Coding Guidelines](https://openems.github.io/openems.io/openems/latest/contribute/coding-guidelines.html#_general))
- When looking through the Code I came across a lot of 'old' Code that does not comply with our latest best practices and does not use Java 17 features. These will be tackled in individual PRs as well - see #2142, #2141 and #2122 for examples. 

**Contributions like this are very welcomed and are perfect tasks for newcomers. Please get in touch if you want to actively take over such a task!**


sfeilmeier:
@sebastianasen I would ask you to have a first look. Later in the process I would like to involve @huseyinsaht and @lukasrgr for the details. Thanks in advance.

io.openems.edge.meter.api/src/io/openems/edge/meter/api/ElectricityMeter.java
```diff
@@ -2,136 +2,211 @@
...
+ * 
+ * <p>
+ * This is the parent of everything that measures electricity flow. Consider
+ * using:
  * <ul>
- * <li>Negative ActivePowerL1/L2/L3 and ConsumptionActivePowerL1/L2/L3 represent
- * Consumption, i.e. power that is 'leaving the system', e.g. feed-to-grid
- * <li>Positive ActivePowerL1/L2/L3 and ProductionActivePowerL1/L2/L3 represent
- * Production, i.e. power that is 'entering the system', e.g. buy-from-grid
+ * <li>{@link SymmetricMeter}: if L1, L2 and L3 are always equal
```
sebastianasen, line 31: I assume the SymmetricMeter will be removed completely, right?
Maybe it makes sense to change the description for symmetric systems, to use the total amount divided by three

ui/src/app/edge/live/common/grid/modal/modal.html
```diff
@@ -11,15 +11,11 @@
...
+        <oe-modal-line *ngIf="meters.length > 1" [name]="meter.alias ?? meter.id"
+            [channelAddress]="meter.id + '/ActivePower'" [converter]="Utils.CONVERT_TO_WATT">
         </oe-modal-line>
-
-        <!-- If meter is asymmetric, show Phases -->
-        <ng-container *ngIf="meter.isAsymmetric">
-            <oe-modal-meter-phases [component]="meter.component" [setTranslatedName]="setTranslatedName">
-            </oe-modal-meter-phases>
-        </ng-container>
+        <oe-modal-meter-phases [component]="meter" [setTranslatedName]="setTranslatedName">
```
sebastianasen, line 17: Phases are not shown for now.
I am not checking every meter type for now.

This is a UI topic more related to @lukasrgr anyway. 


grafik

io.openems.edge.meter.api/src/io/openems/edge/meter/api/SinglePhaseMeter.java
```diff
@@ -25,39 +25,12 @@ public Doc doc() {
...
 	public SinglePhase getPhase();
 
-	/**
-	 * Initializes Channel listeners. Copies the Active-Power Phase-Channel value to
-	 * Active-Power Channel.
-	 *
-	 * @param meter the {@link AsymmetricMeter}
-	 * @param phase the {@link SinglePhase}
-	 */
-	public static void initializeCopyPhaseChannel(AsymmetricMeter meter, SinglePhase phase) {
```
sebastianasen, line 41: Was this never used?

sebastianasen:
For now, I have had a look at the natures and some random implementations in the edge part and a minor one in the UI part without looking deeply into the code. 
I also think that this is a more correct way of defining a meter measuring electricity.

ui/src/app/edge/live/common/grid/modal/modal.html
sfeilmeier: Problem was existing before already. Fixed in https://github.com/OpenEMS/openems/pull/2144/commits/9f98fc08f2200beb4c9ebbbb50ee7bab7afc10c9

io.openems.edge.meter.api/src/io/openems/edge/meter/api/SinglePhaseMeter.java
sfeilmeier: Apparently not :-)

For a new implementation I would like to refactor this feature. For the old method the actual phase had to be known from the beginning. But for some devices this might only be found during execution (e.g. by reading from Modbus). A new feature should be able to handle this dynamically. But for now it is not needed immediately.

sfeilmeier:
@huseyinsaht Could you have a look as well, please? (from a electrotechnical and Java-developer perspective)

io.openems.edge.meter.api/src/io/openems/edge/meter/api/ElectricityMeter.java
```diff
@@ -153,21 +223,30 @@ public enum ChannelId implements io.openems.edge.common.channel.ChannelId {
...
-		 * <li>Unit: mV
+		 * <li>Type: {@link OpenemsType#INTEGER}
+		 * <li>Unit: {@link Unit#MILLIVOLT}
 		 * </ul>
 		 */
 		VOLTAGE_L3(Doc.of(OpenemsType.INTEGER) //
 				.unit(Unit.MILLIVOLT) //
 				.persistencePriority(PersistencePriority.HIGH)), //
+		/**
+		 * Current.
```
clehne, line 234: I am always unsure about the Current. What is it?
Is it max (l1,l2,l3) or is it mean(l1,l2,l3)? Right now  I do not see an added value by providing the CURRENT channel. 
Is it really needed? If so, may be we can have a more in depth textual explanation.

clehne (changes requested):
Good work. I like the idea of replacing AsymmetricMeter/SymmetricMeter by EnergyMeter much. 
Interface looks clean and structured to me. Just did not understand the CURRENT channel. Maybe you can get clearer on it (see my comment there).

io.openems.edge.meter.api/src/io/openems/edge/meter/api/ElectricityMeter.java
sfeilmeier: Thanks @clehne. The problem applies to both `CURRENT` and `VOLTAGE`. I added these values initially, because the first devices where not even three-phased and these values are provided by the devices. Example: SOCOMEC e24 (https://www.socomec.com/files/live/sites/systemsite/files/SCP/6_gestion_energie/countis/countis_e/Countis_E28_48503055.html) registers 0xC588 and 0xC58C.

In fact it might be better to drop both `CURRENT` and `VOLTAGE` completely. Question is, if this should be done in this PR as well or in a separate one.

clehne: Thanks for the explanation. As this PR already includes a lot of files I would suggest to handle the CURRENT and VOLTAGE drop in a separate PR.

clehne (approved):

sfeilmeier:
@lukasrgr Please also have a look

ui/src/app/edge/live/common/consumption/flat/flat.ts
```diff
@@ -28,20 +28,16 @@ export class FlatComponent extends AbstractFlatWidget {
...
-    this.consumptionMeters = this.config.getComponentsImplementingNature("io.openems.edge.meter.api.SymmetricMeter")
+    this.consumptionMeters = this.config.getComponentsImplementingNature("io.openems.edge.meter.api.ElectricityMeter")
       .filter(component => component.isEnabled && this.config.isTypeConsumptionMetered(component));
 
     for (let component of this.consumptionMeters) {
       channelAddresses.push(
         new ChannelAddress(component.id, 'ActivePower'),
+        new ChannelAddress(component.id, 'ActivePowerL1'),
+        new ChannelAddress(component.id, 'ActivePowerL2'),
+        new ChannelAddress(component.id, 'ActivePowerL3'),
```
lukasrgr, line 39: i didnt read into the edge part, but the electricity meter is always writing phasechannels?

lukasrgr (commented):
like this approach.

PR #2147, opened 2023-04-19
PV-Inverters: add read-only mode

sfeilmeier:
Currently all `ManagedSymmetricPvInverter`s set the power limitation - even if in most cases this is not required, because no PV-Limit controller is active. In these cases currently a power limit to 100 % is set via Modbus (i.e. 'no limit'). PV-Inverters often distinguish between Modbus read-only and read/write access. The current behaviour causes Modbus errors if write-access to the PV inverter is not allowed.

This PR adds a new 'readOnly' setting that is activated by default. The idea of a 'read-only mode' is already existing in other places in OpenEMS for such use cases. In 'read-only mode' no PV limit is ever written. If there is a demand for pv limit anyway (e.g. via a `Controller.PvInverter.FixPowerLimit`), then a warning "Read-Only mode is active: PV-Limit failed" (Channel-ID: READ_ONLY_MODE_PV_LIMIT_FAILED) is shown.

### DIFF
--- a/io.openems.edge.pvinverter.sunspec/src/io/openems/edge/pvinverter/sunspec/AbstractSunSpecPvInverter.java
+++ b/io.openems.edge.pvinverter.sunspec/src/io/openems/edge/pvinverter/sunspec/AbstractSunSpecPvInverter.java
@@ -122,7 +122,7 @@ public abstract class AbstractSunSpecPvInverter extends AbstractOpenemsSunSpecCo
 		switch (event.getTopic()) {
 		case EdgeEventConstants.TOPIC_CYCLE_EXECUTE_WRITE:
 			// Get ActivePowerLimit that should be applied
-			IntegerWriteChannel activePowerLimitChannel = (IntegerWriteChannel) this
+			var activePowerLimitChannel = (IntegerWriteChannel) this
 					.channel(ManagedSymmetricPvInverter.ChannelId.ACTIVE_POWER_LIMIT);
 			var activePowerLimitOpt = activePowerLimitChannel.getNextWriteValueAndReset();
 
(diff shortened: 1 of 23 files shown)


io.openems.edge.pvinverter.sunspec/src/io/openems/edge/pvinverter/sunspec/AbstractSunSpecPvInverter.java
```diff
@@ -110,13 +115,28 @@ protected void deactivate() {
...
 		if (!this.isEnabled() || !this.isSunSpecInitializationCompleted()) {
 			this.channel(SunSpecPvInverter.ChannelId.PV_LIMIT_FAILED).setNextValue(false);
+			this.channel(SunSpecPvInverter.ChannelId.READ_ONLY_MODE_PV_LIMIT_FAILED).setNextValue(false);
 			return;
 		}
 
 		switch (event.getTopic()) {
 		case EdgeEventConstants.TOPIC_CYCLE_EXECUTE_WRITE:
+			// Get ActivePowerLimit that should be applied
+			IntegerWriteChannel activePowerLimitChannel = (IntegerWriteChannel) this
```
michaelgrill, line 125: either use type or cast one would be enough :)

michaelgrill (approved):
looks good 👍

io.openems.edge.pvinverter.sunspec/src/io/openems/edge/pvinverter/sunspec/AbstractSunSpecPvInverter.java
sfeilmeier: Indeed... 🤦‍♂️

PR #2148, opened 2023-04-19
Channel: catch errors while updating Process-Image

sfeilmeier:
Channel `nextProcessImage()` should not throw an Exception. Currently an IllegalArgumentException on updating of the process image is handled only very late, e.g. in the CycleWorker. This improvement catches the error early in the Channel itself and prints detailed error logs.


michaelgrill (approved):
👍

PR #2158, opened 2023-04-27
Rework of Timer

DerStoecki:
Previous PR #1754  of Timer reworked 
with @clehne 

This compared to the old PR #1754

is easier to use and easier to understand

TimerManager is kinda like a factory. 
OpenEmsComponent  that needs Timer/want to use Timer
uses @Reference of the TimerManager
And calls TimerManager methods to create a cycles/count/time Timer.

### DIFF
--- a/io.openems.edge.core/src/io/openems/edge/core/timer/DummyTimerByCoreCycles.java
+++ b/io.openems.edge.core/src/io/openems/edge/core/timer/DummyTimerByCoreCycles.java
@@ -1,0 +1,37 @@
+package io.openems.edge.core.timer;
+
+import io.openems.edge.common.channel.Channel;
+import io.openems.edge.common.timer.Timer;
+
+public class DummyTimerByCoreCycles extends AbstractTimer implements Timer {
+
+    private final int maxCoreCycles;
+    private final DummyTimerManager tm;
+    private int refCount;
+
+    public DummyTimerByCoreCycles(DummyTimerManager tm, //
+								  Channel<Integer> channel, //
+								  int maxCoreCycles, //
+								  int startDelayInSecs) {
+	super(tm, channel, startDelayInSecs);
+	this.tm = tm;
+	this.maxCoreCycles = maxCoreCycles;
+	this.refCount = tm.coreCyclesCount;
+    }
+
+    @Override
+    public boolean check() {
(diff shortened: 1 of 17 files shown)


io.openems.edge.core/src/io/openems/edge/core/timer/TimerManagerImpl.java
```diff
@@ -0,0 +1,85 @@
...
+
+    @Override
+    @Deactivate
+    protected void deactivate() {
+	super.deactivate();
+    }
+
+    @Override
+    public void handleEvent(Event event) {
+	this.coreCyclesCount++;
```
tsicking, line 63: `if (!this.isEnabled()) {
	    return;
	}`
und
`	switch (event.getTopic()) { ...
`
sind hier nicht nötig?

io.openems.edge.core/src/io/openems/edge/core/timer/DummyTimerByCoreCycles.java
clehne, line 1: Looks like Java Code Formatting needs to be applied

clehne (approved):
I directly applied the change requests from @tsicking within the documentation. 
Bundle looks generally good to me.  Still the DummyTimerByCoreCycles has a very few lines of unformatted code.

clehne:
@sfeilmeier Would you please take a final look on it. In my opinion it is ready for merging.

io.openems.edge.core/src/io/openems/edge/core/timer/TimerManager.java
```diff
@@ -0,0 +1,185 @@
...
+    /**
+     * Get a timer to count every core cycle.
+     * 
+     * @param channel if this channel is provided, this channel is used to hold the
+     *                state of the timer.
+     * @param count   the count to use.
+     * @return true if the given count has been reached, false else
+     */
+    public default Timer getTimerByCoreCycles(Channel<Integer> channel, int count) {
+	return this.getTimerByCoreCycles(channel, count);
```
tsicking, line 115: Shouldn't this be `getTimerByCoreCycles(channel, count, 0)`? Otherwise the method calls itself.

tsicking (changes requested):
I discovered a small error (I think).

PR #2159, opened 2023-04-27
Edge: Fixed grid buy and sell channel assignment

chrblabla:
While setting up an installation with a grid meter, I figured that the values for energy sold to grid and energy bought from grid are mixed up. This PR corrects this. 

Heads-up: this breaks the data which resides in the time-series database. This should not be a big deal as the data for the channels `_sum/GridSellActiveEnergy` and `_sum/GridBuyActiveEnergy` is mixed up anyways.

The following screenshot shows the pre-PR situation and the mix-up.

Bildschirmfoto 2023-04-27 um 13 23 48

### DIFF
--- a/io.openems.edge.core/src/io/openems/edge/core/sum/SumImpl.java
+++ b/io.openems.edge.core/src/io/openems/edge/core/sum/SumImpl.java
@@ -241,8 +241,8 @@ public class SumImpl extends AbstractOpenemsComponent implements Sum, OpenemsCom
 					gridActivePower.addValue(meter.getActivePowerChannel());
 					gridMinActivePower.addValue(meter.getMinActivePowerChannel());
 					gridMaxActivePower.addValue(meter.getMaxActivePowerChannel());
-					gridBuyActiveEnergy.addValue(meter.getActiveProductionEnergyChannel());
-					gridSellActiveEnergy.addValue(meter.getActiveConsumptionEnergyChannel());
+					gridBuyActiveEnergy.addValue(meter.getActiveConsumptionEnergyChannel());
+					gridSellActiveEnergy.addValue(meter.getActiveProductionEnergyChannel());
 
 					if (meter instanceof AsymmetricMeter) {
 						var m = (AsymmetricMeter) meter;


clehne (commented):
The grid energy handling is a bit special. It is kind of a definition. Without having a deeper look at your code. Are you sure that the problem is within the SumImpl. You may have used a grid meter component where the energy is shuffled within the meter implementation. 
Could you please doublecheck that the meter model (the grid meter) that you have used in your configuration behaves like other meter implementations?
I would think, that the meter has mixed production/consumption energy (energy must be shuffled, depending on the meter type). This error would be propagated to the SumImpl...

tsicking:
@chrblabla That also confused me once (or still confuses me?). Looking at the circle in the live UI, another way of seeing it would be: Production goes into the middle (i.e. enters the system) and consumption goes away from the middle (i.e. leaves the system). Without being 100% sure about it, I think this is how it is defined in OpenEMS, in which case "Sell to grid" would be consumption. Say, your production is comsumed by the grid :)

chrblabla:
Thought about this myself, but this would be highly confusing and doesn't explain the screenshot (where grid feed-in is happening at night). Maybe @sfeilmeier has some thoughts on this?

chrblabla:
After digging a bit deeper, we now understand that the main concept of energy flow is built around **energy entering the system (production)** and **energy leaving the system (consumption)**. This means, that energy bought from the grid is recognized as production energy which mislead us. The result is, that this PR in its current form is incorrect.

But an interesting learning from all of this (and how we understand it currently) is, that when reading out values from energy meters and assigning them to the OpenEMS channels, a distinction has to be made based on meter type, at least between **production** and **consumption** as different registers need to be assigned to different channels.

Energy meters get installed with the load "at the bottom":
- For production this means PV at the top, rest of the system at the bottom. 
- For consumption this means rest of the system at the top, consumer at the bottom. 
- For grid this means grid at the top, rest of system at the bottom.

Example:

General - Frame 11

Based on the example depicted above, we see that a certain register (30073 in case of the Microcare SDM630) needs to be mapped to different channels based on meter type.

What we currently see in the energy meter components is the use of an **invert functionality**. This is great if a meter has been installed "the other way around", but to our understanding a further distinction based on meter type needs to be done regardlessly.

@clehne @tsicking @sfeilmeier are these assumptions correct? If so, we would have to extend the meters with register to channel assignment based on meter type.

phfeustel:
I have created a pull request following @chrblabla proposal to demonstrate the logic in implementation.

Please have a look: https://github.com/OpenEMS/openems/pull/2176

chrblabla:
Closed in favor of the suggested implementation in #2176.

PR #2160, opened 2023-04-27
Fix/UI consumption energy display

phfeustel:
When having a consumption meter (type: CONSUMPTION_METERED) sending **ActiveConsumptionEnergy** values, they weren't shown in the consumption widget below the chart area for historic values. 

We identified the issue that the consumption area was filtering for **ActiveProductionEnergy** from a consumption meter instead of **ActiveConsumptionEnergy**.

This MR solves this issue. We also added a test with example data to validate the fix.

### DIFF
--- a/ui/src/app/edge/history/consumption/widget.component.ts
+++ b/ui/src/app/edge/history/consumption/widget.component.ts
@@ -14,6 +14,8 @@ export class ConsumptionComponent extends AbstractHistoryWidget implements OnIni
     @Input() public period: DefaultTypes.HistoryPeriod;
 
     private static readonly SELECTOR = "consumptionWidget";
+    public readonly CHANNEL_ID_ACTIVE_CONSUMPTION_ENERGY = "ActiveConsumptionEnergy";
+    public readonly CHANNEL_SUM_CONSUMPTION_ACTIVE_ENERGY = new ChannelAddress("_sum", "ConsumptionActiveEnergy");
 
     public data: Cumulated = null;
     public edge: Edge = null;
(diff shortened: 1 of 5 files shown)


ui/src/app/edge/history/consumption/widget.component.ts
```diff
@@ -78,30 +78,34 @@ export class ConsumptionComponent extends AbstractHistoryWidget implements OnIni
...
                 channels.push(
-                    new ChannelAddress(component.id, 'ActiveProductionEnergy'),
+                    new ChannelAddress(component.id, this.CHANNEL_ID_ACTIVE_CONSUMPTION_ENERGY),
                 )
             }
             resolve(channels);
         });
     }
 
+    public getChannelAddressString(component: EdgeConfig.Component, channelId: string): string {
```
lukasrgr, line 96: hey, i fell over your PR: template called functions and variables can be protected

ui/src/app/edge/history/consumption/widget.component.ts
```diff
@@ -28,8 +30,8 @@ export class ConsumptionComponent extends AbstractHistoryWidget implements OnIni
         super(service);
     }
 
-    ngOnInit() {
-        this.service.setCurrentComponent('', this.route).then(response => {
+    async ngOnInit() {
```
lukasrgr, line 33: a function call that you are not controlling shouldnt be asynchronous, did you find this somewhere or did you come up with that?

lukasrgr: is it guaranteed that the lifecycles will be called like expected, when they are async

sfeilmeier:
@lukasrgr FYI, there is a general discussion about this topic in the Community Forum: https://community.openems.io/t/change-energy-channel-assignment-based-on-meter-type/1603

PR #2062, opened 2023-02-13
Asymmetric virtual add meter

pooran-c:
This pull request consists of

1. Asymmetric Virtual add meter.
2. Changes in the Virtual symmetric add meter.
    - Changed the name from "Virtual add" to "SummetricVirtualAdd".
    - Min / Max active power calculation is deleted.
3. Common package.
4. Test cases for symmetric and asymetric virtual meter.

Usecase :
Customer wants to use Asymmetric peak shaving controller on virtual grid meter which is addition of two Asymmetric meters, and wants to read each phase vlaues in online monitoring.

PR #2166, opened 2023-05-03
Virtual meter bug fix: average is computed correctly fixed

JoopAue:
Hi team,

The average channel value for channels such as frequency was incorrectly calculated. Instead of taking the average of all channel values the average was computed as follows. (4 meter example) avg(avg(avg(avg(null, 1), 2), 3, 4) == 3.125 while 2.5 is expected.

The test case has also been corrected.

My only doubt is that I had to add overloads for sum and average to `TypeUtils` has varargs cannot be safely expressed as a Function or BiFunction, such as [here](https://github.com/OpenEMS/openems/compare/develop...JoopAue:openems:bug/virtual-meter-average?expand=1#diff-4f07af2e879a748e7a34c4328eb4361740d85af2d56e82a0f6602ba0ac24ea81R17)

Let me know if there are any questions.

Thanks!
Joop

(@pooran-c for transparency as I believe I'm adjusting [your code](https://github.com/OpenEMS/openems/pull/2062))

### DIFF
--- a/io.openems.edge.common/src/io/openems/edge/common/type/TypeUtils.java
+++ b/io.openems.edge.common/src/io/openems/edge/common/type/TypeUtils.java
@@ -2,6 +2,7 @@ package io.openems.edge.common.type;
 
 import java.lang.reflect.Array;
 import java.util.Arrays;
+import java.util.List;
 import java.util.Optional;
 
 import com.google.gson.JsonElement;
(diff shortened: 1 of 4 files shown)


JoopAue:
Hi team,

Curious if there is any followup on this one. Happy to provide more info or make any changes.
Can also close this one if it isn't a necessary change

sfeilmeier:
Hi @JoopAue and thanks for the code contribution!

Unfortunately due to the recent big code changes (mainly #2144) your patch now has file conflicts. Also I was not fully able to reproduce/understand the issue on first sight - sorry. 

Could you update the patch to latest 'develop' state? Possibly extending also the existing unit tests to be able to proof the case?

Thanks in advance!

JoopAue:
New version in #2278

PR #2174, opened 2023-05-11
feature/evcs api and compleo

tsicking:
This pull request consists of two parts:
1) An extension of the Evcs-Api with additional channels and a Utils-class.
2) The implementation of the Compleo Eco2.0 (and older model Wallbe Eco 2.0) based on that Api.

Note that it doesn't compile, as it is based on the Timer-Api (see PR#2158).


sebastianasen:
@tsicking 
Please do not be disappointed if this pull requst is not immediately reviewed.
As you say, it depends on another PR and has many general changes besides the new charging stations.

PR #2175, opened 2023-05-12
bugfix/meter-microcare

tsicking:
There were two small bugs in the meter:
1) Production and consumption energy were swapped (for both active and reactive energy)
2) The scale factor 3 for frequency was missing


phfeustel:
@tsicking I guess you have a similar idea but I think the energy channel assignment (and power inversion) should depend on the meter type.

Please have a look at our proposal in https://github.com/OpenEMS/openems/pull/2176 and discussion https://github.com/OpenEMS/openems/pull/2159#issuecomment-1535870927

Sn0w3y:
This is true! 
See the Attached Screenshot.
Screenshot_20230902_030620_Chrome

PR #2176, opened 2023-05-15
Switch energy assignment and power inversion based on the meter type.

phfeustel:
Following the discussion in https://github.com/OpenEMS/openems/pull/2159#issuecomment-1535870927, we created a fork of the Microcare SDM 630 as the Eastron SDM 630 meter implementation. 
We decided to create a fork and not change the Microcare meter directly to be able test the effects on the whole OpenEMS thoroughly.

Please have a close look at the logic in line [86 ](https://github.com/VoltStorage/openems/blob/401d447f93d325b1a759b5837c8c6e44bbd7d15d/io.openems.edge.meter.eastron.sdm630/src/io/openems/edge/meter/eastron/sdm630/MeterEastronSdm630Impl.java#L86)where use the `MeterType` to decide if a measured component is providing load into the system or is consuming from the system. Based on this we switch channel energy assignments as well as invert power and current values.


Byonnem:
Dear all,

I am still getting up to speed with openEMS and thus by far not an expert :) What I do not really seem to get in this discussion, is why not stick to the Kirchoff Current law (KCL) convention: The openEMS ecosystem is based around a single node network (to my understanding).

KCL: into node is negative, out of node is positive. This means that PV can be considered always negative, and (un)controllable load is by definition positive. (to my knowledge defined in IEC60375 "Conventions concerning electrical circuits")
Does this conflict on a code level that I do not (yet) understand?

afbeelding

Again, I am by no means an expert, just an eager (electrical) student so explanation is more than welcome :)

phfeustel:
Dear @Byonnem,
thanks for the input. 

When we are talking only about the measurement of energy (not power, current, etc) OpenEMS (according to what you find in the code of the CoreSum module) has a very analogous definition compared to KCL. Only the signs are reversed.
```
System balance =
      Producer Produce + Consumer Produce + ESS Discharge
      - Producer Consume - Consumer Consume - ESS Charge

sell to gird = system balance > 0 ? system balance : last_sell_to_grid_value
buy from grid = system balance < 0 ? system balance : last_buy_from_grid_value
```
If you take `sell to grid` or `buy from grid` into the system balance equation the total system balance will always be 0.

Why the signs are reversed compared to KCL, I cannot tell. Maybe it just feels more "natural" that when producer's production it is positive as well as consumer's consumption is negative and if producer's production > consumer's consumption this leads to a positive "sell to grid" in terms of you are earning money.

PR #2200, opened 2023-05-30
Drop ClockProvider for JUnit tests

sfeilmeier:
It's more straight forward to use a Constructor with Clock for JUnit tests


sfeilmeier:
Found only later, that this approach would break Simulator-App. So we will still have to rely on a @Reference ClockProvider for getting a clock... will open another PR, to clean up the implementations.

PR #2219, opened 2023-06-07
UI: add new eslint rule to remove unused commas

lukasrgr:
- Added eslint rule for unnecessary commas: can be auto fixed with `ng lint --fix`
- extended Workflow trigger, to be triggered not only on push, but also on "pull_request"


lukasrgr:
@sfeilmeier adjusted workflow trigger to be triggered from a fork pullrequest as well. [related Article](https://frontside.com/blog/2020-05-26-github-actions-pull_request/)

sfeilmeier:
> @sfeilmeier adjusted workflow trigger to be triggered from a fork pullrequest as well. [related Article](https://frontside.com/blog/2020-05-26-github-actions-pull_request/). What dou think about that?

Ok, very good! About the 'unnecessary commas' ESlint I am not so sure. I am often using 'unnecessary commas' by purpose, because then adding an additional line later will only cause a git diff in one line and not two. This is similar to what I do e.g. in Java `bnd.bnd` files ([Example](https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.core/bnd.bnd#L20))

lukasrgr:
> > @sfeilmeier adjusted workflow trigger to be triggered from a fork pullrequest as well. [related Article](https://frontside.com/blog/2020-05-26-github-actions-pull_request/). What dou think about that?
> 
> Ok, very good! About the 'unnecessary commas' ESlint I am not so sure. I am often using 'unnecessary commas' by purpose, because then adding an additional line later will only cause a git diff in one line and not two. This is similar to what I do e.g. in Java `bnd.bnd` files ([Example](https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.core/bnd.bnd#L20))

I fell over so many unnecessary comma. That git diff shows 2 lines instead of one, is acceptable and doesnt make anything worse imo.

PR #2237, opened 2023-06-26
Draft: Implementation of SumStateAlert

DerStoecki:
Hey everyone,

We heard you were thinking about/working on the Alerting System.

Since @clehne and I already worked on that, by implementing a SumStateAlert (Even though its not tested completely yet).
We'd like to give you some insight/inspiration on what we were doing.

Hope this comes in handy and I didn't miss anything during the merge process.

Best Regards

### DIFF
--- a/io.openems.backend.alerting/src/io/openems/backend/alerting/Alerting.java
+++ b/io.openems.backend.alerting/src/io/openems/backend/alerting/Alerting.java
@@ -4,6 +4,7 @@ import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 
+import io.openems.backend.alerting.handler.SumStateEdgeHandler;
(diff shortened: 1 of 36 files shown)


sfeilmeier:
@da-Kai Could you have a first look, please?

da-Kai:
Hey @DerStoecki,

I have reviewed your code and noticed that your implamentation is way differnet from mine.

The idea behind my original implementation was that adding a new alerting feature would not require major changes to the existing alerting system. Instead, I envisioned that most of the logic for the SumState alerting should reside in the "Message" and "Handler" classes created for it. However, you have outsourced many things to abstract classes, which, in my opinion, is not very elegant.

I would suggest that we adopt my implementation, which I am also already working on. 
I will extend my implementation a bit more and create a pull request myself, where you can see how I imagined the implementation and we can compare the approaches.

DerStoecki:
Closed bc of dev #2260

PR #2252, opened 2023-07-04
Controller ESS Cycle: refactor + improve

huseyinsaht:
Bug Fix: 
- standby time
- set Active Power

Improvement:
- state machine


io.openems.edge.controller.ess.cycle/src/io/openems/edge/controller/ess/cycle/statemachine/Context.java
```diff
@@ -1,74 +1,103 @@
...
+	/**
+	 * Gets the required AC power set-point for AC- or Hybrid-ESS.
+	 * 
+	 * @param ess           the {@link ManagedSymmetricEss}; checked for
+	 *                      {@link HybridEss}
+	 * @param hybridEssMode the {@link HybridEssMode}
+	 * @param power         the configured target power
+	 * @return the AC power set-point
+	 */
+	protected Integer getAcPower(ManagedSymmetricEss ess, HybridEssMode hybridEssMode, int power) {
```
michaelgrill, line 58: why did you choose to return Integer instead of int?

io.openems.edge.controller.ess.cycle/src/io/openems/edge/controller/ess/cycle/statemachine/Context.java
```diff
@@ -1,74 +1,103 @@
...
+		var afterNow = now.isAfter(this.parsedStartTime.minusSeconds(1));
+		var beforeNow = now.isBefore(this.parsedStartTime.plusSeconds(59));
+		if (afterNow && beforeNow) {
+			return true;
+		}
+		return false;
 	}
 
+	/**
+	 * Is ESS SoC defined ?.
```
michaelgrill, line 88: Is this javadoc finished?
I prefer that these descriptions not end with a '?.'. You could rephrase it like 'Checks if the ESS SoC is defined.'

Also do you need list explanation and remeber the TODO

io.openems.edge.controller.ess.cycle/src/io/openems/edge/controller/ess/cycle/statemachine/ContinueWithChargeHandler.java
```diff
@@ -1,43 +1,50 @@
...
+				+ " Current Cycle [ " + controller.getCompletedCycles() + "] " //
+				+ "out of " + config.totalCycleNumber() + "]");
 
 		return State.CONTINUE_WITH_CHARGE;
 	}
+
+	@Override
+	protected void onExit(Context context) throws OpenemsNamedException {
+		final var controller = context.getParent();
+		controller.setLastStateChangeTime(LocalDateTime.now(context.componentManager.getClock()));
```
michaelgrill, line 48: you could extract this to a method in Context or maybe add event callbacks to the StateMachine the you would have to only set it once

io.openems.edge.controller.ess.cycle/src/io/openems/edge/controller/ess/cycle/statemachine/StateMachine.java
```diff
@@ -50,24 +51,16 @@ public StateMachine(State initialState) {
...
-		case FINAL_SOC:
-			return new FinalSocHandler();
-		case FINISHED:
-			return new FinishedHandler();
-		case UNDEFINED:
-			return new UndefinedHandler();
-		}
-		throw new IllegalArgumentException("Unknown State [" + state + "]");
+		return switch (state) {
+		case UNDEFINED -> new UndefinedHandler();
```
michaelgrill, line 55: just a general question why isnt the factory of the handler included in the enum?

io.openems.edge.controller.ess.cycle/src/io/openems/edge/controller/ess/cycle/Config.java
```diff
@@ -17,29 +17,35 @@
...
 
-	@AttributeDefinition(name = " Standby Time in Minute", description = "Wait Between Charge and Discharge (Just Integer in time).")
-	int standbyTime() default 1;
+	@AttributeDefinition(name = " Standby Time in Minute", description = "Wait Between Charge and Discharge (Just Integer in time minute).")
+	int standbyTime() default 30;
 
-	@AttributeDefinition(name = " Start Time in Hour [YYYY-MM-DD HH:MM]", description = "Start Charge/Discharge Hour (2021-09-02 13:35).")
-	String startTime() default "2021-09-07 13:35";
+	@AttributeDefinition(name = "Start Time in Hour [YYYY-MM-DD HH:MM]", description = "Start Charge/Discharge Hour (2023-03-23 15:35).")
+	String startTime() default "2023-03-23 15:35";
```
michaelgrill, line 36: I would not change configurations like the date to a more recent date. I would pick some generic date like '2000-01-01 12:00'. If you want to provide a configuration example i would do it via documentation

michaelgrill (changes requested):
Minor things

io.openems.edge.controller.ess.cycle/src/io/openems/edge/controller/ess/cycle/Config.java
huseyinsaht: I wanted to write it directly here, becuase it was simple to configure it and no body checks the docs. Or we might just let this " Charge/Discharge Hour (2023-03-23 15:35) " in config as an explanation.

io.openems.edge.controller.ess.cycle/src/io/openems/edge/controller/ess/cycle/statemachine/Context.java
huseyinsaht: null check is done in PowerConstraint apply method, actually we can return it as "int" but apply method was taking a Integer parameter, thats why I wanted to keep it same.

io.openems.edge.controller.ess.cycle/src/io/openems/edge/controller/ess/cycle/statemachine/Context.java
huseyinsaht: not finished, this was a idea to think about it but this is somethng to apply on all controllers. this is not the right the place to implement it, we should do somethng on abstract class level. We can delete this TODO from here.

io.openems.edge.controller.ess.cycle/src/io/openems/edge/controller/ess/cycle/statemachine/ContinueWithChargeHandler.java
huseyinsaht: actually we have somethng to set the next state, we can place it directly there. you have right.

michaelgrill (approved):

sfeilmeier (changes requested):
Need to discuss usage of State-Machine

io.openems.edge.controller.ess.cycle/src/io/openems/edge/controller/ess/cycle/ControllerEssCycleImpl.java
```diff
@@ -66,15 +71,18 @@ public ControllerEssCycleImpl() {
...
-		super.activate(context, this.config.id(), this.config.alias(), this.config.enabled());
-		if (OpenemsComponent.updateReferenceFilter(this.cm, this.servicePid(), "ess", this.config.ess_id())) {
+	private void activate(ComponentContext context, Config config) {
+		super.activate(context, config.id(), config.alias(), config.enabled());
+		if (this.applyConfig(context, config)) {
 			return;
 		}
 	}
 
-	// TODO add @Modified to enable configuration changes during long running cycles
```
sfeilmeier, line 77: @Modified was not actually implemented? Why?

io.openems.edge.controller.ess.cycle/src/io/openems/edge/controller/ess/cycle/ControllerEssCycleImpl.java
```diff
@@ -35,27 +38,29 @@
...
 
+	private LocalDateTime parsedStartTime;
+	private LocalDateTime lastStateChangeTime;
+	private Config config;
+
 	@Reference
 	private ConfigurationAdmin cm;
 
+	@Reference
+	private Power power;
```
sfeilmeier, line 56: Why do you need this reference her? Below you are working with `ess.getPower()`

io.openems.edge.controller.ess.cycle/src/io/openems/edge/controller/ess/cycle/ControllerEssCycleImpl.java
```diff
@@ -84,31 +92,62 @@ protected void deactivate() {
...
 		}
-
-		if (previousState != state) {
-			// State has changed
-			this.lastStateChange = LocalDateTime.now(this.componentManager.getClock());
+		switch (this.config.mode()) {
+		case MANUAL_ON:
+			// get max charge/discharge power
+			var allowedDischargePower = this.ess.getPower().getMaxPower(this.ess, Phase.ALL, Pwr.ACTIVE);
+			var allowedChargePower = this.ess.getPower().getMinPower(this.ess, Phase.ALL, Pwr.ACTIVE);
```
sfeilmeier, line 103: If you anyway provide `ess` to `Context`, then there is no need to read `allowedCharge/Dischare` here m.E.

io.openems.edge.controller.ess.cycle/src/io/openems/edge/controller/ess/cycle/ControllerEssCycleImpl.java
```diff
@@ -84,31 +92,62 @@ protected void deactivate() {
...
+					this.config, //
+					this.componentManager.getClock(), //
+					this.ess, //
+					allowedChargePower, //
+					allowedDischargePower, //
+					this.parsedStartTime);
+
+			// store current state in StateMachine channel
+			var currentState = this.stateMachine.getCurrentState();
+			this._setStateMachine(currentState);
```
sfeilmeier, line 116: Usually we do this as first command in the `handleStateMachine` logic. (see e.g. https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.ess.generic/src/io/openems/edge/ess/generic/symmetric/EssGenericManagedSymmetricImpl.java#L100)

Downside of your approach is, that the State will never get touched if mode is switched to `MANUAL_OFF`. Imagine a UI Widget that displays the State of the Controller: it would still show "START_DISCHARGE", even though the Controller actually does nothing.

io.openems.edge.controller.ess.cycle/src/io/openems/edge/controller/ess/cycle/statemachine/Context.java
```diff
@@ -1,74 +1,126 @@
...
+	 * <ul>
+	 * <li>true
+	 * <li>- if {@link ChannelId#SOC} is defined &&
+	 * <li>- if {@link ChannelId#ALLOWED_CHARGE_POWER} is defined &&
+	 * <li>- if {@link ChannelId#ALLOWED_DISCHARGE_POWER} is defined &&
+	 * </ul>
+	 * 
+	 * @return true if channels are defined.
+	 */
+	public boolean areChannelsDefined() {
```
sfeilmeier, line 93: I'll backport our `ChannelUtils` approach here

sfeilmeier (changes requested):

PR #2256, opened 2023-07-07
ModbusRecordChannel: fix read error for write-only channels

tsicking:
An Edge2EdgeEss tried to read out the write-only channels, causing a read error. This is fixed by setting the modbus register of a write-only channel to "undefined" (like it is done for the registers in between natures).


tsicking:
May I ask for a review on this and the related PR #2257 ? We use the Edge2Edge-Ess in read-write-mode in two productive systems, and without those PRs it wouldn't work. Thanks.

sfeilmeier:
> May I ask for a review on this and the related PR #2257 ? We use the Edge2Edge-Ess in read-write-mode in two productive systems, and without those PRs it wouldn't work. Thanks.

Sury. In fact the problem had been solved in https://github.com/OpenEMS/openems/commit/f0b052746e91a9f21983cae3fd9fb4605990ded0#diff-bb02b130c9bfd04e1cdb847b3f138cd028aef03a1770b6bfc9a6bde6894f26ceR120-R124 before - but your solution is better. 

I merged the changes here in this PR, added a JUnit test and migrated the entire file to switch expressions.

sfeilmeier (approved):

PR #2257, opened 2023-07-07
One more channel added to Modbus slave table of Symmetric ESS

tsicking:
The channels MAX_APPARENT_POWER and CAPACITY were added to the Modbus slave table of SymmetricEss. 
Reason: MAX_APPARENT_POWER is needed for controlling an Edge2EdgeEss from remote (otherwise there may is an exception from the power component), and CAPACITY is needed for the controller "Grid optimized charge".


tsicking:
I've updated this PR. `CAPACITY` has already been added by now. With this PR, setting the maximum apparent power in `Edge2EdgeEssImpl.java` can be omitted.

clehne (approved):

clehne:
@sfeilmeier Same here (see my comment on #2265). This PR is also old and it simply may have got lost in the crowd of PRs last year? Without this the Edge2EdgeEss will not work.

sfeilmeier (approved):

PR #2259, opened 2023-07-10
allowed for language preferences to persist after reload

avstoyanov:
added a check for the locally stored language in the login method in websocket.ts and that made language changes persist after reload.
Train of thought was that since the method would be run to authenticate the user, it would run on every reload and therefore language was always being set to "authenticateResponse.user.language.toLocaleLowerCase()" instead of the specified language.

I'm also not sure why it's being set to that instead of calling navigator.language, but since I don't exactly know what it does I'm leaving it be.

on another fork I just use 
```
public static readonly getBrowserLang = () =>
        (navigator.language || (navigator.languages || ["en"])[0]).substring(0, 2);

```
in the language.ts file and call getBrowserLang to get the initial language at login

### DIFF
--- a/ui/src/app/shared/service/websocket.ts
+++ b/ui/src/app/shared/service/websocket.ts
@@ -158,7 +158,7 @@ export class Websocket implements WebsocketInterface {
       this.sendRequest(request).then(r => {
         let authenticateResponse = (r as AuthenticateResponse).result;
 
-        let language = Language.getByKey(authenticateResponse.user.language.toLocaleLowerCase());
+        let language = Language.getByKey(localStorage.LANGUAGE || authenticateResponse.user.language.toLocaleLowerCase());
         localStorage.LANGUAGE = language.key;
         this.service.setLang(language);
         this.status = 'online';


sfeilmeier:
@lukasrgr Could you take over, please?

lukasrgr (changes requested):
Hello together,

in my opinion, it does not make sense to call the authenticateRequest first, if you already have that information in localStorage, on the other hand this change is unnecessary and can lead to errors if you log in with a new user, because then its using the language saved in localStorage.

Do you need this workaround?

avstoyanov:
Ah you make a good point. Well the issue I was trying to solve is the language getting reset to German every time the page gets reloaded, which meant that there was no way to make sure that every part of the page got translated when the language was switched (probably because of some implementation error that seemed like a lot of work to fix), so I opted for a workaround that would just let me refresh and have everything translated instantly.

For the issue of logging in with a new user, there could be a check for which user is being logged in, or the implementation of local storage can be improved to include some identifier for the user that is logged in... Actually I'm not entirely sure if that will be an issue in the first place, since that would require 2 separate instances of the website to need to be loaded with different languages on the same machine.

I think associating the language preference with a user account makes more sense, though it is somewhat more complicated. Anyway, I think it would be less of a problem if the initial locale detection worked, but it doesn't (I am not in Germany and I don't speak German, so why does it default to German?)

All that being said, I could probably fix locale detection using that line I included in the original comment if you think that checking for currently stored language on login would cause too many issues. But that doesn't fix the problem of the page remaining mostly untranslated after switching languages. I do think that my proposed solution allows for less language switching and overall a smoother experience than the current solution, even if it did have proper locale detection

avstoyanov:
> in my opinion, it does not make sense to call the authenticateRequest first, if you already have that information in localStorage

I am also not sure what you mean by this. Nothing else is changed other than adding a check for locally stored language in the authenticateRequest that already is being called first. Otherwise localStorage gets overwritten by the authenticateRequest every time the tab is reloaded

lukasrgr:
> > in my opinion, it does not make sense to call the authenticateRequest first, if you already have that information in localStorage
> 
> I am also not sure what you mean by this. Nothing else is changed other than adding a check for locally stored language in the authenticateRequest that already is being called first. Otherwise localStorage gets overwritten by the authenticateRequest every time the tab is reloaded

The language is been set within Metadata, what Metadata are you using, if you use backend with fileMetadata you can find it [here] (https://github.com/OpenEMS/openems/blob/52f34d3c34663b9b81edc32230ea02097372ac6b/io.openems.backend.metadata.file/src/io/openems/backend/metadata/file/MetadataFile.java#L86C49-L86C49). If you do not use the backend then set it [here](https://github.com/OpenEMS/openems/blob/52f34d3c34663b9b81edc32230ea02097372ac6b/io.openems.common/src/io/openems/common/session/Language.java#L19)

PR #2269, opened 2023-07-14
UI bugfix: history energy chart

sfeilmeier:
- show ESS charge/discharge for systems without DC production
- show 'null' for 'null' values


lukasrgr (approved):
Thanks for helping

PR #2278, opened 2023-07-20
Virtual meter bug fix: average is computed correctly

JoopAue:
New pull request after applying changes to latest version of OpenEMS. Closing original PR #2166

The average channel value for channels such as frequency and voltage was incorrectly calculated. Instead of taking the average of all channel values the average was computed as follows. 

Example based on the adjusted test case for voltage:
**old:** avg(avg(avg(null, 10), 20), 30) == 22.5 ❌
**new** avg(10, 20, 30) == 20 ✅

The test case has also been adjusted surface the problem without the fix and pass correctly with the new code.

@sfeilmeier, hope this addresses your concerns in #2166

### DIFF
--- a/io.openems.edge.common/src/io/openems/edge/common/type/TypeUtils.java
+++ b/io.openems.edge.common/src/io/openems/edge/common/type/TypeUtils.java
@@ -2,6 +2,7 @@ package io.openems.edge.common.type;
 
 import java.lang.reflect.Array;
 import java.util.Arrays;
+import java.util.List;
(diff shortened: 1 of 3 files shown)


sfeilmeier (approved):
Thank you for your patience!

I cleaned the code a bit (uniform naming, Stream instead of for-loop, []::new instead of [0], Checkstyle).

JoopAue:
Thank you as well for taking the time!

PR #2286, opened 2023-07-27
ElectricityMeter: Use correct channels and avoid null pointer.

sebastianasen:
In calculateSumActiveProductionEnergyFromPhases() wrong channels were used.
Additionally it caused a null pointer, when the result would have been null, because it was internally trying to call intValue() to parse it into a long.


sfeilmeier (approved):
😯

PR #2288, opened 2023-07-31
Add CAN bridge using Linux SocketCAN for Kunbus Connect S, Connect+

clehne:
Add a CAN bridge with a JNI implementation for Linux Kernels V4 and V5. 
The implementation should work on every Linux system with enabled CAN Socket API. IT is tested and well used on  
* Kunbus Connect+ (Linux Kernel V4)
* Kunbus Connect S (Linux Kernel V5)

You can easily map CAN data from single CAN frames to OpenEMS channel. It is used almost like the Modbus Bridge. 
Unfortunately now, I can not provide any driver bundle which uses the CAN interface.


clehne:
Please see my additional comments here: https://community.openems.io/t/new-openems-can-bridge/1815

hydroid7:
Let's discuss it on the community!

hydroid7:
@clehne do you have any idea why the CI build does not pass?

hydroid7:
At the moment the binary is compiled for x86_64. I think we also should include the compiled binary for ARM.

clehne:
> @clehne do you have any idea why the CI build does not pass?

Hi @hydroid7, thanks for your hard work on this! I have rebased the PR on the latest development version. It turns out that the MetaTaskManager was still missing. This was a component needed for Modbus TCP Bridge and I am unsure if we should check it in again (as a common file). Therefore I made it part of the edge.bridge.can.common bundle. 
Unfortunately right now we do not support customers with a need for CAN. Will see if we can find a way to test it and to provide an ARM binary.

hydroid7:
I'll test this on our device, @clehne can you also do a short test to ensure that this works as expected?

hydroid7:
Tested on real device.

PR #2300, opened 2023-08-01
UI: Refactoring Peakshaving Symmetric

mahdiataie:


lukasrgr:
@mahdiataie can you fix the build first?

lukasrgr:
@AnasShetla i think you could continue with this after the prioritized tasks are done

PR #2304, opened 2023-08-05
Feature/update sunspec generator

nlamarti:
This pull request changes the generator for the sunspec model to use the json files as those are updated and maintained. I used the org.json as I did not want to change the generator completely so the classes have stayed the same just added a constructor that takes a json object where applicable.

P.S.: _Couldn't really find a template to use for a pull request so if this pull request needs any more info please let me know_


sfeilmeier:
Thank you for the PR! As we just recently discussed SunSpec in the community forum I mentioned this PR there. See https://community.openems.io/t/sma-pv-inverter-switching-to-sunspec-profile-2-0/1813/4

nlamarti:
I don't mind switching it to GSON just didn't want to do more rewriting than necessary, Am currently working on the implementation of the Socomec PMS ESS driver that is why I did this because I need some models that are not in the xml's

sfeilmeier:
Thanks! Ok. I just wanted to avoid that different people are working on the same part of the Code. Ideally you can synchronize your work with @clehne and @tsicking, who are currently working on a similar topic.

clehne:
Thanks for pointing us to this PR. We haven't done any work on the sunspec generator so far. Will take this into account, as soon as our sunspec inverter delivers the 700er code block.  Unfortunately this is still pending...

clehne:
Update: Our inverter delivers the new models now. 

@nlamarti So far your code looks good to me. If you don't mind, we will work directly on your PR, switch it to GSON library and add a few 700er models. Would that be ok for you? If so, could you please enable the "allow editing by maintainers" option on this PR. Thank you.

tsicking (changes requested):
Very nice! I have some small comments. The only really important one is the one in line 540 about the scale factors.

tsicking:
I just noted another thing: In the old version, the length of a model was the length without the first two points (ID and L), so the length of S1 was 66. This number was directly written into the xml-file.
In the json-files, the length is not written, instead you calculate it from the points, but you include the first two. Therefore, every model's length is increased by 2, so S1 now has length 68.

sfeilmeier:
> @nlamarti Please add your github user Email to #2324 (it ends with [..nlamarti@users.noreply.github.com](mailto:..nlamarti@users.noreply.github.com))? Than we can add you as "Co-authored-by".

@clehne FYI, you can generate the Co-authored-by line with this tool: https://coauthoredby.netlify.app/

PR #2315, opened 2023-08-14
Docs: Fix typing error in backend-to-backend section

mahdiataie:
Du stop the subscription --> To stop the subscription


sfeilmeier (approved):
Thanks!

PR #2316, opened 2023-08-14
UI: Fix notch color of pwa for ios devices

AnasShetla:
* adjust notch color in pwa for ios devices to match the background color.


lukasrgr:
This feature will be coming with a FEMS Backport

PR #2322, opened 2023-08-16
Bug Fix: FENECON Commercial 30 Off-Grid

huseyinsaht:


sfeilmeier:
@huseyinsaht the typo is quite obvious - but could you still provide a JUnit test that validates the fix? This way we will never do this same error again in the future.

sfeilmeier (approved):

PR #2323, opened 2023-08-21
UI: adjust energymonitor for mobile devices

lukasrgr:
### Changes
* Fix EnergyMonitor being too small in width on mobile devices
* Hide ion-card on mobile devices to create more space

### DIFF
--- a/ui/src/app/edge/history/common/energy/flat/flat.ts
+++ b/ui/src/app/edge/history/common/energy/flat/flat.ts
@@ -15,7 +15,7 @@ export class FlatComponent extends AbstractFlatWidget {
     protected autarchyValue: number | null;
     private static readonly EXCEL_TYPE = 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet;charset=UTF-8';
     private static readonly EXCEL_EXTENSION = '.xlsx';
-
+    protected readonly isMobile = this.service.isSmartphoneResolution;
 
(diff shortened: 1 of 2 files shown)


venu-sagar (approved):

ui/src/app/edge/history/common/energy/flat/flat.ts
```diff
@@ -15,7 +15,7 @@ export class FlatComponent extends AbstractFlatWidget {
     protected autarchyValue: number | null;
     private static readonly EXCEL_TYPE = 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet;charset=UTF-8';
     private static readonly EXCEL_EXTENSION = '.xlsx';
-
+    protected readonly isMobile = this.service.isSmartphoneResolution;
```
sfeilmeier, line 18: I always recommend using the same names (i.e. 'isMobile' or 'isSmartphoneResolution') in both places. I'd go for 'isSmartphoneResolution' because it is more expressive.

sfeilmeier (commented):
Are you finished with your reviews? Should I do the final merge?

ui/src/app/edge/history/common/energy/flat/flat.ts
lukasrgr: ok

sfeilmeier (approved):

PR #2324, opened 2023-08-21
SunSpecCodeGenerator: refactor, add 7xx models, use JSON input files

tsicking:
Based on PR 2304 by [nlamarti](https://github.com/nlamarti), this new PR handles the comments made there. Comparing the old and new output file `DefaultSunSpecModel.java`, the following changed:

- The descriptions "Ref 3: 8.14.3.2, Ref 4: 17" in the 12x models disappeared, because they are not in the json-files.
- The models 302-304 are now ignored. They only consist of one repeating block (which OpenEMS cannot handle as of now), and as far as I can see, no component uses them currently.
- The models 701-715 are added
- Descriptions of some points changed

Note that the length of model 701 is 155>126=maximum length of a modbus task in j2mod. As the `AbstractOpenemsSunSpecComponent` adds one task for each block, this causes an error (see issue 2311). I have fixed this already, this will be done in a seperate PR.

Furthermore, I used modern syntax in the switch-statements in `SunSpecPoint.java`, as I had to update them anyway due to the new pointn type STRING32. 

Lastly, I had to write a new method `getAsStringOrElse` in `JsonUtils.java`, and the other changes in this file are caused by the automatic code clean up.


clehne (approved):
Great replacement of the existing code. 

Like that you sorted the json files before generating the output code. This makes the generated code more easily comparable. 

Still a bit unsure about missing S303. Could imagine that some Inverters use it to transmit the temperature. But as no one is using it, it is probably okay to miss it.

clehne:
@sfeilmeier  As Sunspec is core functionality, could someone from your team have a look at it.

io.openems.common/src/io/openems/common/utils/JsonUtils.java
```diff
@@ -719,6 +718,23 @@ public static Optional<String> getAsOptionalString(JsonElement jElement, String
...
+	 *
+	 * @param jElement    the {@link JsonElement}
+	 * @param memberName  the name of the member
+	 * @param alternative the alternative value
+	 * @return the {@link String} value or the alternative value
+	 */
+	public static String getAsStringOrElse(JsonElement jElement, String memberName, String alternative) {
+		try {
+			return getAsString(jElement, memberName);
+		} catch (OpenemsNamedException e) {
```
sfeilmeier: Exceptions should not be used in normal control flow (e.g. https://dzone.com/articles/exceptions-as-controlflow-in-java)

I am replacing it with `return getAsOptionalString(jElement, memberName).orElse(alternative);`

Also JUnit test was missing.

sfeilmeier (approved):
Thank you for your contribution!

io.openems.common/src/io/openems/common/utils/JsonUtils.java
tsicking: Thanks a lot, I wasn't aware of that. I'll keep it in mind.

sfeilmeier: You're welcome. Usually this is not a big issue, but we found this to be a problem when we analyzed performance issues in the Backend Edge.Websocket using Visual.VM (https://visualvm.github.io/). During peak situations (e.g. restart of server) we had thousands of Exceptions per second which caused significant slow down.

PR #2325, opened 2023-08-21
Weblate: Translated strings

saeedniko:
Currently translated at 97.5% (513 of 526 strings)

Translation: FEMS/FEMS
Translate-URL: http://weblate.example.com/projects/fems/fems/de/


sfeilmeier:
Looks much better know. Unfortunately weblate still orders the entries differently (alphabetically), but we are likely not able to change that. Good work!

Issue #2311, opened 2023-08-09
Too long read tasks in Sunspec


tsicking:
### Bug Report or Feature Request (mark with an `x`)
```
- [x] bug report -> please search issues before submitting
- [ ] feature request
```

### Bug description or desired functionality.

The abstact Sunspec component adds one read task for each sunspec model. If the length of the model is > 126 (which is the case e.g for model S702), the task is too long and needs to be splitted (like it's done in Edge2Edge components).

PR #2336, opened 2023-08-31
Modbus SunSpec: split modbus tasks with more than 126 registers

tsicking:
This PR solves issue #2311. Similar to the Edge2Edge-component, if a modbus task is longer than the limit of j2mod (= 126), it is split. 

Note: When adding a JUnit test to see if it is working (I created a Dummy SunSpec Component which uses all default SunSpec models), I found an error in the `AbstractOpenemsSunSpecComponent`, namely that channels with a constant scale factor (e.g. S305_LAT) are not working. I will treat this in a seperate PR, which will also include my small JUnit test.

### DIFF
--- a/io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/sunspec/AbstractOpenemsSunSpecComponent.java
+++ b/io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/sunspec/AbstractOpenemsSunSpecComponent.java
@@ -1,6 +1,8 @@
 package io.openems.edge.bridge.modbus.sunspec;
 
+import java.util.ArrayDeque;
 import java.util.ArrayList;
+import java.util.Deque;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -21,6 +23,7 @@ import io.openems.edge.bridge.modbus.api.ElementToChannelConverter;
 import io.openems.edge.bridge.modbus.api.ElementToChannelScaleFactorConverter;
 import io.openems.edge.bridge.modbus.api.ModbusProtocol;
 import io.openems.edge.bridge.modbus.api.ModbusUtils;
+import io.openems.edge.bridge.modbus.api.element.AbstractModbusElement;
 import io.openems.edge.bridge.modbus.api.element.DummyRegisterElement;
 import io.openems.edge.bridge.modbus.api.element.ModbusElement;
 import io.openems.edge.bridge.modbus.api.element.ModbusRegisterElement;
@@ -36,6 +39,8 @@ import io.openems.edge.common.taskmanager.Priority;
  * This class provides a generic implementation of SunSpec ModBus protocols.
  */
 public abstract class AbstractOpenemsSunSpecComponent extends AbstractOpenemsModbusComponent {
+	
+	private static final int MAXIMUM_TASK_LENGTH = 126; 
 
 	private final Logger log = LoggerFactory.getLogger(AbstractOpenemsSunSpecComponent.class);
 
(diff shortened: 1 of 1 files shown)


io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/sunspec/AbstractOpenemsSunSpecComponent.java
```diff
@@ -37,6 +40,11 @@
  */
 public abstract class AbstractOpenemsSunSpecComponent extends AbstractOpenemsModbusComponent {
 
+	/**
+	 * Limit of a task length in j2mod.
+	 */
+	private static final int MAXIMUM_TASK_LENGTH = 126;
```
sfeilmeier, line 46: If this is really a hard limit, we should actually put this constant deeper into our Modbus Bridge. I am not able to find 126 as an explicit limit in j2mod (apart from https://github.com/steveohara/j2mod/blob/development/src/main/java/com/ghgande/j2mod/modbus/msg/WriteFileRecordRequest.java#L201, which appears to be exactly this hard-coded limit), but there is `256` as a maximum message length in https://github.com/steveohara/j2mod/blob/development/src/main/java/com/ghgande/j2mod/modbus/Modbus.java#L256C19-L256C19

I am not saying your approach is wrong, just:
- it would be perfect if we could use a constant from j2mod directly
- we should validate all Tasks e.g. in `defineModbusProtocol()` to throw an exception as early as possible

io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/sunspec/AbstractOpenemsSunSpecComponent.java
```diff
@@ -348,8 +356,37 @@ protected void addBlock(int startAddress, SunSpecModel model, Priority priority)
...
+	 * @param priority	the reading priority
+	 * @throws OpenemsException on error
+	 */
+	private void addReadTasks(Deque<ModbusElement> elements, Priority priority) throws OpenemsException {
+		var length = 0;
+		var taskElements = new ArrayDeque<ModbusElement>();
+		var element = elements.pollFirst();
+		while (element != null) {
+			if (length + element.length > MAXIMUM_TASK_LENGTH) {
+				this.modbusProtocol.addTask(//
```
sfeilmeier, line 375: I would have personally extracted this to another method, as it repeats 4 rather complicated lines of code right now. But it serves the job. 👍

To create an array from a list (or stream), you can also simply use `taskElements.toArray(AbstractModbusElement[]::new);`

sfeilmeier:
I have some remarks, but we can tackle them also in #2337. Thanks!

io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/sunspec/AbstractOpenemsSunSpecComponent.java
tsicking: Okay, thanks. We can refactor that later on.

io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/sunspec/AbstractOpenemsSunSpecComponent.java
tsicking: I was also thinking about this, as other modbus components might run into the same problem, and it's really hard to find the error if you don't know the limit (I was lucky I did, and I am grateful to whoever wrote it in the comment in AbstactEdge2Edge).

PR #2337, opened 2023-08-31
SunSpec: improvements

tsicking:
There are SunSpec Channels which have a constant scale factor, e.g. S305_LAT has a scale factor of -7. The current implementation of the `AbstractOpenemsSunSpecComponent` cannot handle these, but only those whose scale factor is read via modbus. This PR handles this issue.
Moreover, a DummySunSpecComponent with all default models is added for JUnit testing, as well as a small JUnit test checking that the modbus tasks are not too long.

Note that the changes of PR #2336 are also in here, as I didn't know how to stack this PR on top of the other one...


sfeilmeier:
Hi @tsicking,

I added some commits that fit into this file, because the JUnit test you provided made it very convenient to work with. Thanks a lot! 👍
- [DummySunSpecComponent: initialize SunSpec modals in ASC order](https://github.com/OpenEMS/openems/pull/2337/commits/0177fa9582e6bc4a56b5de43ed669fbd1756e6e9)
- [Refactor addReadTasks to preprocessModbusElements](https://github.com/OpenEMS/openems/pull/2337/commits/e1078838a9063cb00a50d4a304bbd061b87086d2) to be able to reuse it for WriteTasks
- [Create one FC16WriteRegistersTask for all writeable registers of one block](https://github.com/OpenEMS/openems/pull/2337/commits/6dc2b1bd1f8cd65e8596a7ef06513a53b1f35b67) This solves some perfomance problems we had with multiple WriteTasks being executed on every Cycle (e.g. write ActivePower and ReactivePower channels)
- [Extract generateElementToChannelConverter](https://github.com/OpenEMS/openems/pull/2337/commits/b294bfb212421a328bb9543592b5f72ec2593f0b) logic was split to two places before; also now I am always applying the 'isDefined' check - even for Points with Scale-Factor

I was not yet able to test this in hardware. I would appreciate if you could.

sfeilmeier (approved):

PR #2339, opened 2023-09-01
SMA PV-Inverter: use 70x models (since 2023)

tsicking:
SMA Sunny Tripower updated to make it compatible with newer firmware versions. See community post: https://community.openems.io/t/sma-pv-inverter-switching-to-sunspec-profile-2-0/1813
`AbstractSunSpecPvInverter` had to be updated too because SunSpec models S7xx are now supported.

Note that it won't work without PR #2336, as SunSpec Model S701 is longer than 126 registers.


sfeilmeier (approved):

sfeilmeier:
Well done! Thank you for your contribution - and sorry for the long delay.

PR #2345, opened 2023-09-05
Generic-Ess: fix debugLog for max charge power

sfeilmeier:
Due to a sign flaw the allowed max charge power was shown wrongly, i.e. if battery says "-56000" and battery-inverter "-92000", it would show "-92000". Fixed and proven with JUnit test

Also: add listener to execute `allowedChargeDischargeHandler` also if ESS StartStop changes.


huseyinsaht (changes requested):
review

huseyinsaht (commented):
review

huseyinsaht (approved):

PR #2347, opened 2023-09-06
Added ModbusSlave-funtionality to controller

DerWahreKlinki:
The status of the controller (regulation active or not) is readable via modbus


sfeilmeier:
@venu-sagar Could you please do a first review. The approach itself seems to be a good idea. Should it be reused in OpenEMS UI as well? There is `CUMULATED_ACTIVE_TIME`, so maybe `ACTIVE` would be sufficient instead of `REGULATION_ACTIVE`.

PR #2349, opened 2023-09-06
UI: Removing duplicate channels from SubscribeChannels and fix unsubscribing

mahdiataie:
### Changes
* remove duplicated channels in subscribeChannelsRequest
* liveDataService from Injectable to Directive: 
  * lifecycles dont work for Injectables
* added unsubscribeFromChannels, to be able to unsubscribe without passing the id used for the subscription


mahdiataie:
Main problem was with angular lifecycle and ngOnDestroy was not triggered. The problem was with routing and now problem with lifecycle is fixed. Some workaround needed for unsubscribing all channels correctly when we go from live to history, @lukasrgr

lukasrgr (approved):

PR #2353, opened 2023-09-08
Added Modbus feature; added posibility not to necessarily need an ESS

DerWahreKlinki:
Quick 'n' dirty approach to get the controller running without an ESS attached.

@sfeilmeier : code base of existing PR was too old. After 30 minutes of struggeling with ess and github I decided to delete and created a new branch & PR


sebastianasen (approved):
I am not sure if thher could be another solution without using the static string "NONE", but for now it would fix your problem and allow to use the Component without an ess reference.

DerWahreKlinki:
I´ve made the attribute optional in config. If the attribute left empty than no ESS is used by the controller.

Not tested with real ESS. But it´s working with a simulated one.

PR #2357, opened 2023-09-12
Backend: fix subscribe system log

michaelgrill:
System log was only send to one UI when mulitple UIs of the same user were opened.

added unique id for each websockte connection
replaced subscribe with token to ui websocket id


sfeilmeier (approved):
Thanks for the fix!

PR #2358, opened 2023-09-14
Docker workflow

SMHRambo:
Here is my version of openems dockerfile and github workflow.


sfeilmeier:
@da-Kai: You are the specialist on this field. Could you review this please?

For reference: https://community.openems.io/t/docker-und-ui-varianten/1863/9

tools/docker/docker-compose.yml
da-Kai, line 1: I don't see the point of this docker compose, since launching backend and edge on the same hardware, except for testing purposes, doesn't make much sense to me.

One Docker compose for a backend+UI+Influx and one for Edge+UI would probably be the most useful.

da-Kai (commented):
**Thank you for your contribution.** You seem to be more knowledgeable than me about Docker and especially about linuxserver.io images.

The workflow file actually looks good. But
As in the community post shared by Stefan: *https://community.openems.io/t/docker-und-ui-varianten/1863/8* it is necessary to build 2 different images for Backend-UI and Edge-UI, which is not the case in this config.

Also copying the *ui/src/themes/openems/environments/* folder is not really optimal and potentially error prone

I tried to test the images in my WSL. I was able to build all the images. And the containers for edge and backend worked fine, but not the UI, this one just gave me nginx config errors instead.

da-Kai:
I will close this Pull-Request, as Docker Workflows were implemented in #2568

PR #2361, opened 2023-09-18
Update gettingstarted.adoc

NiallDarwin:
Added instruction to "Disable the Api Controller: Controller Api Websocket." in 5.2, configure openEMS edge for backend

### DIFF
--- a/doc/modules/ROOT/pages/gettingstarted.adoc
+++ b/doc/modules/ROOT/pages/gettingstarted.adoc
@@ -319,6 +319,8 @@ Next we will configure OpenEMS Edge to connect to the OpenEMS Backend _**Edge.We
 
 . Switch back to the http://localhost:8080/system/console/configMgr[Apache Felix Web Console Configuration for OpenEMS Edge icon:external-link[]].
 
+. Disable the _**Api Controller: Controller Api Websocket**_ because Edge should not communicate directly with UI anymore.
+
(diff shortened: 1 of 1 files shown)


doc/modules/ROOT/pages/gettingstarted.adoc
```diff
@@ -319,6 +319,8 @@ Next we will configure OpenEMS Edge to connect to the OpenEMS Backend _**Edge.We
 
 . Switch back to the http://localhost:8080/system/console/configMgr[Apache Felix Web Console Configuration for OpenEMS Edge icon:external-link[]].
 
+. Disable the _**Api Controller: Controller Api Websocket**_ because Edge should not communicate directly with UI anymore.
```
sfeilmeier, line 322: Thanks for the contribution - but why do you think it is necessary to disable the Controller? 

Edge is never "communicating directly with UI", but the local OpenEMS UI is actively opening the connection to OpenEMS Edge. There is no harm in having the Controller still active.

NiallDarwin: Honestly I do not understand precisely the implications of having this open. When learning something new I think it is better to have less happening because fault finding is easier. [KISS](https://en.wikipedia.org/wiki/KISS_principle). If this controller remains active there is the possibility of hiding an error in backend/UI config because the UI can receive data from it.

PR #2371, opened 2023-09-28
UI: add check for pv inverter nature

lukasrgr:


lukasrgr:
@sebastianasen i removed all of the components that are implementing the nature MANAGED_SYMMETRIC_PV_INVERTER, is this correct?

sebastianasen:
@lukasrgr In principle, yes. Without having a closer look - how is this handled for charger (EssDcCharger). 
Currently there is only a separate check for Kostal.Piko.Charger in isProducer.

lukasrgr:
> @lukasrgr In principle, yes. Without having a closer look - how is this handled for charger (EssDcCharger). Currently there is only a separate check for Kostal.Piko.Charger in isProducer.

So if i remember right, the type is set for all essDcCharger, but couldnt find this in code, anyway im not touching this

ui/src/app/shared/edge/edgeconfig.ts
```diff
@@ -286,6 +286,17 @@ export class EdgeConfig {
         return false;
     }
 
+    /**
+     * Is the given meter implementing given nature
```
sfeilmeier, line 290: It could be any component, not only a meter

ui/src/app/shared/type/widget.ts
```diff
@@ -15,6 +15,7 @@ export enum WidgetNature {
     'io.openems.edge.evcs.api.Evcs',
     'io.openems.impl.controller.channelthreshold.ChannelThresholdController', // TODO deprecated
     'io.openems.edge.io.api.DigitalInput',
+    MANAGED_SYMMETRIC_PVINVERTER = 'io.openems.edge.pvinverter.api.ManagedSymmetricPvInverter'
```
sfeilmeier, line 18: Are you sure this is a good idea, to add it here? `WidgetNature` is used in `widget.ts` to evaluate the UI widgets. It is not required to add PV-Inverters there.

sfeilmeier (changes requested):

tsicking:
Hi, is there any update on this? I think it would be good to have this feature, because currently we have to add every new PV-Inverter to the hard coded list or add a config property for the meter type, which is not very handy.

PR #2372, opened 2023-09-29
UI: Fixing Infinity sign shown in charts & the wrong date formating of IBN

AnasShetla:
The issue of infinity sign was because if you add, subtract, multiply, divide, or exponentiate null , JavaScript converts null to 0.
so we used the Utiles.substractSafely to handle it.


lukasrgr (approved):
Thanks for optimizing code

ui/src/app/edge/history/common/energy/chart/chart.ts
```diff
@@ -101,7 +101,8 @@ export class ChartComponent extends AbstractHistoryChart {
             },
             converter: () =>
               data['ProductionActivePower']?.map((value, index) =>
-                value - data['GridSell'][index] - data['EssCharge'][index])?.map(value => HistoryUtils.ValueConverter.NEGATIVE_AS_ZERO(value)),
+                Utils.subtractSafely(Utils.subtractSafely(value, data['GridSell'][index]), data['EssCharge'][index]))
```
sfeilmeier, line 104: I suggest changing the Utils method to accept a varargs parameter:

```
  public static subtractSafely(...values: number[]): number {
    var result = null;
    values.forEach(v => {
      if (v != null) {
        if (result == null) {
          result = v;
        } else {
          result -= v;
        }
      }
    })
    return result;
  }
```

then using it as 
```
            converter: () =>
              data['ProductionActivePower']?.map((value, index) => Utils.subtractSafely(value, data['GridSell'][index], data['EssCharge'][index]))
                ?.map(value => HistoryUtils.ValueConverter.NEGATIVE_AS_ZERO(value)),
```

Also there is a subtraction just above this line:
```
              return energyValues.result.data['_sum/ProductionActiveEnergy'] - energyValues.result.data['_sum/GridSellActiveEnergy'] - energyValues.result.data['_sum/EssDcChargeEnergy'];
```
Shouldn't you use Utils.subtractSafely there as well?

ui/src/app/edge/settings/profile/profile.component.html
```diff
@@ -34,7 +34,7 @@
               </ion-item>
               <ion-item lines="none" *ngIf="edge.firstSetupProtocol">
                 <ion-label translate>Index.FIRST_SETUP_PROTOCOL</ion-label>
-                {{ edge.firstSetupProtocol | date:'dd.MM.YYYY hh:mm' }}
+                {{ edge.firstSetupProtocol | date:'dd.MM.YYYY HH:mm' }}
```
sfeilmeier, line 37: Good catch! 👍

https://angular.io/api/common/DatePipe

sfeilmeier (changes requested):

ui/src/app/edge/history/common/energy/chart/chart.ts
AnasShetla: Thanks for the insightful review, 
yes i should have used Utils.substractSafely there also.
i am working on the requested changes.

sfeilmeier (changes requested):
Please @lukasrgr review first

ui/src/app/shared/service/utils.ts
```diff
@@ -110,14 +110,18 @@ export class Utils {
    * @param v1 
    * @param v2 
    */
-  public static subtractSafely(v1: number, v2: number): number {
```
lukasrgr, line 113: update jsDoc

ui/src/app/shared/service/utils.ts
```diff
@@ -110,14 +110,18 @@ export class Utils {
...
-    if (v1 == null) {
-      return v2;
-    } else if (v2 == null) {
-      return v1;
-    } else {
-      return v1 - v2;
-    }
+  public static subtractSafely(...values: number[]): number {
+    var result = null;
+    values.forEach(v => {
```
lukasrgr, line 115: i think you could also filter out all null values before, would be better readable

lukasrgr (changes requested):
You

lukasrgr:
@AnasShetla if you didnt get notified, can you finish this PR?

lukasrgr (approved):

sfeilmeier (approved):

PR #2385, opened 2023-10-10
Fix odoo login

slash-ds:


miettal:
odoo-openems 16.0 support passing has_multiple_edges attribute. so updating is one way.

// it strange that why passing has_multiple_edges from odoo-openems to backend is needed.
// because `jDevices.size() > 1` cheking(like your PR) is enough to determine has_multiple_edges.

sfeilmeier:
> // it strange that why passing has_multiple_edges from odoo-openems to backend is needed.
> // because `jDevices.size() > 1` cheking(like your PR) is enough to determine has_multiple_edges.

@miettal: The reason is performance. If you read the Odoo OpenEMS Addon carefully, you will see, that it stops early for global read-only/admin roles, avoiding possibly thousands of entries - and instead delivering an empty `devices` array: https://github.com/OpenEMS/odoo-openems/blob/16.0/openems/controllers/openems_backend.py#L33-L44

The `devices` array is actually a relic of earlier times, when we did not yet have pagination in UI.

I'll close this. Feel free to ask back if this does not solve your question.

PR #2388, opened 2023-10-11
Feature/pylontech battery implementation

thomasjbhayes:
Implementation of Pylontech Powercube M2 battery as OpenEMS component. 

Design of battery implementation is based on Fenecon Commercial Battery implementation.

The implementation still has some issues, in particular around the StateMachine.

In unit tests, the StateMachine progresses correctly through the relevant states when starting the battery. In practice, it remains UNDEFINED, even if the other relevant states (startStopTarget and SystemStatus) are the same as in the unit tests.


tsicking:
Hi Thomas,

I started with an implementation of the battery about two months ago and didn't realize there is already this PR. I will review it in the next days.

Best regards,
Thomas

thomasjbhayes:
> Hi Thomas,
> 
> I started with an implementation of the battery about two months ago and didn't realize there is already this PR. I will review it in the next days.
> 
> Best regards, Thomas

Hi Thomas,

Good to hear that you are also interested in this battery! Unfortunately I realised that I have not been disciplined enough with my development and have added extra commits to this branch/PR since finishing the battery development. I must remove my last commit from this PR as it is nothing to do with the Pylontech battery.

WIll do this an update the PR

thomasjbhayes:
@tsicking I have now updated the PR to remove the last commit. I have had the OpenEMS device working with the battery in practice for a couple of months now and it is working ok for us.

thomasjbhayes:
Hi Thomas,

Thank you for checking out and inspecting the code.

I will tidy up the code and push another commit in a while fixing the checkstyle issues. I will also remove the additional files as well. Strangely I had thought I removed the last commit on here but it is there again now. In any case I will tidy it up.

With regards to the low temperature alarms, we have also experienced these and I would like to know more about your experience with this. We also had this issue over the past few weeks where we were unable to charge/discharge the battery when the cell temperature was below 10 degrees.

I spoke with Pylontech's engineer who said that the behaviour of the battery is as follows:

- **< 10°C - Low temperature alarm** - The battery can still charge and discharge, but current is reduced
-  **< 0°C - Low temperature protection** - The battery relay/contactor will open and the battery cannot charge/discharge

In my code I have encoded the temperature alarm as  `io.openems.common.channel.Level.FAULT`, which means that the StateMachine prevents the battery and ESS from running when this is raised. This is probably not the correct approach, instead I think it would make sense to allow charge/discharge but _ensuring that the current limits are respected_ by the EMS & Inverter. 

What do you think? How have you handled this in your code?

thomasjbhayes:
More generally, I have encoded all the Pylontech "Alarm" channels as `Level.FAULT`. Given Pylontech's explanation of the difference between "Alarm" and "Protection", these should probably be `Level.WARNING`.

tsicking:
Yes, I also had them on `Level.FAULT` first and changed it to `Level.WARNING` so that the state machine doesn't go into error and you can still use the battery. I see that the charge max current goes down to 30A, and even to 2A if it is very cold. Then the Generic ESS will automatically take care of it by adjusting its allowed charge power.

clehne:
@thomasjbhayes  @tsicking 
Just a short note: I am a bit unlucky about Pylontech naming the Temperature Flag as "TEMPERATURE_ALARM"
Because I would intuitivly map an ALARM to an OpenEMS Fault-State. Other BMSes provide warnings, errors and alarms. 
Pylontech only offers ALARMS. 
So to be more consistent on the OpenEMS side I would prefer to map the Temperature Alarm modbus registers to temperatureWarning OpenEMS Channels. and make them of type `Level.WARNING`. This confuses the developer looking at the pylontech driver bundle. But it is more intuitive for the guys operating different battery vendors. 
Also I would probably add an UnderTemperature State.Fault channels when the temperature is below 0,1 or 2 °C to signal an error, when or before the battery goes offline.

thomasjbhayes:
@clehne @tsicking I agree that it is unfortunate that Pylontech named these registers as 'Alarm' when realistically they are warnings. I have changed all the alarm channels to Level.WARNING already but I think I should change the name too (to warning).

Pylontech use the term 'protection' for the statuses when the contactor/relay opens and the system shuts down due to an alarm condition (e.g temperature < 0). To me I feel intuitively that 'Alarm' is the best name for this status, but in this case I think that would invite too much confusion. Do you have a suggestion for these cases?

I have deleted the unneeded files for the other bundles from this PR in a new commit I added. Previously I tried to revert the old commit which just confused matters.

I have also ran the checkstyle plugin and the Eclipse code formatter and have dealt with almost all the issues. The  outstanding issue is related to comments on very long lines of code. The Eclipse plugin is suggesting I do it one way (where the comment is split across multiple lines but all lines are aligned vertically), but the Checkstyle suggests something different.
To me the code as it is now makes more sense from a style perspective. Is it ok if I keep the comment style as suggested by the Code Formatter or do I need to strictly follow the checkstyle? 

It's my first time contributing to the public repo so I am trying to learn the proper convention.

thomasjbhayes:
I have pushed a couple of commits to rename the 'alarm' channels to 'warning' so that they will be more easily understood. 
I have left the 'Protection' channels as-is. I am open to renaming these if someone suggests a better name. I don't want to use 'Alarm' for these as I think it would be confusing given that Pylontech calls the 'Warnings' alarms.

I have fixed the checkstyle issues as well. Except for the ones I mentioned in my last comment where the checkstyle plugin is wanting me to change the indentation on the comments in a way that is contradictory to the Code Formatter.
I will take your lead on the preferred styling for these comments.

tsicking:
Hi, we are reading out the values from the device not using the `BatteryProtection` and it's the same behaviour: The charge max current goes down to 30A and then to 2A, but the discharge max current remains at 148A all the time.

thomasjbhayes:
Ok - I will validate with Pylontech about this behaviour just to be sure that the batteries are behaving correctly. Seems strange to me that there would only be limitations on charging and not discharging.

Have you experienced the behaviour of the battery at lower temperatures (below 0 degrees)? I believe the charge and discharge currents should go to zero, but I do not have the ability to cool the battery in our test setup to these temperatures so have not experienced it.

thomasjbhayes:
Hi @tsicking 
I have reviewed this with Pylontech and this is expected behaviour. The discharge current does not ramp down, due to the physical characteristics of the battery, the discharge current is relatively unaffected by temperature, until it reaches the threshold for the 'Protection' state at which it directly goes to zero.

tsicking:
I'm sorry; I accidentally pushed two commits... I can't find how to revert the PR to your last commit c4f4e5d527cd2ba5b7a3c193965b9d4d683b0569 (Fix checkstyle issues). Can you do it? Otherwise I'll find out tomorrow how I can do it.
Best regards and sorry for the inconvenience...

thomasjbhayes:
Hi @tsicking ,

**Disregard my previous comment**

I have now reverted the two most recent commits and force pushed to the branch. 

**Please try pulling the PR before making any changes**

io.openems.edge.battery.pylontech/test/io/openems/edge/battery/pylontech/powercubem2/PylontechPowercubeM2BatteryImplTest.java
```diff
@@ -0,0 +1,101 @@
...
+				.addReference("setModbus", new DummyModbusBridge(MODBUS_ID)) //
+				.activate(MyConfig.create() //
+						.setId(BATTERY_ID) //
+						.setModbusId(MODBUS_ID) //
+						.setModbusUnitId(1) //
+						.setStartStop(StartStopConfig.STOP) //
+						.build())
+				.next(new TestCase("Battery running").input(STATUS_CHANNEL, PylontechPowercubeM2Battery.Status.CHARGE)
+						.input(STATE_MACHINE, StateMachine.State.RUNNING))
+				.next(new TestCase("Stopping").output(STATE_MACHINE, StateMachine.State.GO_STOPPED)
```
tsicking, line 75: Adding a line break before each `.input()` and `.output()` would make it a bit easier to read I find.

io.openems.edge.battery.pylontech/test/io/openems/edge/battery/pylontech/powercubem2/PylontechPowercubeM2BatteryImplTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+	private static final ChannelAddress STATE_MACHINE = new ChannelAddress(BATTERY_ID,
+			PylontechPowercubeM2Battery.ChannelId.STATE_MACHINE.id());
+	private static final ChannelAddress STATUS_CHANNEL = new ChannelAddress(BATTERY_ID,
+			PylontechPowercubeM2Battery.ChannelId.SYSTEM_STATUS.id());
+
+	@Test
+	public void wakeUpBattery() throws Exception {
+
+		final var clock = new TimeLeapClock(Instant.parse("2020-01-01T01:00:00.00Z"), ZoneOffset.UTC);
```
tsicking, line 34: Is the clock necessary? You can construct a `DummyComponentManager` without a clock too. The same for the other test and the `startBatteryTest` in the other file.

io.openems.edge.battery.pylontech/test/io/openems/edge/battery/pylontech/powercubem2/PylontechPowercubeM2GenericEssIntegrationTest.java
```diff
@@ -0,0 +1,85 @@
...
+	private static final ChannelAddress STATUS_CHANNEL = new ChannelAddress(BATTERY_ID,
+			PylontechPowercubeM2Battery.ChannelId.SYSTEM_STATUS.id());
+	private static final ChannelAddress START_STOP = new ChannelAddress(BATTERY_ID,
+			StartStoppable.ChannelId.START_STOP.id());
+	
+	
+	@Test
+	public void startBatteryTest() throws Exception {
+		
+		final var clock = new TimeLeapClock(Instant.parse("2020-01-01T01:00:00.00Z"), ZoneOffset.UTC);
```
tsicking, line 37: See my comment in the other test file.

io.openems.edge.battery.pylontech/test/io/openems/edge/battery/pylontech/powercubem2/PylontechPowercubeM2GenericEssIntegrationTest.java
```diff
@@ -0,0 +1,85 @@
...
+					.build())
+			.next(new TestCase("Start in UNDEFINED state. The StartStop is UNDEFINED to begin with")
+					.input(STATUS_CHANNEL, Status.IDLE)
+					.output(STATE_MACHINE, StateMachine.State.UNDEFINED)
+					.output(START_STOP, StartStop.UNDEFINED))
+			.next(new TestCase("Check that nothing is changing without external input")
+					.output(STATUS_CHANNEL, Status.IDLE)
+					.output(STATE_MACHINE, StateMachine.State.UNDEFINED)
+					.output(START_STOP, StartStop.UNDEFINED))
+			.next(new TestCase("Check that nothing is changing without external input")
```
tsicking, line 59: This test case is added twice.

tsicking (changes requested):
I have had a first more detailed look at the PR now. Have a look at my comments.
I know that the checkstyle rules contradict the code formatter when it comes to indentation of comments. Still the checkstyle violations have to be fixed, because some automatic checks will fail otherwise. One can try for workarounds like starting the comment in the line after or before the actual code, or maybe some of the comments can be removed completely.
I haven't looked at the implementation in every detail, but more at the formal things first. 
Good that you added a JUnit test which is a bit more elaborate than the standard test.

thomasjbhayes:
Hi @tsicking ,

Thanks for this initial review. I will work through these issues now and fix the comments too.

Thomas

io.openems.edge.battery.pylontech/src/io/openems/edge/battery/pylontech/powercubem2/PylontechPowercubeM2BatteryImpl.java
```diff
@@ -0,0 +1,1150 @@
...
+		);
+	}
+
+	private final Logger log = LoggerFactory.getLogger(PylontechPowercubeM2BatteryImpl.class);
+
+	@Reference
+	protected ConfigurationAdmin cm;
+
+	@Reference
+	protected ComponentManager componentManager; // Is this ever initialised?
```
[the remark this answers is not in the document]
thomasjbhayes: Indeed :) It was a learning experience for me in the OSGi framework as I wrote this

io.openems.edge.battery.pylontech/src/io/openems/edge/battery/pylontech/powercubem2/statemachine/Context.java
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	/**
+	 * Checks if battery is awake - i.e in charge / discharge modes.
+	 * 
+	 * @return boolean which says if battery is awake.
+	 */
+	protected boolean isBatteryAwake() {
+		Status status = this.getParent().getSystemStatus();
+		if (status == Status.CHARGE | status == Status.DISCHARGE | status == Status.IDLE) {
```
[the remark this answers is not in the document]
thomasjbhayes: Yes you are right - good spot

io.openems.edge.battery.pylontech/test/io/openems/edge/battery/pylontech/powercubem2/PylontechPowercubeM2BatteryImplTest.java
thomasjbhayes: Yes - this was unnecessary

io.openems.edge.battery.pylontech/src/io/openems/edge/battery/pylontech/powercubem2/statemachine/StoppedHandler.java
```diff
@@ -0,0 +1,20 @@
...
+import io.openems.common.exceptions.OpenemsError.OpenemsNamedException;
+import io.openems.edge.common.statemachine.StateHandler;
+import io.openems.edge.battery.pylontech.powercubem2.statemachine.StateMachine.State;
+import io.openems.edge.common.startstop.StartStop;
+
+public class StoppedHandler extends StateHandler<State, Context> {
+
+	@Override
+	public State runAndGetNextState(Context context) throws OpenemsNamedException {
+
```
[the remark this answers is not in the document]
thomasjbhayes: Yes, good suggestion.

io.openems.edge.battery.pylontech/src/io/openems/edge/battery/pylontech/powercubem2/PylontechPowercubeM2BatteryProtectionDefinition.java
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	@Override
+	public PolyLine getDischargeVoltageToPercent() {
+		return PolyLine.empty();
+	}
+
+	@Override
+	public PolyLine getChargeTemperatureToPercent() {
+		return PolyLine.empty();
```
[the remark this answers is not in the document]
thomasjbhayes: Great :)

io.openems.edge.battery.pylontech/src/io/openems/edge/battery/pylontech/powercubem2/PylontechPowercubeM2BatteryProtectionDefinition.java
```diff
@@ -0,0 +1,59 @@
...
+
+public class PylontechPowercubeM2BatteryProtectionDefinition implements BatteryProtectionDefinition {
+
+	/*
+	 * Most values not defined. Those that are defined come from Pylontech engineer
+	 * 
+	 */
+	@Override
+	public int getInitialBmsMaxEverChargeCurrent() {
+		return 148; // [A]
```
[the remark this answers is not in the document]
thomasjbhayes: Yes. I am not sure how to handle this one. I don't believe the number of piles is defined until the modbus registers are read - when the BatteryProtectionDefinition is instantiated we don't know the number. Perhaps it is better not to define this as the BMS should protect the battery by providing the actual charge/discharge limits?

thomasjbhayes:
Hi @tsicking
Thank you for taking the time to review this PR. 
It was a very useful exercise as I realised there were several areas where I had not finished cleaning up code after my initial implementation. Especially with regards to the comments - most of them were obsolete and could be deleted (helping with the checkstyle too).

It is not passing the checkstyle tests and I have amended several aspects in accordance with your comments.

The BatteryProtectionDefinition is still one area where I am unsure - per my comment, we will not know the number of piles on initialisation and therefore I am not sure if we can put a current limit in the Definition.

More broadly, I have not tested the code that handles multiple piles in practice yet. We have only got one pile for testing, however we plan on testing with multiple over the coming months. I have received an MBMS from Pylontech and will soon test with this and a single pile (right now it is direct to pile).

thomasjbhayes:
My apologies, I realised I had left an occurrence of TimeLeapClock in one of the tests and this was causing the build to fail.

thomasjbhayes:
Hello,

My attention was away from this matter over the past couple of months. But I think the only outstanding thing was that I forgot to resolve one of @tsicking 's requested changes which I had previously actually worked on in the code.

I think I have marked all of the changes as done now - is it possible to merge the PR now?

io.openems.edge.battery.pylontech/src/io/openems/edge/battery/pylontech/powercubem2/PylontechPowercubeM2BatteryProtectionDefinition.java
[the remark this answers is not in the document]
thomasjbhayes: I think I can resolve this comment - if you can implement later. WDYT?

thomasjbhayes:
As far as I can tell, the only unresolved issue is around the current limits in the BatteryProtection. I think it is best to not include these in BatteryProtection as the current will depend on the number of piles which is a dynamic variable read from the BMS - we will not have access to this when the BatteryProtection is instantiated so better to rely on the value coming from the BMS.
In any case the BMS will provide this protection of the battery as well.

WDYT? @tsicking

thomasjbhayes:
Hello @tsicking ,

Just returning to this PR. Can you please let me know what you think about my logic around the Battery Protection in the comment above?

Thanks,
Thomas

io.openems.edge.battery.pylontech/src/io/openems/edge/battery/pylontech/powercubem2/statemachine/StateMachine.java
```diff
@@ -0,0 +1,70 @@
...
+		}
+	}
+
+	public StateMachine(State initialState) {
+		super(initialState);
+	}
+
+	@Override
+	public StateHandler<State, Context> getStateHandler(State state) {
+		switch (state) {
```
tsicking, line 53: You can use modern switch-syntax here:
```java
return switch (state) {
    case UNDEFINED -> new UndefinedHandler();
   ...
```

io.openems.edge.battery.pylontech/src/io/openems/edge/battery/pylontech/powercubem2/statemachine/GoStoppedHandler.java
```diff
@@ -0,0 +1,40 @@
...
+
+	@Override
+	protected void onEntry(Context context) throws OpenemsNamedException {
+		this.state = BatteryGotoSleepState.WAIT_FOR_SLEEP;
+	}
+
+	@Override
+	public State runAndGetNextState(Context context) throws OpenemsNamedException {
+		
+		switch (this.state) {
```
tsicking, line 24: Modern switch syntax, as in `StateMachine.java`

io.openems.edge.battery.pylontech/src/io/openems/edge/battery/pylontech/powercubem2/statemachine/UndefinedHandler.java
```diff
@@ -0,0 +1,37 @@
...
+import io.openems.edge.common.statemachine.StateHandler;
+
+public class UndefinedHandler extends StateHandler<State, Context> {
+
+	@Override
+	public State runAndGetNextState(Context context) {
+
+		var battery = context.getParent();
+
+		switch (battery.getStartStopTarget()) {
```
tsicking, line 13: Modern switch syntax as in `StateMachine.java`

io.openems.edge.battery.pylontech/src/io/openems/edge/battery/pylontech/powercubem2/statemachine/GoRunningHandler.java
```diff
@@ -0,0 +1,42 @@
...
+
+	@Override
+	protected void onEntry(Context context) throws OpenemsNamedException {
+		this.state = BatterySwitchOnState.WAIT_FOR_WAKE_UP;
+	}
+
+	@Override
+	public State runAndGetNextState(Context context) throws OpenemsNamedException {
+
+		switch (this.state) {
```
tsicking, line 25: Modern switch syntax as in `StateMachine.java`

io.openems.edge.battery.pylontech/src/io/openems/edge/battery/pylontech/powercubem2/PylontechPowercubeM2BatteryImpl.java
```diff
@@ -0,0 +1,862 @@
...
+		return channelId;
+	}
+
+	/**
+	 * Gets modbus offset for a given pileNumber. pileNumber starts from 1.
+	 * 
+	 * @param pileNumber pileNumber The pile the offset is for
+	 * @return The Modbus offset for the pile
+	 */
+	private int getOffsetForPile(int pileNumber) {
```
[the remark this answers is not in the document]
thomasjbhayes: Much simpler!

thomasjbhayes:
Hi @tsicking 

I have updated the PR to reflect your comments above. I have decided to leave the issue of the number of piles / initial battery protection current for later.

I have updated the switch statement syntax etc. 

I have also brought the branch up to date with origin/develop via a merge - there were a few changes to BatteryProtection and other APIs which have required a few changes to the code but all seems good.

The tests and builds are passing now, however 'codecov/project' is failing now.

I have tested the merged code with the physical device and it seems to be ok.

sfeilmeier (approved):
This looks good enough from my side to merge it now - and we can incorporate further improvements on `develop` then. Thanks!

PR #2405, opened 2023-10-28
Feature/io Shelly Pro 3

lukaskirner:
Adding Shelly Pro 3 to available io devices


sfeilmeier:
Thank you! It looks very good on first sight, but has the same flaws as the existing implementations - with possibly blocking the Core Cycle. I'd suggest waiting till #2381 is merged - ok?

lukaskirner:
Servus @sfeilmeier, should be ok. Let me know if I can implement the new Version.

Sn0w3y:
Hello @lukaskirner it is merged now - still some functions missing but it should work. See my Pull here:

[Link](https://github.com/OpenEMS/openems/pull/2496/files)

PR #2260, opened 2023-07-10
Implementation of SumState-Alerting (BREAKING UI+Backend/Odoo changes!) 

da-Kai:
## Description:
Notify specific users of the persistent presence of a negative sumState.

## Planed Features

- [x] React and send mail on sumState
- [x] Improved data structure for growing alerting requirements and settings
- [x] Unit-Tests
- [x] UI settings

### Odoo changes

https://github.com/OpenEMS/odoo-openems/pull/3

PR #2409, opened 2023-10-30
Backend: UI-Websocket: get user.globalRole instead of default guest role on AlertingConfigs

DerStoecki:
This commit is one of many prepare PRs / commits for a bigger PR: SumStateAlert #2260
By separating the big #2260 into smaller commits we would like to make the review easier.

Co-authored-by: Kai Jeschek [99220919+da-Kai@users.noreply.github.com](mailto:99220919+da-Kai@users.noreply.github.com)

### DIFF
--- a/io.openems.backend.uiwebsocket/src/io/openems/backend/uiwebsocket/impl/OnRequest.java
+++ b/io.openems.backend.uiwebsocket/src/io/openems/backend/uiwebsocket/impl/OnRequest.java
@@ -425,7 +425,7 @@ public class OnRequest implements io.openems.common.websocket.OnRequest {
 	/**
 	 * Handles a {@link GetUserAlertingConfigsRequest}.
 	 *
-	 * @param user    {@User} who called the request
+	 * @param user    {@link User} who called the request
 	 * @param request the {@link SetUserAlertingConfigsRequest}
 	 * @return the JSON-RPC Success Response Future
 	 * @throws OpenemsException on error
@@ -435,7 +435,7 @@ public class OnRequest implements io.openems.common.websocket.OnRequest {
 		var edgeId = request.getEdgeId();
 		List<AlertingSetting> users;
 
-		if (user.getRole(edgeId).orElse(user.getGlobalRole()).isLessThan(Role.ADMIN)) {
+		if (userIsAdmin(user, edgeId)) {
 			users = List.of(this.parent.metadata.getUserAlertingSettings(edgeId, user.getId()));
 		} else {
 			users = this.parent.metadata.getUserAlertingSettings(edgeId);
@@ -448,7 +448,7 @@ public class OnRequest implements io.openems.common.websocket.OnRequest {
 	/**
 	 * Handles a {@link SetUserAlertingConfigsRequest}.
 	 *
-	 * @param user    {@User} who called the request
+	 * @param user    {@link User} who called the request
 	 * @param request the {@link SetUserAlertingConfigsRequest}
 	 * @return the JSON-RPC Success Response Future
 	 * @throws OpenemsException      on error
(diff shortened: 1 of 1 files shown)


io.openems.backend.uiwebsocket/src/io/openems/backend/uiwebsocket/impl/OnRequest.java
```diff
@@ -435,7 +435,7 @@ private CompletableFuture<? extends JsonrpcResponseSuccess> handleGetUserAlertin
 		var edgeId = request.getEdgeId();
 		List<AlertingSetting> users;
 
-		if (user.getRole(edgeId).orElse(Role.GUEST).isLessThan(Role.ADMIN)) {
+		if (user.getRole(edgeId).orElse(user.getGlobalRole()).isLessThan(Role.ADMIN)) {
```
clehne, line 438: This and the change in L460 means, that a user, who does not have a role for an edge, can use his global rights for the edge. I think of a usecase where an installer, which has no access to an edge, can access this edge with this installer role. 
As we are embedded in the OpenEMS UI security concept, this is not a problem. 

But maybe there are ways to use the Backend JSON-RPC interface to use this weakness.

@da-Kai Is it really needed to increase the rights here? If so, do you see any security weakness here?

clehne (approved):
Is it really needed so set

io.openems.backend.uiwebsocket/src/io/openems/backend/uiwebsocket/impl/OnRequest.java
da-Kai: @clehne it rather bothers me that I used a low-fixed-value as fallback (_"Defaulting"_) instead of Optional::isPresent.

i would prefer to check whether the user has admin rights using the following method:
`var userIsAdmin = user.getRole(edgeId).map(role -> role.isAtLeast(Role.ADMIN)).orElse(false);`
and to rewrite the if-statements with it.

DerStoecki: Changed this for get and set UserAlertingSettings

Thank you!

da-Kai (changes requested):
thanks @DerStoecki,
this looks way better 👍

io.openems.backend.uiwebsocket/src/io/openems/backend/uiwebsocket/impl/OnRequest.java
```diff
@@ -435,7 +435,7 @@ private CompletableFuture<? extends JsonrpcResponseSuccess> handleGetUserAlertin
 		var edgeId = request.getEdgeId();
 		List<AlertingSetting> users;
 
-		if (user.getRole(edgeId).orElse(Role.GUEST).isLessThan(Role.ADMIN)) {
+		if (userIsAdmin(user, edgeId)) {
```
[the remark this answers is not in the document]
DerStoecki: oof my bad 
thank you :D 
I will change this

da-Kai (approved):

io.openems.backend.uiwebsocket/src/io/openems/backend/uiwebsocket/impl/OnRequest.java
```diff
@@ -474,6 +473,10 @@ private CompletableFuture<? extends JsonrpcResponseSuccess> handleSetUserAlertin
 		return CompletableFuture.completedFuture(new GenericJsonrpcResponseSuccess(request.getId()));
 	}
 
+	private static Boolean userIsAdmin(User user, String edgeId) {
```
sfeilmeier, line 476: Method should return a `boolean` and not `Boolean`

sfeilmeier (changes requested):

sfeilmeier (approved):

PR #2420, opened 2023-11-02
UI: evcscontroller -> automatic mode -> uses sunny-icon again

DerStoecki:
In the past automatic mode used the sunny-outline icon.
This is missing here (idk if it was intentional).


lukasrgr:
@DerStoecki this was not intentional, but youre right

lukasrgr (approved):

PR #2423, opened 2023-11-03
Influx db/get latest value

DerWahreKlinki:
Added unimplemented method getLatestValue for influxDB to get latest value i.e. for a calculated energymeter.
Used by calculateEnergyFromPower

### DIFF
--- a/io.openems.edge.timedata.influxdb/src/io/openems/edge/timedata/influxdb/TimedataInfluxDbImpl.java
+++ b/io.openems.edge.timedata.influxdb/src/io/openems/edge/timedata/influxdb/TimedataInfluxDbImpl.java
@@ -215,41 +215,32 @@ public class TimedataInfluxDbImpl extends AbstractOpenemsComponent
 		// TODO implement this method
 		return emptySortedMap();
 	}
-	
+
 	@Override
 	public CompletableFuture<Optional<Object>> getLatestValue(ChannelAddress channelAddress) {
 	    return CompletableFuture.supplyAsync(() -> {
 	        try {
-	            var result = this.influxConnector.queryLastData(
-	                Optional.empty(), channelAddress, this.config.measurement());
+	            SortedMap<ChannelAddress, JsonElement> sortedMap = this.influxConnector.queryLastData(Optional.empty(), channelAddress, this.config.measurement());
 
-	            // If there are no results, return empty immediately
-	            if (result.isEmpty()) {
+	            if (sortedMap != null && !sortedMap.isEmpty() && sortedMap.containsKey(channelAddress)) {
+	                JsonElement latestValue = sortedMap.get(channelAddress);
+
+	                // Check if it´s a number and can be converted to long
+	                if (latestValue.isJsonPrimitive()) {
+	                	if (latestValue.getAsJsonPrimitive().isNumber()) {
+	                		return Optional.of(latestValue.getAsLong());
+	                	}
+	                }
+	            } else {
+	                // No data found
 	                return Optional.empty();
 	            }
-	            
-	            var latestValues = result.get(result.lastKey());
-	            
-	            // If the channelAddress is not in the latest values, return empty
-	            if (!latestValues.containsKey(channelAddress)) {
-	                return Optional.empty();
-	            }
-	            
-	            // Retrieve the latest value from the result
-	            var element = latestValues.get(channelAddress);
-
-	            // If the element is a number, return its long value
-	            if (element.isJsonPrimitive() && element.getAsJsonPrimitive().isNumber()) {
-	                return Optional.of(element.getAsLong());
-	            }
-	            
-	        } catch (OpenemsNamedException e) {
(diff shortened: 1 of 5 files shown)


sfeilmeier:
@michaelgrill Could you please have a look?

io.openems.edge.timedata.influxdb/src/io/openems/edge/timedata/influxdb/TimedataInfluxDbImpl.java
```diff
@@ -216,17 +216,39 @@ public SortedMap<Long, SortedMap<ChannelAddress, JsonElement>> queryResendData(Z
...
+	            SortedMap<ZonedDateTime, SortedMap<ChannelAddress, JsonElement>> result = this.influxConnector.queryLastData(
+	                Optional.empty(), channelAddress, this.config.measurement());
+
+	            // Retrieve the latest value from the result
+	            if (!result.isEmpty()) {
+	                SortedMap<ChannelAddress, JsonElement> latestValues = result.get(result.lastKey());
+	                if (latestValues.containsKey(channelAddress)) {
+	                    JsonElement element = latestValues.get(channelAddress);
+	                    if (element.isJsonPrimitive() && element.getAsJsonPrimitive().isNumber()) {
+	                    	return Optional.of(element.getAsLong());
```
michaelgrill, line 234: what i usually prefer to avoid nested if statements in general is to just return inside ifs so the first if here wouldnt be 
`if (!result.isEmpty()) { ... }` it would be like `if (result.isEmpty()) { return Optional.empty(); } ...`

but thats up to you 3 nestings isnt that bad.

[More detailed description](https://stackoverflow.com/a/47619546)

io.openems.shared.influxdb/src/io/openems/shared/influxdb/proxy/QueryProxy.java
```diff
@@ -132,6 +132,27 @@ public abstract SortedMap<ZonedDateTime, SortedMap<ChannelAddress, JsonElement>>
...
+	 * @param influxConnection a Influx-Connection
+	 * @param bucket           the bucket name; 'database/retentionPolicy' for
+	 *                         InfluxDB v1
+	 * @param measurement      the influx measurement
+	 * @param influxEdgeId     the Edge-ID
+	 * @param channel          the {@link ChannelAddress}
+	 * @return the query result
+	 * @throws OpenemsNamedException on error
+	 */
+	public abstract SortedMap<ZonedDateTime, SortedMap<ChannelAddress, JsonElement>> queryLastData(
```
michaelgrill, line 147: i think this method is pretty similar to the one below `queryFirstValueBefore` (the name of the method may be a little bit off) can we somehow merge those?

michaelgrill (changes requested):
Thanks for implementing. The last comment would be important the others are just minor things

io.openems.shared.influxdb/src/io/openems/shared/influxdb/proxy/QueryProxy.java
DerWahreKlinki: yeah...no...maybe... queryFirstValueBefore is used by many modules. As I am not a real java programmer I don´t know if I can provide this method. 
What about a compromise: you check in my implementation and I´ll put "queryFirstValueBefore" on my ToDo list?

michaelgrill: only because its used in other models doesnt mean you cant use it or have to provide it only the flux implementation is missing currently but i could help you with that.

I just dont want to add a method which already exists. It would just lead to more confusion i guess and its also harder to maintain. 

What we also can do is to keep the method in the InfluxConnector and call the `queryFirstValueBefore` method inside but we pass the one channel as a `Set` and just the current time. It would look like this i guess:
```
this.queryProxy.queryFirstValueBefore(this.bucket, this.getInfluxConnection(), measurement, influxEdgeId,
				ZonedDateTime.now(), Set.of(channelAddress));
```

DerWahreKlinki: Hi Michael,
I´ve re-coded the pull-request. 
- InfluxConnector: queryLastData makes use of the queryFirstValueBefore-method
- FluxProxy: added buildFetchFirstValueBefore and convertFirstValueBeforeQueryResult
- TimedataInfluxDbImpl: added method getLatestValue

michaelgrill (changes requested):
Some last things... but no breaking change anymore. Thank you so much for your patience and after those last things i think we can merge 😄

DerWahreKlinki:
Hi Michael,
Auto-Formatting feature was new to me. I was just used to the checkstyle plugin. Sorry.

FluxProxy now uses parameter relative value for start time.

michaelgrill (approved):
Thank you 😄 👍

PR #2427, opened 2023-11-09
remove `addChatterMessage` from OdooUtils and OdooHandler

da-Kai:
The Odoo Chatter implementation is not working.
As it is currentrly also not used, i would suggest removing it.
What is your opinion @sfeilmeier ?


sfeilmeier (approved):
Agree. Thank you.

It will anyway be smarter in future to store any information in a "structured" way and not in the chatter.

PR #2428, opened 2023-11-10
OEM-Bundle: implement a OpenEMS OEM bundle

sfeilmeier:
This approach should make it easier to handle proprietary distributions of OpenEMS Edge and Backend


sfeilmeier:
@clehne: As discussed previously. Would this approach suite your requirements?

clehne (changes requested):
Like that. It is a really good starting point for an `oem` provider. I only have some "technical" change requests. 

My thoughts went in a similar direction. Internally, we have a rudimentary prototype of an own oem provider, which I wanted to publish later. Luckily it is not far away from this PR. It provides us some more information:

1. a provider for APP Urls, because we add a `setWebsiteURL(..)` to every App (see example [here](https://github.com/OpenEMS/openems/blob/cc5ccddccd38799f0c2fd503dd2869cdb252f94c/io.openems.edge.core/src/io/openems/edge/app/evcs/KebaEvcs.java#L155) (this makes syncing a pain.) 

2. we removed AggregatedInflux:[AllowedChannels](https://github.com/OpenEMS/openems/blob/cc5ccddccd38799f0c2fd503dd2869cdb252f94c/io.openems.backend.timedata.aggregatedinflux/src/io/openems/backend/timedata/aggregatedinflux/AllowedChannels.java#L12) and provide the list of averageChannels and cumulatedChannels via oemProvider.

3. Also we made kind of a whitelist for the influxDb, which is delivered by our oemProvider implementation. This helps us keep our productive database clean, because it forces us to take care of `componentId`s. 

Would be nice if we can get at least 1. and 2. into the Oem provider. But I would do this in further PRs. 
What do you think about 3. a whitelist for influx db? It is probably not your usecase. But could it be useful to others.

io.openems.edge.core/src/io/openems/edge/app/common/props/CommonProps.java
```diff
@@ -109,8 +108,11 @@ PARAM extends BundleParameter> AppDef<APP, PROP, PARAM> installationHintOfUnoffi
 				(app, property, l, parameter) -> TranslationUtil.getTranslation(parameter.bundle,
 						"unofficialAppWarning.text1"), //
 				(app, property, l, parameter) -> TranslationUtil.getTranslation(parameter.bundle,
-						"unofficialAppWarning.text2")),
-				def -> def.setAutoGenerateField(!OpenemsConstants.MANUFACTURER.equals("OpenEMS Association e.V.")));
+						"unofficialAppWarning.text2"))
+		// TODO ,def ->
+		// def.setAutoGenerateField(!OpenemsConstants.MANUFACTURER.equals("OpenEMS
+		// Association e.V."))
+		);
```
sfeilmeier, line 115: @michaelgrill I need your support here, how to solve this.

io.openems.common/src/io/openems/common/oem/DummyOpenemsBackendOem.java
```diff
@@ -0,0 +1,13 @@
...
+ * A default {@link DummyOpenemsBackendOem} for OpenEMS Backend.
+ */
+public class DummyOpenemsBackendOem implements OpenemsBackendOem {
+
+	@Override
+	public String getInfluxdbTag() {
+		return "edge";
+	}
+
+}
```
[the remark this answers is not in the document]
sfeilmeier: Good point. A check for this was just introduced this week https://github.com/OpenEMS/openems/pull/2431 by our friends from Japan. Seems, there is a similar rule for Checkstyle -> https://checkstyle.sourceforge.io/checks/misc/newlineatendoffile.html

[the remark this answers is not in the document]
clehne: Great hint. Thanks.

io.openems.edge.core/src/io/openems/edge/app/common/props/CommonProps.java
michaelgrill: As i already talked with you about this in general handling apps, @clehne may also have an opinion on this. We were discussing to move the apps maybe completely to the OEM specific bundle then everyone has their own "available" apps, but it would probably result in a lot of duplicated code when not sharing a simple app like "KEBA EVCS" or so.

For now you can just remove those lines 🙂

sfeilmeier: Thanks Michael. As also Christians feedback was positive, I'll continue with this PR to have a first OEM bundle. From there we'll discuss individually how we extract the remaining OEM specific changes like AggregatedInflux, Apps, etc.

clehne: > As i already talked with you about this in general handling apps, @clehne may also have an opinion on this. We were discussing to move the apps maybe completely to the OEM specific bundle then everyone has their own "available" apps, but it would probably result in a lot of duplicated code when not sharing a simple app like "KEBA EVCS" or so.
> 
> For now you can just remove those lines 🙂

@michaelgrill Right now I do not have a final opinion on apps handling. I think also that we should finish this PR and delay the app handling to a new PR. 

In general it would be good to 

- have a common base for most of the apps. 
- have a OEM specific link for each app url and probably for the app image.
- have the opportunity for own apps. 

The keba EVCS app is a good example. We tried to use it but this leads to problems with our evcs cluster app for charge parks. Later we needed a lot of workarounds for e.g. alfen, compleo apps just because of the keba evcs app.
In the end we  decided to put our own different keba app beside the official keba app. 
I think this is a good compromise for now, as your and our company proceed with different speeds. But later we should find a way to integrate both sets of apps into one base.

PR #2429, opened 2023-11-11
fix: convert new line code CRLF -> LF in ui/ files.

miettal:
and insert new entry of gitattribute for web-dev related files suffixes.

Actually I applied this command to ui/ files. // note: this sed is bsd sed, not gnu sed.
 - find . -type f -name "*.json" -exec sed -i '' 's/\r$//' {} \\;
 - find . -type f -name "*.ts" -exec sed -i '' 's/\r$//' {} \\;

for increasing maintenanceability I changed new line code.


lukasrgr:
@miettal i think you could also do it like described here, would be better for the future: [git docs](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings#global-settings-for-line-endings)

miettal:
as you says, git has new line code conversion function. but its only works when checkout/commit.
In that case, if you have configured the settings, new line code will be fix at next commit, but the function changing and new code changing will be displayed together as a diff.
So if I change the code all at once like this PR, the next commit will have minimal diff.
this is reason of making this PR and just my suggestion.

just using "core.autocrlf input" is one method too. if project judge enougth to use "core.autocrlf input", I just fllow that ;)

miettal:
@lukasrgr Sorry There was one mistake and I misunderstood your comment.
Actually I added modification to .gitattributes file to force LF setting in this PR first-commit,
but in rebase process, mistakenly I dropped this modification. and I didnt realize it.

Immedietry I'll add this modification to .gitattribute.

miettal:
done

sfeilmeier (approved):

PR #2430, opened 2023-11-11
fix: use tailing comma to always-multiline

miettal:
eslint src/ --fix

1. Typically, using trailing comma minimize diff and easy to review, merging.
2. on the other hand, snipping trailing comma has compatibility old javascript implementation.

in this project, /ui codes are using typescript so its not need to care about 2.
so I changed tailing comma style.

### DIFF
--- a/ui/.eslintrc.json
+++ b/ui/.eslintrc.json
@@ -59,7 +59,7 @@
         "curly": "error",
         "comma-dangle": [
           "error",
-          "never"
+          "always-multiline"
         ],
         "eol-last": [
           "error",
--- a/ui/src/app/app-routing.module.ts
+++ b/ui/src/app/app-routing.module.ts
@@ -68,13 +68,13 @@ const routes: Routes = [
       {
         path: 'live', data: { navbarTitle: environment.uiTitle }, providers: [{
           useClass: LiveDataService,
-          provide: DataService
-        }], component: EdgeLiveComponent
+          provide: DataService,
+        }], component: EdgeLiveComponent,
       },
       {
         path: 'history', providers: [{
           useClass: HistoryDataService,
-          provide: DataService
+          provide: DataService,
         }], component: HistoryParentComponent, children: [
           { path: '', component: EdgeHistoryComponent },
           // History Chart Pages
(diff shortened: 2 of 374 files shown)


sfeilmeier (approved):

sfeilmeier:
I fully agree with your reasoning and approach. Thanks for the contribution! 👍

sfeilmeier (changes requested):
Unfortunately there are merge conflicts now and I do not have write permissions on your branch. Please fix and request review again. Thanks!

miettal:
Thank you for review, I'll rebase this PR to latest commit.

sfeilmeier (approved):

lukasrgr:
@miettal i dont agree, that this will be easier to review, reducing it from a two line change to one line doesnt make it easier, why is this change necessary?

miettal:
Typically less diff makes it easier to review.
and it help to keeping line history(like git blame).

miettal:
@lukasrgr I found good material of trailing comma explanation! this is easy to understand!
https://medium.com/@nikgraf/why-you-should-enforce-dangling-commas-for-multiline-statements-d034c98e36f8

lukasrgr:
@miettal im not fully convinced but if reviewing etc. is really an issue for others, then i have to agree

miettal:
// there are many code styling and mindset of code styling. no right way and it's also depends on environment / project. Anyway let's experience with everyone!

ikegam:
Jump in from same company as @miettal,

"trailing comma" is just a style for keeping same style for every lines.
Traditional language like C, enum of Java can not support this style but js can support so we made this PR. It can make unified diff for all elements in structure. W/o trailing comma, if we touch last element it makes two lines, but middle line makes only one line as you know. This is only advanced point as I know.

We, Girasol Energy, are just downstream of yours so we will follow your decision.

BTW, just my feeling, it seems mixed up in same directory with different newline code which are /n and /r/n is more likely to be fixed I guess.

ikegam:
Just I asked @miettal to rebase and modify LF / CRLF related PR. please wait some days.

PR #2432, opened 2023-11-13
Improve Alerting tests using Clock

da-Kai:
* Change Alerting to use a Clock instance for retrieving current dateTime
* Rewrite tests to be time independent


io.openems.backend.alerting/src/io/openems/backend/alerting/Alerting.java
```diff
@@ -38,66 +42,83 @@
 })
 public class Alerting extends AbstractOpenemsBackendComponent implements EventHandler {
 
+	// Maximum number of messages constructed at the same time
 	private static final byte THREAD_POOL_SIZE = 2;
+	// Queue size from which warnings are issued
 	private static final byte THREAD_QUEUE_WARNING_THRESHOLD = 50;
 
-	protected final Scheduler scheduler;
+	private static final Executor createDefaultExecutorService() {
```
sfeilmeier, line 50: @michaelgrill: This is something we use in a lot of other places in Backend - but in a different way. Do you see room for synergies/code reuse?

da-Kai: @michaelgrill, any suggestions or comments?

michaelgrill: sry for the late answer @da-Kai as we discussed it would would make sense to only have one Thread for debug messages and somewhere a central component which executes them every 10 sec.

The only question is can we do this also in this PR or should we make a separate one

sfeilmeier (approved):

PR #2439, opened 2023-11-21
Allow unauthenticated access to /live on Backend2Backend REST API

phfeustel:
**Motivation**
When running the OpenEMS backend in (docker) containers and/or connect them via (cloud) load balancers, it is handy to have an easy to call liveness probe which does not need any authentication.

_Side note_: As the `Backend2BackendRest` service is also reliant on an up-and-running `Metadata`service, I guess this check could even serve as a readiness probe, even though the `Metadata` isn't used actually.

### DIFF
--- a/cnf/pom.xml
+++ b/cnf/pom.xml
@@ -414,5 +414,17 @@
 			<artifactId>rrd4j</artifactId>
 			<version>3.9</version>
 		</dependency>
+		<dependency>
+			<!-- Testing e.g. b2brest.RestHandlerTest-->
+			<groupId>org.assertj</groupId>
+			<artifactId>assertj-core</artifactId>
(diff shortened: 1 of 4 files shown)


tsicking:
Hi Phil and Chris,
While I am honoured that I have been chosen as a reviewer for this PR, I have to say that I work predominantly on the OpenEMS edge, and not on the backend. All I can say is that
1) "Allow unauthenticated access" seems a bit dangerous to me
2) Two dependencies are added to `pom.xml`. I am always very careful with changing this file.
I can't judge whether the impact of this PR justifies that, but I would be very careful with that.

sfeilmeier:
@michaelgrill What's your opinion on this?

PR #2445, opened 2023-11-28
UI: Methods in shared/EdgeConfig.ts returning EdgeConfig.Component[]  -> sort via component alias

DerStoecki:
## Description

src/app/shared/edge/edgeconfig.ts -> methods returning an EdgeConfig.Component[] sort their content via component.alias.

## Methods 

- getComponentsByFactory
- getComponentsImplementingNature

## Reason

This is mainly a beautification within the Flat/Modal Widgets of Live and History.

We got feedback of users, and they find it more visually appealing, when the widget contents are sorted by their "name" (alias)
This simple sorting would solve this request.

This may be in your interest too.

Co-authored-by: Ulrich Laupp <38915524+laupp@users.noreply.github.com>

### DIFF
--- a/ui/src/app/shared/edge/edgeconfig.ts
+++ b/ui/src/app/shared/edge/edgeconfig.ts
@@ -180,7 +180,7 @@ export class EdgeConfig {
         for (let componentId of componentIds) {
             result.push(this.components[componentId]);
         }
-        return result;
+        return result.sort((c1, c2) => c1.alias.localeCompare(c2.alias));
     }
 
     /**
@@ -230,7 +230,7 @@ export class EdgeConfig {
                 result.push(...this.getComponentsImplementingNature("io.openems.edge.meter.api.SymmetricMeter"));
         }
 
-        return result;
+        return result.sort((c1, c2) => c1.alias.localeCompare(c2.alias));
     }
 
     /**


lukasrgr:
i dont think this makes sense, seperating the "common" widgets and controller specific makes sense, alias shouldnt be deciding in my opinion. I think it makes sense for controller widgets.

lukasrgr:
i wouldnt do the sorting here, you could do this in widget.ts

DerStoecki:
Hey,

Thank you for your input and sorry for my late reply.

I thought about your comment and I think we should discuss on this topic a little bit more (or shift the discussion to the community forum if this is preferred.)

I'd like to know:
* why a general sort with alias is not a good idea, since (especially the common) widgets display their components with the corresponding alias and value, instead of the componentId (Examples are given when click on the details tag)?
* Does this break something, that I am not aware of?

But let me explain our usecase a bit more in detail.

This change should only change the sort order of components _within_ a widget and shouldn't sort the widgets themself.

Currently the widgets are unsorted / picked up by component Id. Resulting in some randomness (for the user) as you can see within the following details tag, especially if they adapt the component alias, resulting in this:

<details>
<summary> **Click Me** to show some pictures showing the difference between unsorted/sorted by alias</summary>

## Unsorted


### History Flat Widget

UnsortedHistoryWidget


### History Chart

UnsortedHistoryChart


### Live Flat Widget

UnsortedLiveFlatWidget


Therefore I made the change within the edgeconfig.ts

Resulting in this:

## Sorted

### History Flat Widget

SortedHistoryFlatWidget



### History Chart

SortedHistoryChart

### Live Flat Widget

SortedFlatWidget

</details>


Additionally:

You suggested, that I could make those changes in the widget.ts 
However: The sorting within the widget contents wouldn't work.

If I'd change the widget.ts and do a sorting by alias in the same way you sort the channelthreshold by the output address like:


<details>
<summary> explicitely sort ChannelThresholdControllers by their outputChannelAddress </summary>

https://github.com/OpenEMS/openems/blob/f6aa692fc5f97961b6c06c16452f51265cc9c8f3/ui/src/app/shared/type/widget.ts#L104-L116

</details>

I wouldn't change the sort order of  components within widgets, as you can see here for example:

<details>
<summary>  /ui/src/app/edge/live/common/production/flat/flat.ts </summary>

https://github.com/OpenEMS/openems/blob/f6aa692fc5f97961b6c06c16452f51265cc9c8f3/ui/src/app/edge/live/common/production/flat/flat.ts#L23-L26

</details>
Where the flat.ts calls the method directly to get Components implementing Nature X, rather than picking the component from widgets list.

If I shouldn't create a general sort by alias,  I'd need to go through every flat/modal/history widget/modal/chart and see if I'd need to sort the components by their alias, since they display different components.

My personal conclusion is, that a general sorting by alias makes sense, since the UI displays components showing their alias (especially the common widgets/charts).

I hope this made things more clear. 
Thank you for your input in advance.

sfeilmeier:
@DerStoecki Thanks for the detailed explanation. This sounds reasonable to me. I'll leave it to @lukasrgr to decide if the technical solution is the best way to do it.

lukasrgr:
Hey, i misunderstood it then, pretty good explanation, i think you actually have a good solution here. I just want to remind, that its still not guaranteed that the alias is set, especially for older systems. Use the helper var Name.METER_ALIAS_OR_ID.
so the sorting would do ```ids``` as well as ```alias```, just to be clear

DerStoecki:
Hey @lukasrgr ,

Thank you for your input. I adapated the comparison as you suggested. 
Thank you!

lukasrgr (commented):
Perfect

lukasrgr (approved):

PR #2468, opened 2023-12-20
fix: reproduce single_document.adoc from nav.adoc and sed script

miettal:
This PR changes reparing single document generating shell script and supporting nav.adoc-based single document generation without management single_document.adoc

background:

Occording history of git-log, there was single_document.adoc to generate single page document output, but currently is not. so create_single_page_{pdf, html}.sh fails.

// I and ikegam guessed the reason is avoiding duplication of managing nav.adoc and single_document doc. because almost user mainly use multiple page document(based-on nav.adoc) on webui.

idea:

Therefore We forcus to difference nav.adoc and single_document.adoc and plant to re-use nav.adoc as single_documet.adoc by using sed because there are only little difference.

This commit is maded by multiple contributors

### DIFF
--- a/doc/tools/create_single_page_html.sh
+++ b/doc/tools/create_single_page_html.sh
@@ -7,4 +7,26 @@ cd ${DIR}/../modules/ROOT/pages
 
 echo "Creating ${DIR}/../modules/ROOT/pages/single_document.html"
 
-asciidoctor -b html5 -o single_document.html single_document.adoc
+SED_ARG="s/xref:/include::/"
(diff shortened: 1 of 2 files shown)


miettal:
As example, I share generated pdf
[single_document.pdf](https://github.com/OpenEMS/openems/files/13727824/single_document.pdf)

sfeilmeier (approved):

sfeilmeier:
Thank you for reactivating the single page!

PR #2489, opened 2024-01-17
UI: modal-line element INPUT with optional unit

JosefRick:
Implements a new modal-line element INPUT_UNIT. It's the similar to INPUT with the option to change the unit with the following code line. 
eg: [control]="{type: 'INPUT_UNIT', properties: {unit: 'mA'}}">

### DIFF
--- a/ui/src/app/shared/genericComponents/modal/modal-line/modal-line.html
+++ b/ui/src/app/shared/genericComponents/modal/modal-line/modal-line.html
@@ -29,7 +29,17 @@
                     <ion-label class="regularFont">&nbsp;W&nbsp;</ion-label>
                 </ion-item>
             </td>
-
+          
+            <!-- 'Input' Line, name and changeable value with unit -->
+            <td style="width: 30%" class="align_right" *ngIf="control.type === 'INPUT_UNIT'" [formGroup]="formGroup">
+                <ion-item>
+                    <ion-input class="regularFont" type="number" style="text-align: right;"
+                        [formControlName]="controlName">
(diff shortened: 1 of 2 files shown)


lukasrgr (changes requested):
this is a design error, i would add a optional unit to the control object and use "W" as fallback

JosefRick:
Yes, that is a better option. I have changed it now. Please take a look at it.

lukasrgr (changes requested):
i think there is still a cleaner solution for this. Create a Enum e.g. InputUnits, that holds in this case only WATT = 'W'. Use this then as the type for the unit. Define the defaultValue in typescript then

JosefRick:
I'm not sure if this is really a clean solution, because for the control RANGE there is also the same way to define the unit if i understand it correct. So if we do it with Enum now, there are different possibilities for each control and so I think it is more difficult to understand which definition is the right one for each control. An alternative idea is, we make a new unit.ts in the shared folder so that every unit definition on the genericComponents can use it and it is defined for all at one point.

lukasrgr:
@JosefRick so you would define a fallback type "W" in this, do i understand correctly?

JosefRick:
@lukasrgr Yes, with this code I have defined a fallback 'W' if no properties are specified in the control type.
 For example _[control]="{type: 'INPUT', properties: {unit: 'mA'}}"_ in the modal.html will display the unit _mA_, and if no properties are specified, like this _[control]="{type: 'INPUT'}"_ the code will use _W_ as the unit.

PR #2493, opened 2024-01-25
Minor corrections to the edge component implementation guide

samuelBloch:
There was a small formatting error in section the 1.4. Implement the OpenEMS Component and the wrong button name in section 1.7. Enable the Component.


sfeilmeier (approved):
Thanks!

PR #2494, opened 2024-01-25
Improvement Meter B Control

tsicking:
The driver of the Meter "B-Control" is improved in the following aspects:
1. The modbus tasks were combined to longer tasks reducing the number of overall tasks. We faced issues where only two of the three phase currents were present and therefore the overall current was calculated incorrectly. This might have happened because the phase currents were read out in different tasks.
2. The frequency was read out from the wrong register. This has been fixed.
3. Energy values per phase were added.
4. The listeners for calculating the overall current and the average voltage were installed twice, once in the constructor and once in the method `addCalculateChannelListeners()`. The latter one has been removed.
5. The method `addCalculateChannelListeners()` is now called in the `activate()`-method instead of the `defineModbusProtocol()`-method. This effectively doesn't make a difference, but I wouldn't expect a listener to be installed in `defineModbusProtocol()`.
6. The private variable `meterType` has been removed, as the config is saved anyway and one gets the meter type from there.

Co-authored-by:  Thomas Sicking <91258335+tsicking@users.noreply.github.com>

### DIFF
--- a/io.openems.edge.meter.bcontrol.em300/src/io/openems/edge/meter/bcontrol/em300/MeterBControlEM300Impl.java
+++ b/io.openems.edge.meter.bcontrol.em300/src/io/openems/edge/meter/bcontrol/em300/MeterBControlEM300Impl.java
@@ -22,6 +22,7 @@ import io.openems.edge.bridge.modbus.api.AbstractOpenemsModbusComponent;
 import io.openems.edge.bridge.modbus.api.BridgeModbus;
 import io.openems.edge.bridge.modbus.api.ModbusComponent;
 import io.openems.edge.bridge.modbus.api.ModbusProtocol;
+import io.openems.edge.bridge.modbus.api.element.DummyRegisterElement;
 import io.openems.edge.bridge.modbus.api.element.UnsignedDoublewordElement;
 import io.openems.edge.bridge.modbus.api.element.UnsignedQuadruplewordElement;
 import io.openems.edge.bridge.modbus.api.task.FC3ReadRegistersTask;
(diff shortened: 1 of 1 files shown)


io.openems.edge.meter.bcontrol.em300/src/io/openems/edge/meter/bcontrol/em300/MeterBControlEM300Impl.java
```diff
@@ -130,43 +133,75 @@ protected ModbusProtocol defineModbusProtocol() throws OpenemsException {
...
 						m(ElectricityMeter.ChannelId.VOLTAGE_L3, new UnsignedDoublewordElement(142))));
 
 		if (this.config.invert()) {
-			modbusProtocol.addTask(new FC3ReadRegistersTask(512, Priority.LOW, //
-					m(ElectricityMeter.ChannelId.ACTIVE_CONSUMPTION_ENERGY, new UnsignedQuadruplewordElement(512),
-							SCALE_FACTOR_MINUS_1),
-					m(ElectricityMeter.ChannelId.ACTIVE_PRODUCTION_ENERGY, new UnsignedQuadruplewordElement(516),
-							SCALE_FACTOR_MINUS_1)));
+			modbusProtocol.addTasks(//
+					new FC3ReadRegistersTask(512, Priority.LOW, //
```
sfeilmeier, line 161: There is quite some duplicated code. We could consider implementing it in such a way - but it's up to you.

```
		var inv = this.config.invert();
		new FC3ReadRegistersTask(512, Priority.LOW, //
				m(inv//
						? ElectricityMeter.ChannelId.ACTIVE_CONSUMPTION_ENERGY //
						: ElectricityMeter.ChannelId.ACTIVE_PRODUCTION_ENERGY), //
						new UnsignedQuadruplewordElement(512), SCALE_FACTOR_MINUS_1),
```

I am merging it now anyway, because the PR is very good the way it is. 👍

sfeilmeier (approved):

PR #2504, opened 2024-01-30
Add Ireland and UK as options in OpenEMS

thomasjbhayes:
Ireland and UK were not options as countries in OpenEMS.

This commit adds them within the UI enums, i8n, and the SafetyCountry enum in OpenEMS edge.core.


venu-sagar (approved):

sfeilmeier:
@thomasjbhayes: What is the use-case to add the countries?

Problem is, that e.g. `io.openems.edge.core/src/io/openems/edge/app/enums/SafetyCountry.java` is used in FENECON Home Apps currently - but they are not officially allowed in Ireland and UK.

In UI the change is not critical, but for certain OpenEMS distributions (like FEMS) it might be feasible to restrict the selectable countries.

thomasjbhayes:
Hi Stefan,

Please excuse my delay in getting back to you.

The use case is that we plan to distribute our ESS using OpenEMS in Ireland and the UK and found that the countries were not selectable in the UI. Do the proposed changes cause any issue for Fenecon? 

Thomas

PR #2505, opened 2024-02-01
Update InitializeEdgesWorker.java so it Caches the Edges every 20 Minutes

Sn0w3y:


sfeilmeier:
@Sn0w3y: Is this ready for review?

Sn0w3y:
> @Sn0w3y: Is this ready for review?

Yes, would be ! :) Nothing much needed to change in fact

PR #2517, opened 2024-02-05
Implement Shelly 3 Pro Relais-"Board"

Sn0w3y:
Implementing Shelly 3Pro using the new HTTP Bridge.

[Shelly3Pro](https://shelly-api-docs.shelly.cloud/gen2/0.14/Devices/ShellyPro3/)


Sn0w3y:
Example JSON:

Relais 1:
    {
        "id": 0,
        "source": "init",
        "output": False,
    }
Relais 2:
    {
        "id": 1,
        "source": "init",
        "output": True,
    }
Relais 3:
    {
        "id": 2,
        "source": "init",
        "output": False,
    }

Sn0w3y:
@michaelgrill please check usage of HttpBridge again :)

io.openems.edge.io.shelly/src/io/openems/edge/io/shelly/shelly3pro/IoShelly3ProImpl.java
```diff
@@ -0,0 +1,185 @@
...
+			return;
+		}
+
+		switch (event.getTopic()) {
+		case EdgeEventConstants.TOPIC_CYCLE_EXECUTE_WRITE //
+			-> this.eventExecuteWrite();
+		}
+	}
+
+	private void processHttpResult(JsonElement result, Throwable error) {
```
michaelgrill, line 123: make sure properly handle everything inside this method if you use the method `HttpBridge.subscribeJsonEveryCycle(String,BiConsumer<JsonElement, Throwable>)` to handle only exactly one parameter is not null then the following test would fail
```
	@Test
	public void testProcessHttpResult() throws Exception {
		final var shelly = new IoShelly3ProImpl();
		new ComponentTest(shelly) //
				.addReference("httpBridge", new DummyBridgeHttp()) //
//		.addReference("httpBridgeFactory", new DummyBridgeHttpFactory()) //
				.activate(MyConfig.create() //
						.setId(COMPONENT_ID) //
						.setIp("127.0.0.1") //
						.build()) //
		;

		shelly.processHttpResult(null, new RuntimeException("Some http error"));
	}
```

you can also use the method `BridgeHttp.subscribeJsonEveryCycle(String, ThrowingConsumer<JsonElement, Exception>, Consumer<Throwable>)` then you always have a none null parameter and the exception would be properly handled. e. g.

```
this.httpBridge.subscribeJsonEveryCycle(url, result -> {
    // ... handle result
}, error -> {
    // ... handle error (also gets triggered if the result handle method throws an exception)
});
```

but thats up to you which version you want to use.

michaelgrill (changes requested):
Could you please check if this version has the current changes from develop. I had import issues with the `BridgeHttpFactory`

Sn0w3y:
@michaelgrill please check again :)

PR #2518, opened 2024-02-05
Update IoShelly25Impl to use HttpBridge

Sn0w3y:
Updated Shelly2.5 to use the new HttpBridge from @michaelgrill


Sn0w3y:
@michaelgrill is the usage correct ?

io.openems.edge.io.shelly/src/io/openems/edge/io/shelly/shelly25/IoShelly25Impl.java
```diff
@@ -114,54 +108,55 @@ public void handleEvent(Event event) {
...
-		} catch (OpenemsNamedException | IndexOutOfBoundsException e) {
-			relay1IsOn = null;
-			relay2IsOn = null;
-			this.logError(this.log, "Unable to read from Shelly API: " + e.getMessage());
-			this._setSlaveCommunicationFailed(true);
-		}
-		this._setRelay1(relay1IsOn);
-		this._setRelay2(relay2IsOn);
+		String url = this.baseUrl + "/status";
+		this.httpBridge.get(url).whenComplete((result, error) -> {
```
michaelgrill, line 112: for reading values this should not be called every cycle. The single use methods `get`, `post`, `delete`... are intended to be used rarely and only in certain situations.

In your case you want to read the value every cycle then you can use the `HttpBridge.subscribeJsonEveryCylce` method. Make sure to also remove unused subscribes to events after the change.

You can also have a look at the implemention in the [IoShelly3EmImpl.java](https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.io.shelly/src/io/openems/edge/io/shelly/shelly3em/IoShelly3EmImpl.java)

io.openems.edge.io.shelly/src/io/openems/edge/io/shelly/shelly25/IoShelly25Impl.java
```diff
@@ -100,68 +113,84 @@ public void handleEvent(Event event) {
...
 
+		try {
+			JsonElement jsonElement = JsonUtils.getAsJsonElement(result);
+			final var relays = JsonUtils.getAsJsonArray(jsonElement, "relays");
+			for (int i = 0; i < relays.size(); i++) {
+				final var relay = JsonUtils.getAsJsonObject(relays.get(i));
+				final var relayIson = JsonUtils.getAsBoolean(relay, "ison");
+				final var relayOverpower = JsonUtils.getAsBoolean(relay, "overpower");
+				final var relayOvertemp = JsonUtils.getAsBoolean(relay, "overtemperature");
+				this.digitalOutputChannels[i].setNextWriteValue(relayIson);
```
michaelgrill, line 144: Its wrong to set the write value here. The write value is only for setting a value which then should be written to the external device. In the "read" method you only want to set the value you re reading which you re already doing with `this._setRelay1(relay1IsOn)`. Most of the times the write value is only set by other components which are using the relay to trigger something and not the component which declares it.

io.openems.edge.io.shelly/src/io/openems/edge/io/shelly/shelly25/IoShelly25Impl.java
```diff
@@ -100,68 +113,84 @@ public void handleEvent(Event event) {
...
 		if (Objects.equals(readValue, writeValue.get())) {
-			// read value = write value
 			return;
 		}
-		this.shellyApi.setRelayTurn(index, writeValue.get());
+		final String url = this.baseUrl + "/relay/" + index + "?turn=" + (writeValue.get() ? "on" : "off");
+		this.httpBridge.get(url).whenComplete((t, e) -> {
+			if (e != null) {
+				this.logError(this.log, "HTTP request failed: " + e.getMessage());
+				this._setSlaveCommunicationFailed(true);
```
michaelgrill, line 192: i would probably remove the `_setSlaveCommunicationFailed` here because it gets set in the read method which is triggered every cycle

michaelgrill (changes requested):

io.openems.edge.io.shelly/src/io/openems/edge/io/shelly/shelly25/IoShelly25Impl.java
```diff
@@ -75,20 +89,19 @@ public BooleanWriteChannel[] digitalOutputChannels() {
...
-		for (WriteChannel<Boolean> channel : this.digitalOutputChannels) {
+		for (BooleanWriteChannel channel : this.digitalOutputChannels) {
 			String valueText;
 			var valueOpt = channel.value().asOptional();
 			if (valueOpt.isPresent()) {
-				valueText = valueOpt.get() ? "x" : "-";
+				valueText = valueOpt.get() ? "ON" : "OFF";
 			} else {
-				valueText = "?";
+				valueText = "Unknown";
```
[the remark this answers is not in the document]
Sn0w3y: Because we have it on the other Shelly Devices aswell personally i would also not make "ON" and "OFF" but x or -as we have it on KM-Tronic Relays but in this case:

Changed because of keeping up with the other Shelly Devices (Shelly 3EM and Sehlly Plug for example) See here:

https://github.com/OpenEMS/openems/blob/f885ae9368134142d7d899cb2e3461eb58ada4e8/io.openems.edge.io.shelly/src/io/openems/edge/io/shelly/shellyplug/IoShellyPlugImpl.java#L107C1-L108C1

and:

https://github.com/OpenEMS/openems/blob/f885ae9368134142d7d899cb2e3461eb58ada4e8/io.openems.edge.io.shelly/src/io/openems/edge/io/shelly/shelly3em/IoShelly3EmImpl.java#L118

io.openems.edge.io.shelly/src/io/openems/edge/io/shelly/shelly25/IoShelly25Impl.java
```diff
@@ -100,68 +113,84 @@ public void handleEvent(Event event) {
...
+		this._setSlaveCommunicationFailed(result == null);
+		Boolean relay1 = null;
+		Boolean overtemp1 = null;
+		Boolean overpower1 = null;
+		Boolean relay2 = null;
+		Boolean overtemp2 = null;
+		Boolean overpower2 = null;
 
+		try {
+			JsonElement jsonElement = JsonUtils.getAsJsonElement(result);
```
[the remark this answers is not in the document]
Sn0w3y: This is already done here:

https://github.com/OpenEMS/openems/blob/387ec874b3f2b3be1578a642835c6ee611240c3e/io.openems.edge.io.shelly/src/io/openems/edge/io/shelly/shelly25/IoShelly25Impl.java

sfeilmeier:
@Sn0w3y: Why did you close this (and other) Pull-Requests?

There is never a reason to close a Pull-Request. If it's not finished, just mark it as "Work in progress". If the code changed a lot, you can force-push your changes. As Pull-Requests can never be deleted on Github, opening new ones and closing them again all the time adds a lot of clutter...

Sn0w3y:
> There is never a reason to close a Pull-Request. If it's not finished, just mark it as "Work in progress". If the code changed a lot, you can force-push your changes. As Pull-Requests can never be deleted on Github, opening new ones and closing them again all the time adds a lot of clutter...

Thanks for pointing out the best way to handle Pull Requests (PRs). You're right about marking them as "Work in Progress" and updating with force-pushes to avoid clutter. I closed the PR to try and keep things neat because I thought too many commits in one PR could be confusing.

But I get what you're saying about always keeping PRs open for better tracking and less clutter from opening and closing new ones. I will do my best to not close them in the future. Also i will try to not Commit 100 times in one PR.

Thanks for the feedback, and I'll keep this in mind for next time!

sfeilmeier:
>  I closed the PR to try and keep things neat because I thought too many commits in one PR could be confusing.

This is not an issue for me. When I review a PR, I only look at PR title, its description and the "Files changed" tab - not at the individual commits.

Sn0w3y:
> This is not an issue for me. When I review a PR, I only look at PR title, its description and the "Files changed" tab - not at the individual commits.

Okay, i am very sorry for this inconvenience..

Can we still work on the new request? Would be glad if we can do so

PR #2524, opened 2024-02-06
Correct Energy Calculation on Shelly Plug S

Sn0w3y:
Now it is using:

Execute on Cycle Event "Execute Write". instead of setting it from the device


io.openems.edge.io.shelly/src/io/openems/edge/io/shelly/shellyplug/IoShellyPlugImpl.java
```diff
@@ -140,19 +151,34 @@ private void processHttpResult(JsonElement result, Throwable error) {
 			final var meters = JsonUtils.getAsJsonArray(result, "meters");
 			final var meter1 = JsonUtils.getAsJsonObject(meters.get(0));
 			final var power = Math.round(JsonUtils.getAsFloat(meter1, "power"));
-			final var energy = JsonUtils.getAsLong(meter1, "total") /* Unit: Wm */ / 60 /* Wh */;
```
sfeilmeier, line 143: Can you explain why it is better to calculate energy ourselves compared to using the Shelly native total value? I am using Shelly Plug S myself and did not recognize problems till now.

io.openems.edge.io.shelly/src/io/openems/edge/io/shelly/shellyplug/IoShellyPlugImpl.java
```diff
@@ -50,10 +59,12 @@ public class IoShellyPlugImpl extends AbstractOpenemsComponent
...
 
-	private BridgeHttp httpBridge;
 
-	@Reference
-	private BridgeHttpFactory httpBridgeFactory;
+	private volatile Timedata timedata;
+
+  @Reference
+  private BridgeHttpFactory httpBridgeFactory;
+  private BridgeHttp httpBridge;
```
sfeilmeier, line 67: This formatting change breaks the checkstyle CI build

sfeilmeier (changes requested):

io.openems.edge.io.shelly/src/io/openems/edge/io/shelly/shellyplug/IoShellyPlugImpl.java
Sn0w3y: I recall a post mentioning efforts to rewrite scripts in openEMS to avoid using device-reported energy values. This approach was considered because relying on native device metrics could result in inaccurately high energy values for a day, especially if the device had been operational before being integrated into openEMS.

Sn0w3y: Sure, Stefan, I've gone through the values already, and the differences are minimal on some days and on others they differ up to 0.7kWh. So not such a big difference in general :)
My main worry was indeed about the possibility of changing the device, leading to inaccuracies in these values. 
Could you clarify what you mean by adjusting them? :)

Feel free to close this PR if it's no longer necessary! :)

PR #2525, opened 2024-02-06
Update ControllerApiMqttImpl.java to reconnect on Connection Failure

Sn0w3y:
After the initial connection attempt in the activate method, we schedule the attemptConnect method to be called periodically every 60 seconds. This ensures that if the MQTT client becomes disconnected for any reason (e.g., network issues, broker restarts), the system will automatically attempt to re-establish the connection at regular intervals.

See Issue [here](https://github.com/OpenEMS/openems/issues/2510)

### DIFF
--- a/io.openems.edge.controller.api.mqtt/src/io/openems/edge/controller/api/mqtt/ControllerApiMqttImpl.java
+++ b/io.openems.edge.controller.api.mqtt/src/io/openems/edge/controller/api/mqtt/ControllerApiMqttImpl.java
@@ -52,17 +52,17 @@ public class ControllerApiMqttImpl extends AbstractOpenemsComponent
 		implements ControllerApiMqtt, Controller, OpenemsComponent, EventHandler {
 
 	protected static final String COMPONENT_NAME = "Controller.Api.MQTT";
-	private final ScheduledExecutorService scheduledExecutorService = Executors.newSingleThreadScheduledExecutor();
-	private volatile ScheduledFuture<?> reconnectFuture = null;
-
-	private final Logger log = LoggerFactory.getLogger(ControllerApiMqttImpl.class);
-	private final SendChannelValuesWorker sendChannelValuesWorker = new SendChannelValuesWorker(this);
-	private final MqttConnector mqttConnector = new MqttConnector();
-
-	private final AtomicInteger reconnectionAttempt = new AtomicInteger(0);
 	private static final long INITIAL_RECONNECT_DELAY_SECONDS = 5;
 	private static final long MAX_RECONNECT_DELAY_SECONDS = 300; // 5 minutes maximum delay.
 	private static final double RECONNECT_DELAY_MULTIPLIER = 1.5;
+
+	private final Logger log = LoggerFactory.getLogger(ControllerApiMqttImpl.class);
+	private final ScheduledExecutorService scheduledExecutorService = Executors.newSingleThreadScheduledExecutor();
+	private final SendChannelValuesWorker sendChannelValuesWorker = new SendChannelValuesWorker(this);
+	private final MqttConnector mqttConnector = new MqttConnector();
+	private final AtomicInteger reconnectionAttempt = new AtomicInteger(0);
+
+	private volatile ScheduledFuture<?> reconnectFuture = null;
 
 	@Reference(policy = ReferencePolicy.DYNAMIC, policyOption = ReferencePolicyOption.GREEDY, cardinality = ReferenceCardinality.OPTIONAL)
 	private volatile Timedata timedata = null;
@@ -199,7 +199,8 @@ public class ControllerApiMqttImpl extends AbstractOpenemsComponent
 			this.publish(ControllerApiMqtt.TOPIC_EDGE_CONFIG, config.toJson().toString(), //
 					1 /* QOS */, true /* retain */, new MqttProperties() /* no specific properties */);
 
-			// Trigger sending of all channel values, because a Component might have disappeared
+			// Trigger sending of all channel values, because a Component might have
+			// disappeared
 			this.sendChannelValuesWorker.sendValuesOfAllChannelsOnce();
 		}
 	}
--- a/io.openems.edge.controller.api.mqtt/src/io/openems/edge/controller/api/mqtt/ControllerApiMqttImpl.java
+++ b/io.openems.edge.controller.api.mqtt/src/io/openems/edge/controller/api/mqtt/ControllerApiMqttImpl.java
@@ -14,6 +14,7 @@ import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.ConfigurationPolicy;
 import org.osgi.service.component.annotations.Deactivate;
+import org.osgi.service.component.annotations.Modified;
 import org.osgi.service.component.annotations.Reference;
 import org.osgi.service.component.annotations.ReferenceCardinality;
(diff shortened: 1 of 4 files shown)


sfeilmeier:
It's surprising to me, that we have to go such a "hacky" approach. `MqttConnector` already sets the `setAutomaticReconnect` option, so maybe we are using something else in a wrong way. I found this, but could not find a quick solution: https://github.com/spring-projects/spring-integration/issues/3822

If we have to go that way, I would make `private ScheduledExecutorService scheduledExecutorService;` final. There is no benefit in creating it only during `activate()` and you can avoid the null-checks.

Sn0w3y:
> If we have to go that way, I would make `private ScheduledExecutorService scheduledExecutorService;` final. There is no benefit in creating it only during `activate()` and you can avoid the null-checks.

I understand where you're coming from, @sfeilmeier, but I'd like to offer a different perspective:

By initializing the ScheduledExecutorService in the activate() method, we gain the flexibility to adjust to configuration changes dynamically. This approach allows for finer control over how resources are allocated, which is essential given the inherently dynamic nature of OSGi services.
Opting to instantiate the ScheduledExecutorService only when the component is actually activated helps us use system resources more judiciously. This ensures that threads are allocated only when there's a genuine need for them, which is a more efficient use of resources.
Handling the ScheduledExecutorService lifecycle within the activate() and deactivate() methods enhances our ability to manage errors and recover from them. This strategy is closely integrated with the component's lifecycle, allowing for more context-aware error handling and potentially smoother recovery processes.
Adhering to the OSGi guidelines for resource management—specifically, the practice of initializing and disposing of resources like the ScheduledExecutorService in sync with the component's activation and deactivation—leads to cleaner and more maintainable code. It ensures that components behave in a predictable manner within the OSGi framework, which is beneficial for long-term maintenance and reliability.
To sum up, while setting the ScheduledExecutorService as final and initializing it outside the activate() method might seem to streamline the code by eliminating the need for null checks and minimizing the risk of encountering uninitialized fields, this approach overlooks the benefits of flexibility, efficient resource management, and lifecycle alignment with OSGi principles. These aspects are indispensable for crafting robust, efficient, and easily maintainable OSGi components.

sfeilmeier:
Thank you for the detailed explanation.

> By initializing the ScheduledExecutorService in the activate() method, we gain the flexibility to adjust to configuration changes dynamically

This would only be true if we implemented a `@Modified()` method, right? Otherwise there is technically no difference between constructor and @Activate, apart of the downside of not having the executor final. I might be too picky sometimes, but I have just seen too many NullPointerExceptions that were only possible because of such programming decisions. In my opinion and experience I would generally favour code quality over performance optimizations - with exceptions only for really performance critical operations (like our optimized websocket handling in the Backend)

Maybe we can agree on the approach that we took for "DebugCycleExecutor", combining constructor and activate() method: https://github.com/OpenEMS/openems/blob/develop/io.openems.backend.core/src/io/openems/backend/core/debugcycle/DebugCycleExecutor.java#L48-L53

For terminating, you can use the existing ThreadPoolUtils method: `ThreadPoolUtils.shutdownAndAwaitTermination();` (https://github.com/OpenEMS/openems/blob/develop/io.openems.backend.core/src/io/openems/backend/core/debugcycle/DebugCycleExecutor.java#L60)

Sn0w3y:
@sfeilmeier should be applied now ?

Dynamic Reconnection Interval: Added a new configuration parameter (reconnectionAttemptInterval) to the Config interface. This parameter allows users to specify the interval between reconnection attempts to the MQTT broker, enhancing the component's adaptability to network conditions or broker availability.

Enhanced Lifecycle Management: Modified the component's initialization and shutdown processes. The ScheduledExecutorService, responsible for managing reconnection attempts, is now initialized in the component's constructor and configured in the @Activate method. This ensures that the executor is ready for use immediately and adapts to configuration updates.

Graceful Shutdown: Implemented a more graceful shutdown process using ThreadPoolUtils.shutdownAndAwaitTermination in the @Deactivate method. This approach ensures that ongoing reconnection attempts are properly completed before the component is deactivated, preventing potential resource leaks.

Configuration Update Handling: Introduced a @Modified method to dynamically handle configuration changes. This method allows the component to adjust its behavior, such as changing the reconnection attempt interval, without needing to be fully deactivated and reactivated.

io.openems.edge.controller.api.mqtt/src/io/openems/edge/controller/api/mqtt/ControllerApiMqttImpl.java
```diff
@@ -67,35 +73,72 @@ public ControllerApiMqttImpl() {
...
+		this.config = config;
+		this.topicPrefix = String.format(ControllerApiMqtt.TOPIC_PREFIX, config.clientId());
+
+		this.logInfo(this.log, "Configuration modified. Adjusting MQTT connection settings.");
+
+		scheduleReconnect();
+	}
+
+	private void scheduleReconnect() {
+		if (this.scheduledExecutorService != null && !this.scheduledExecutorService.isShutdown()) {
```
sfeilmeier, line 99: It's usually preferable to not implement `modified()` because it requires complicated logic like this. If you choose the adaptive approach described above, maybe `activate()` is sufficient?

Restarting the thread is only required if the `reconnectionAttemptInterval` changes. You could validate this here, instead of always restarting the executor.

Also you should be able to use `ThreadPoolUtils.shutdownAndAwaitTermination` here as well, right?

io.openems.edge.controller.api.mqtt/src/io/openems/edge/controller/api/mqtt/ControllerApiMqttImpl.java
```diff
@@ -67,35 +73,72 @@ public ControllerApiMqttImpl() {
...
+		this.scheduledExecutorService.scheduleAtFixedRate(this::attemptConnect, 0,
+				this.config.reconnectionAttemptInterval(), TimeUnit.SECONDS);
+	}
+
+	private final Object connectLock = new Object();
+
+	private void attemptConnect() {
+		synchronized (connectLock) {
+			try {
+				if (mqttClient == null || !mqttClient.isConnected()) {
```
sfeilmeier, line 112: It's usually cleaner to reduce indentation in the code. Better:

```
if (mqttClient != null && mqttClient.isConnected()) {
   return;
}
````

File: io.openems.edge.controller.api.mqtt/src/io/openems/edge/controller/api/mqtt/ControllerApiMqttImpl.java [hunk 3]
```diff
@@ -67,35 +73,72 @@ public ControllerApiMqttImpl() {
...
 	}
 
 	@Override
 	@Deactivate
 	protected void deactivate() {
 		super.deactivate();
-		this.mqttConnector.deactivate();
-		this.sendChannelValuesWorker.deactivate();
+		ThreadPoolUtils.shutdownAndAwaitTermination(this.scheduledExecutorService, 10); // Graceful shutdown with
+																						// timeout
```
sfeilmeier, line 135: formatting is wrong

sfeilmeier (changes requested):

Sn0w3y:
@sfeilmeier maybe now everything is fixed i guess - if i forgot something just hit me up :)

Sn0w3y:
@sfeilmeier maybe now everything is fixed i guess - if i forgot something just hit me up :)

Sn0w3y:
Some news on this topic?

Sn0w3y:
up

PR #2531, opened 2024-02-08
UI: Install/Update Component filter to lower case

DerStoecki:
## Issue

When entering CamelCased / Uppercase entries in the search bar of component install/update -> the components won't show/cannot be found since the filter is CaseSensitive, while the filteredComponents are not.

see :
https://github.com/OpenEMS/openems/blob/2c938aab7a19407e6ab5dcac7b33c5e15da971df/ui/src/app/edge/settings/component/update/index.component.ts#L49-L55

## Fix

set filter to lowerCase

Co-authored-by:  Ulrich Laupp  <38915524+laupp@users.noreply.github.com>

### Before

Before

### After

After


sfeilmeier (approved):
Perfect. Thank you!

PR #2532, opened 2024-02-08
UI: AppCenter: RedeemKey -> Update of the SelectField

DerStoecki:
### Issue

In the AppCenter -> redeemKey option
On selecting a different key -> the key doesn't get visually updated. This is a bit irritating for the user.

RedeemKey

### Fix

formly-select-field.extended -> Apply ChangeDetectorRef -> detectChanges() after setValue()

Note:

I am not sure if ChangeDetectorRef in this context is the best solution. But it worked.

### DIFF
--- a/ui/src/app/shared/formly/formly-select-field.extended.ts
+++ b/ui/src/app/shared/formly/formly-select-field.extended.ts
@@ -1,4 +1,4 @@
-import { Component } from "@angular/core";
+import { ChangeDetectorRef, Component } from "@angular/core";
 import { ModalController } from "@ionic/angular";
 import { FieldWrapper } from "@ngx-formly/core";
 import { FormlySelectFieldModalComponent } from "./formly-select-field-modal.component";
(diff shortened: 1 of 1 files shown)


lukasrgr (approved):

lukasrgr:
@michaelgrill i sae nothing breaking in here, do you agree?

michaelgrill:
I m a little bit irritated what exactly is not working. 
Is it when someone has multiple keys registered and chooses a different one and then the select box wont get update?
In my case this was working fine. 

Btw it should not be important but i tested it with Chrome Browser.

DerStoecki:
> I m a little bit irritated what exactly is not working. Is it when someone has multiple keys registered and chooses a different one and then the select box wont get update? In my case this was working fine.
> 
> Btw it should not be important but i tested it with Chrome Browser.

yes thats the issue, someone has multiple registered keys and chooses a different one.
but it is odd indeed.
Especially bc a colleague of mine, who found the issue, uses chrome too.
I am using Firefox however.

DerStoecki:
Converting to Draft, until I synched and tested again

PR #2538, opened 2024-02-13
feat: add japanese language support to ui

miettal:
To be honest, it's not a perfect translation.
This translation include a lot machine translation(google-translate), however we checked one by one and fixed clearly strange japanese. So all japanese translation is not fluent japanese, but it is enough to unerstand by native japanese speakers.

We can improve japanese translation regularly, and follow upstream changes. Please consider to add japanese translation. Thank you.


sfeilmeier (approved):
Thank you!

PR #2519, opened 2024-02-05
Implement Shelly 3EM

Sn0w3y:
Link to Shelly [Product](https://www.shelly.com/de/products/shop/shelly-3em-1)

This is a 3-phase Meter, which can be used as a Grid Meter for example.

PR #2539, opened 2024-02-13
HTTP-Bridge add support for time based subscribes

michaelgrill:
- "Fixed" bug where an exception gets throw during deactivation of a component using the HTTP-Bridge (ty @Sn0w3y https://github.com/OpenEMS/openems/pull/2381#issuecomment-1928794069)

now to get an instance of a `BridgeHttp` the following code needs to be added

```
@Reference
private BridgeHttpFactory httpBridgeFactory;
private BridgeHttp httpBridge;

@Activate
private void activate() {
    this.httpBridge = this.httpBridgeFactory.get();
}

@Deactivate
private void deactivate() {
    this.httpBridgeFactory.unget(this.httpBridge);
    this.httpBridge = null;
}
```
as far as i know the exception which is happening is not intended to happen but for now to avoid unexpected behaviour because of the exception this is the prefered way to use it.

- Made CycleSubscriber thread safe to avoid concurrent modification exceptions

- Added time based subscribtions
  simple example to add a subscription with a delay of 1 minute between the calls would be
  
```
 final var delayProvider = DelayTimeProviderChain.fixedDelay(new Delay(1L, TimeUnit.MINUTES);
 this.httpBridge.subscribeTime(delayProvider, "http://127.0.0.1/status", t -> {
 	// process data
 }, t -> {
 	// handle error
 });
```


Sn0w3y:
> Added time based subscribtions
simple example to add a subscription with a delay of 1 minute between the calls would be

But does it make sense in terms of openEMS? 

As far as i know everything should be Cycle-Based right? 

So a delay is probably not really the best way to handle it - or do i get something wrong here? :)

Greets

Sn0w3y:
@michaelgrill could you check if my fix for my Pull is now using it correctly?

[[PullRequest](url)](https://github.com/OpenEMS/openems/pull/2519/files#diff-3d5ae9c7ded7d622bee13862ce0399cbb1d8f040a1fae7f863b6a9e6dddb53f1)

io.openems.edge.bridge.http/src/io/openems/edge/bridge/http/UrlFetcherImpl.java
```diff
@@ -1,61 +1,58 @@
...
-				// Check valid for all?
-				if (status < 300) {
-					future.complete(body);
-				} else {
-					throw new OpenemsException(
-							"Error while reading Endpoint " + urlString + ". Response code: " + status + ". " + body);
-				}
-			} catch (OpenemsNamedException | IOException e) {
-				future.completeExceptionally(e);
+	public String fetchEndpoint(final Endpoint endpoint) throws Exception {
```
da-Kai, line 18: "throws Exception" is not really good practice.

io.openems.edge.bridge.http/src/io/openems/edge/bridge/http/api/BridgeHttp.java
```diff
@@ -1,28 +1,37 @@
 package io.openems.edge.bridge.http.api;
 
+import static java.util.Collections.emptyMap;
```
da-Kai, line 3: please dont

> So when should you use static import? Very sparingly! Only use it when you'd otherwise be tempted to declare local copies of constants, or to abuse inheritance (the Constant Interface Antipattern).

https://docs.oracle.com/javase/8/docs/technotes/guides/language/static-import.html

io.openems.edge.bridge.http/src/io/openems/edge/bridge/http/UrlFetcher.java
```diff
@@ -1,24 +1,17 @@
...
 	 * Creates a {@link Runnable} to execute a request with the given parameters.
 	 * 
-	 * @param urlString      the url to fetch
-	 * @param connectTimeout the connection timeout
-	 * @param readTimeout    the read timeout
-	 * @param future         the {@link CompletableFuture} to fulfill after the
-	 *                       fetch
+	 * @param endpoint the {@link Endpoint} to fetch
+	 * 
 	 * @return the {@link Runnable} to run to execute the fetch
```
da-Kai, line 12: update javadoc return

io.openems.edge.bridge.http/test/io/openems/edge/bridge/http/api/BridgeHttpTest.java
```diff
@@ -0,0 +1,178 @@
...
+import io.openems.edge.bridge.http.BridgeHttpImpl;
+import io.openems.edge.bridge.http.CycleSubscriber;
+import io.openems.edge.bridge.http.DummyUrlFetcher;
+import io.openems.edge.bridge.http.api.BridgeHttp.Endpoint;
+
+public class BridgeHttpTest {
+
+	private DummyUrlFetcher fetcher;
+	private CycleSubscriber cycleSubscriber;
+	private BridgeHttp bridgeHttp;
```
da-Kai, line 27: BridgeHttpImpl? To save you some cast expressions

io.openems.edge.bridge.http/src/io/openems/edge/bridge/http/api/BridgeHttp.java
Sn0w3y: what about usidn Collections.emptyMap() with import java.util.Collections; ?

michaelgrill: I understand, but i would also consider `emptyMap` as well as `emptyList` are still readable enough without the class name, because the method name says everything of what to expect. A static import of `of` i would understand it could be from `Map.of` or `List.of` and it would affect the readability.
And also especially these two methods should be known (imo) because they decrease the need of null checks and unexpected NullPointerExceptions.
(Its also more to write :) )

But anyway if you still think it enhances readability i will add it :)

io.openems.edge.bridge.http/src/io/openems/edge/bridge/http/UrlFetcherImpl.java
```diff
@@ -3,59 +3,63 @@
...
+				// Read HTTP response
+				var content = new StringBuilder();
+				String line;
+				while ((line = in.readLine()) != null) {
+					content.append(line);
+					content.append(System.lineSeparator());
 				}
-			} catch (OpenemsNamedException | IOException e) {
-				future.completeExceptionally(e);
+				body = content.toString();
```
da-Kai, line 49: Vorschlag:
```Java
body = in.lines().collect(Collectors.joining(System.lineSeparator()));
```

io.openems.edge.bridge.http/src/io/openems/edge/bridge/http/time/Delay.java
da-Kai, line 1: is there a reason, you didn't just use [`java.time.Duration`](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html) instead of declaring a custom class ?

michaelgrill: yes because i then easily can use it in the schedule method of the ThreadPool
`this.pool.schedule(Runnable, Delay.amount(), delay.unit())`
and a Duration when trying to get milliseconds and seconds i would have to duration.getSeconds * 1000 + duration.getNanos / 1000

da-Kai: 1. https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#toMillis--
2. `this.pool.schedule(Runnable, duration.toMillis(), TimeUnit.MILLISECONDS)`

io.openems.edge.bridge.http/src/io/openems/edge/bridge/http/time/DelayTimeProviderChain.java
```diff
@@ -13,75 +14,73 @@
...
-	 * Creates a {@link DelayTimeProviderChain} which returns a the
-	 * {@link Delay#zero()} on request.
+	 * Creates a {@link DelayTimeProviderChain} which returns a the {@link Duration}
+	 * of zero on request.
 	 * 
 	 * @return a {@link DelayTimeProviderChain} of zero delay
 	 */
 	public static DelayTimeProviderChain immediate() {
-		return new DelayTimeProviderChain(() -> Delay.zero());
+		return new DelayTimeProviderChain(() -> Duration.ofSeconds(0));
```
da-Kai, line 23: There is also a `Duration.ZERO`, but okay

da-Kai (approved):

sfeilmeier (approved):
Thank you both for the professional code & review! 👍

io.openems.edge.bridge.http/src/io/openems/edge/bridge/http/api/HttpMethod.java
```diff
@@ -0,0 +1,24 @@
...
+	POST(true), //
+	PUT(true), //
+	DELETE(false), //
+	PATCH(true), //
+	HEAD(false), //
+	OPTIONS(false), //
+	TRACE(false) //
+	;
+
+	private final boolean bodyAllowed;
```
sfeilmeier, line 14: @michaelgrill I would just make this `public final boolean isBodyAllowed;` - and avoid the getter.

PR #2541, opened 2024-02-14
feature: implement the Entes MPR-15S-22 meter as modbus device

samuelBloch:
This pull request implements the Entes MPR-15S-22 network analyzer. The meter communicates over a serial modbus connection.


Sn0w3y:
Just a little thought from me - would't it be better to name it as follows:

io.openems.edge.meter.entes

and underneath the meters as packages to keep up if a new "Entes" meter is implemented? see here aswell:

https://community.openems.io/t/neues-meter-sdm-120/2385/7?u=sn0w3y

samuelBloch:
Thanks for the input, I renamed the bundle. Could you please check if this is what you meant?

PR #2544, opened 2024-02-15
Update InitializeEdgesWorker.java to refresh Edges every 30 minutes

Sn0w3y:
Right now if a new Edge Device is added to the Backend we always need to restart the Backend. This fix handles this as it calls the InitializeEdgesWorker every 30 Minutes without setting the Edges Offline again after first run.


sfeilmeier (changes requested):
Please check if all changes are actually required: https://github.com/OpenEMS/openems/pull/2544/files

All comments must be in English language.

I am not 100 % convinced that this approach is a good idea... but let's discuss it on a "clean as possible" PR.

Sn0w3y:
@sfeilmeier i guess it looks better now :)

sfeilmeier:
@michaelgrill: Can you please add your opinion?

PR #2546, opened 2024-02-17
UI: fixed some translations

Sn0w3y:


sfeilmeier (approved):

ui/src/app/user/user.component.html
```diff
@@ -23,8 +23,8 @@
               </ion-label>
             </ion-item>
             <ion-item lines="inset">
-              <ion-label>
-                Login
+              <ion-label translate>
+                Login.title
```
sfeilmeier, line 27: Note: best practice for translation keys is, to write them in UPPER_CASE. We are planning to continously migrate all existing ones to this format.

See example: https://github.com/ngx-translate/core?tab=readme-ov-file#4-define-the-translations

PR #2553, opened 2024-02-23
add optional config parameter to specify an optional MQTT topic prefix

samuelBloch:
We need an prefix to the default MQTT topic, so an optional config parameter has been added to achive this. If the optional prefix is set, OpenEMS Edge data is published to the MQTT broker under this topic: <optional_prefix>/edge/<edge_id>/...


sfeilmeier (approved):
I hope it's ok, that I silently updated the PR:

- rename new setting to `topicPrefix` with default value
- refactor to unit-testable `createTopicPrefix` method
- add JUnit test

PR #2569, opened 2024-03-11
Modbus bridge skip cycle function

cvabc:
OpenEMS's core cycle time is usually 1 second, but some modbus devices can be very slow. It's possible to lower the core cycle time but the whole system is slowed down.

The function `setSkipCycles` can be applied to each modbus tasks, for example:
```
...
new FC16WriteRegistersTask(40149, //
    m(EssSmaSunnyIsland.ChannelId.SET_ACTIVE_POWER, new SignedDoublewordElement(40149)), //
    m(EssSmaSunnyIsland.ChannelId.SET_CONTROL_MODE, new UnsignedDoublewordElement(40151)), //
    m(EssSmaSunnyIsland.ChannelId.SET_REACTIVE_POWER, new SignedDoublewordElement(40153))).setSkipCycles(1),
new FC16WriteRegistersTask(43090, //
    m(EssSmaSunnyIsland.ChannelId.GRID_GUARD_CODE, new UnsignedDoublewordElement(43090))).setSkipCycles(2), //
...
```
The `FC16WriteRegistersTask` task at register address 40149 is set to skip 1 cycle, which means every 2 cycles the register writing task is performed, while the `FC16WriteRegistersTask` task at register address 43090 is performed every 3 cycles.

`setSkipCycles` can also be applied to protocol level, for example:
```
@Override
protected ModbusProtocol defineModbusProtocol() throws OpenemsException {
    return new ModbusProtocol(...).setSkipCycles(2);
}
```
protocol level skip cycle setting will override the settings in each task.

### DIFF
--- a/io.openems.edge.bridge.modbus/readme.adoc
+++ b/io.openems.edge.bridge.modbus/readme.adoc
@@ -28,6 +28,32 @@ Read-Tasks can have two different priorities, that are defined in the ModbusProt
 - `LOW`: only one task of all defined LOW priority tasks of all components registered on the same bridge is executed per Cycle
 Write-Tasks always have `HIGH` priority, i.e. a set-point is always executed as-soon-as-possible - as long as the Component is not marked as defective
 
+=== Skip cycles
+
+OpenEMS's core cycle time is usually 1 second, but some modbus devices can be very slow. It's possible to lower the core cycle time but the whole system is slowed down.
+
+The function `setSkipCycles` can be applied to each modbus tasks, for example:
+```
+...
+new FC16WriteRegistersTask(40149, //
+    m(EssSmaSunnyIsland.ChannelId.SET_ACTIVE_POWER, new SignedDoublewordElement(40149)), //
+    m(EssSmaSunnyIsland.ChannelId.SET_CONTROL_MODE, new UnsignedDoublewordElement(40151)), //
+    m(EssSmaSunnyIsland.ChannelId.SET_REACTIVE_POWER, new SignedDoublewordElement(40153))).setSkipCycles(1),
+new FC16WriteRegistersTask(43090, //
+    m(EssSmaSunnyIsland.ChannelId.GRID_GUARD_CODE, new UnsignedDoublewordElement(43090))).setSkipCycles(2), //
+...
+```
+The `FC16WriteRegistersTask` task at register address 40149 is set to skip 1 cycle, which means every 2 cycles the register writing task is performed, while the `FC16WriteRegistersTask` task at register address 43090 is performed every 3 cycles.
+
+`setSkipCycles` can also be applied to protocol level, for example:
+```
+@Override
+protected ModbusProtocol defineModbusProtocol() throws OpenemsException {
+    return new ModbusProtocol(...).setSkipCycles(2);
+}
+```
+protocol level skip cycle setting will override the settings in each task.
+
 === Channels
 
 Each Modbus Bridge provides Channels for more detailed information:
--- a/io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/api/ModbusProtocol.java
+++ b/io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/api/ModbusProtocol.java
@@ -78,4 +78,10 @@ public class ModbusProtocol {
 		this.taskManager.getTasks() //
 				.forEach(Task::deactivate);
 	}
+
+	public ModbusProtocol setSkipCycles(int cycles) {
+		this.taskManager.getTasks() //
+				.forEach(t -> t.setSkipCycles(cycles));
+		return this;
+	}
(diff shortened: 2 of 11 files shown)


sfeilmeier:
Thank you for the suggested improvement. Can you name a hardware that is affected by this problem, that reading is very slow? 

My doubt with your implementation is, that the cycle time is not static. We run productive systems with cycle times of < 500ms. In that case a hard-coded `SkipCycles` value might not be sufficient. Should it be configurable? Or should the Modbus-Bridge be smarter in automatically detecting such devices?

cvabc:
We will use the Connex BESS https://www.connexxsys.com/products/ess-hn/, which should be at least 2 seconds between queries.
I think it's hard for the modbus bridge to detect the real delay time of this device by checking the timestamp of modbus reply.
As for the configurable parameter, do you mean input time in seconds and calculate the cycles needed?

sfeilmeier:
> We will use the Connex BESS https://www.connexxsys.com/products/ess-hn/, which should be at least 2 seconds between queries. I think it's hard for the modbus bridge to detect the real delay time of this device by checking the timestamp of modbus reply. As for the configurable parameter, do you mean input time in seconds and calculate the cycles needed?

Where does the value of "2 seconds" come from? Is it a requirement in the documentation or did you experience this delay in the lab? Does it affect all queries (i.e. modbus read and write requests) or only modbus write-requests?

If it's the latter, in principle the Modbus Bridge should be smart enough to handle slow responses. You would only see "CycleTimeIsTooShort" warning (which you could hide in the Sum-component).

If it's only about writing to write-registers more seldomly, maybe it could be handled in the `setActivePower()`-method.

cvabc:
The "2 seconds" interval is the minimum time needed by discussing with their engineer. The receving buffer will overflow and command will be lost if the invertal is lower than 2 seconds. The interval requirement in the official document is even longer. 

Unfortunately I haven't experimented with cycle time lower than 2 seconds on this device and am not sure whether the modbus bridge can detect the problem. But if so, can we safely ignore the "CycleTimeIsTooShort" warning? Can the bridge extend queries to more than 1 cycle automatically and safely?

In our case, I think both reading and writing will need to be slowed down. For the writing part, do you mean the `applyPower()` function for this _device_? It seems `setActivePower()` is mainly used by _controllers_.

sfeilmeier:
> Unfortunately I haven't experimented with cycle time lower than 2 seconds on this device and am not sure whether the modbus bridge can detect the problem. But if so, can we safely ignore the "CycleTimeIsTooShort" warning? Can the bridge extend queries to more than 1 cycle automatically and safely?

If the modbus request to the ESS fails, the Modbus Bridge should safely handle the situation and try again later. If the error is only internal to the ESS, than this won't be sufficient.

> In our case, I think both reading and writing will need to be slowed down. For the writing part, do you mean the `applyPower()` function for this _device_? It seems `setActivePower()` is mainly used by _controllers_.

Sorry, I meant the [applyPower()](https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.ess.api/src/io/openems/edge/ess/api/ManagedSymmetricEss.java#L611-L625) method. If it were only about writing registers, you could have added some logic there to for example not repeatedly write the same value.

To be honest I do not fully unterstand if your implementation could cause any implications in the complex logic of the Modbus Bridge. 

If I had to implement such a feature I believe I would skip cycles already in the [handleEvent()](https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/api/AbstractModbusBridge.java#L126) method. And skipping logic, as suggested, should be in unit time (seconds or milliseconds) and not cycle in my opinion.

cvabc:
Thanks for the answer!

I will explain my understanding of your idea:
Instead of operating on the low level task queue, override the `handleEvent()` function in `BridgeModbus(Serial|Tcp)Impl.java` and add the skip cycle logic inside. The number of cycles to be skipped is calculated from the core cycle time and parameter `period` (ms), which is configured in `Config(Serial|Tcp).java`.

Do you have any suggestion on my understanding?

sfeilmeier:
Yes, that's exactly how I meant it. This does not mean your approach is wrong, it's just I did not think through all the side-effects it could have and the Modbus Bridge logic is already quite complex. 

What's your opinion on that? Do you think your approach is better and we should stick with it?

cvabc:
I will compare these 2 methods.
### This PR
#### Advantage
- Finer task level control.
#### Disadvantage
- Low level modifications that may affect other components.
- Coupling of skip parameter inside device drivers. Ideally, device drivers should not care about modbus parameters.

### Your idea
#### Advantage
- Simplicity and reliability.
- Parameter decoupling. Everything related to modbus communication is configured inside the modbus bridge.
#### Disadvantage
- Cannot set different parameters for different tasks.


Overall I would prefer reliability over functionality.

PR #2587, opened 2024-03-19
Janitza UMG 511: fix inverted active power

tsicking:
Previously, the overall active power was not inverted if configured, only the phase powers were. This is fixed here.


sfeilmeier (approved):
We do not have a Janitza UMG 511 with activated `invert` in production, but the change sounds feasible. Thank you!

PR #2589, opened 2024-03-22
Update setup-ide.adoc

NiallDarwin:
Corrected error in ng serve instruction


NiallDarwin:
I was in looking at the wrong part of the docs. PR removed.

PR #2594, opened 2024-03-29
Center Text in Footer

Sn0w3y:


sfeilmeier:
@lukasrgr: please check. Maybe I did a mistake while backporting?

lukasrgr:
@Sn0w3y what was your issue?

Sn0w3y:
@lukasrgr, no problem at all, just a minor change but, in my opinion, a more elegant way to display the values in the footer.

lukasrgr (approved):

Sn0w3y:
> @Sn0w3y i had to deny my approval, i realized our internal version and openEMS are drifting apart at one point

Actually i do not understand :o Means no centered Footer Data ?
Or could you then fix it by applying the same style than the internal Version ?

They are indeed drifting apart - also on the size of the Text in the Footer aswell

Greetings !

lukasrgr:
@Sn0w3y found it, apparently the new esLint configuration overwrote style applied in ts,

Apparently ```ng lint --fix``` is not always the best solution

https://github.com/OpenEMS/openems/blob/4b3f35b938c7de6fdc59acb5ba39465c196850ab/ui/src/app/shared/footer/footer.ts#L24

Could you be so kind and fix the syntax in the applied style?

Sn0w3y:
@lukasrgr shoul be applied now ?

Greetings !

lukasrgr (approved):

Sn0w3y:
@lukasrgr should we exclude this from nglint possibly ?

sfeilmeier (approved):

PR #2601, opened 2024-04-01
Docs: update implementing a device

Sn0w3y:
Added the Case to test only the Module on which you are working on.

./gradlew :io.openems.edge.ess.sungrow:clean for example - it leads to a faster building/testing Time

### DIFF
--- a/doc/modules/ROOT/pages/contribute/coding-guidelines.adoc
+++ b/doc/modules/ROOT/pages/contribute/coding-guidelines.adoc
@@ -34,6 +34,7 @@
 * Add comments to code that is difficult to understand
 * Prefer functional programming (e.g. use Lambda expressions)
 * Add JUnit tests.
+* Run ./gradlew :io.openems.edge.io.shelly:clean in the root Directory of openEMS to check the tests of your module.
 * Run ./tools/prepare-commit.sh script
 
 === Tool support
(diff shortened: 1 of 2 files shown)


doc/modules/ROOT/pages/contribute/coding-guidelines.adoc
```diff
@@ -34,6 +34,7 @@
 * Add comments to code that is difficult to understand
 * Prefer functional programming (e.g. use Lambda expressions)
 * Add JUnit tests.
+* Run ./gradlew :io.openems.edge.io.shelly:clean in the root Directory of openEMS to check the tests of your module.
```
sfeilmeier, line 37: The official spelling is **OpenEMS** - at least _we_ should stick to it.

"directory" should be lower-case here.

If fact I believe this note would better fit in https://github.com/OpenEMS/openems/blob/develop/doc/modules/ROOT/pages/edge/implement.adoc than in the general coding-guidelines.

sfeilmeier (changes requested):

doc/modules/ROOT/pages/contribute/coding-guidelines.adoc
Sn0w3y: Under which Section do you think it is better suited ?

sfeilmeier: Maybe just add it as a separate chapter `1.11`?!

sfeilmeier (approved):

PR #2602, opened 2024-04-01
Controller.Api.MQTT: reconnect on connection failure

Sn0w3y:
## Add Reconnection Logic to MQTT Controller

This PR introduces reconnection logic to the MQTT Controller, ensuring robustness in the face of connection losses with the MQTT broker. The implementation schedules reconnection attempts with an exponential backoff strategy, aiming to minimize the impact on both the broker and the controller while trying to re-establish a lost connection.

### Key Changes:

1. **Scheduled Reconnection Attempts**: Utilizes a single-threaded `ScheduledExecutorService` to manage reconnection attempts, allowing for delayed execution with exponential backoff.

2. **Adaptive Delay Calculation**: Implements an adaptive delay calculation using a base delay, a maximum delay cap, and a multiplier to manage the delay between reconnection attempts, preventing aggressive reconnection attempts.

3. **Integration with Component Lifecycle**: Hooks the reconnection logic into the component's activation process, ensuring attempts to connect or reconnect are made upon activation. Similarly, ensures proper shutdown of the executor service upon deactivation to clean up resources.

## Integration:
This implementation ensures that the MQTT Controller attempts to reconnect to the broker following a connection loss, employing an exponential backoff strategy to balance between prompt reconnection and avoiding overwhelming the broker or the network.


Sn0w3y:
Maybe this Feature would be a "nice to have" ?

See Issue #2510:

image

sfeilmeier (approved):

PR #2609, opened 2024-04-04
SunSpec: Channel values synchronized 

tsicking:
A small fix in the mapping between SunSpec channel values and OpenEMS channel values: Previously, this was done by an `onUpdate()`-listener, meaning that the OpenEMS channel was one cycle behind the SunSpec channel. This is fixed here by using an `onSetNextValue()`-listener instead.


clehne (approved):
Ok.

sfeilmeier (approved):

PR #2612, opened 2024-04-08
Generate maven poms and releases within bnd workspace.

splatch:
This pull request introduces several changes:
1) It tidy up bnd files across modules, setting version in one place (cnf/build.bnd). 2) The license headers and vendor declaration are also updated. 3) There are additional properties which enable deployment of released artifacts to maven repositories.

Few notes, as there is an ongoing concern with `Bundle-Version` policy taken in the project. At the moment all bundles use `1.0.0.${tstamp}` (`1.0.0.SNAPSHOT` after update) which is, even from OSGi point of view, a moving target.

Above is not in line with tags which use completely different format (`2024.4.0`). At the moment the release version is not propagated to gradle/bnd build making it difficult to navigate between both. For example release tag 2024.4.0 resulted in bundle version is 1.0.0.202403311530 due to different time offsets. It is possible to synchronize these through `Bundle-Version: ${env;VERSION;1.0.0}.SNAPSHOT` construction. Then, release scripts and pipelines needs to be updated in order to populate build environment.


sfeilmeier:
@da-Kai I'd appreciate if you could review this after your holiday.

See Community-Post for reference: https://community.openems.io/t/maven-artifact-repository/2482/7

splatch:
One side effect which I observed with my changes, the `Bundle-Version` for wrapper bundles includes `SNAPSHOT` in version. Probably have to tweak this to stick with static version of wrapped bundle.

splatch:
Can I get someone's interest to make initial review before another holiday season starts? :-)

Cheers,
Łukasz

.github/workflows/build.yml
```diff
@@ -65,6 +67,14 @@ jobs:
           mkdir -p ./jacoco
           echo $PR_NUMBER > ./jacoco/jacoco_report_number
           echo '![Code Coverage]'"(https://img.shields.io/badge/Code%20Coverage-${coverage}%25-${color}?style=flat)" > ./jacoco/jacoco_report_badge
+
+      - name: Deploy Edge+Backend assets
+        if: success() && github.event_name == 'push' && github.ref_name == 'develop'
+        run: ./gradlew buildEdge buildBackend release -DmavenPublishRepository=RemoteMavenRepository
+        env:
+          CI_DEPLOY_USERNAME: ${{ secrets.CI_DEPLOY_USERNAME }}
+          CI_DEPLOY_PASSWORD: ${{ secrets.CI_DEPLOY_PASSWORD }}
```
da-Kai, line 76: This build yml runs on every push to every pull-request. Is it sensible to deploy every time?

cnf/build.bnd
```diff
@@ -18,9 +18,23 @@
 		pretty = true; \
 		local = ${build}/release
 
--releaserepo: Release
 -baselinerepo: Release
 
+Bundle-Vendor: FENECON GmbH
```
da-Kai, line 23: I guess default `Bundle-Vendor` should be 'OpenEMS Association e.V.' what do you think @sfeilmeier?

ui/src/app/changelog/view/component/changelog.component.ts
da-Kai, line 1: these changes are probably not intended

tools/create-release.sh
da-Kai, line 1: your changes to the shell scripts don't seam to realy change everything.

da-Kai (commented):
Hi @splatch,
nice work 👍

I think the "1.0.0.SNAPSHOT" is just a placeholder, as this version string is not in use right now. So no need to preserve it.

It would be feasable to update it everytime the in code Version string is updated in this function.
https://github.com/OpenEMS/openems/blob/a946ba09d5bb7f491f15256a215f6fdb926526bd/tools/common.sh#L43C1-L63C2

PR #2615, opened 2024-04-11
Modbus TCP bridge skip cycle function

cvabc:
Related PR https://github.com/OpenEMS/openems/pull/2569

OpenEMS's core cycle time is usually 1 second, but some modbus devices can be very slow. It's possible to lower the core cycle time but the whole system is slowed down.

The new configuable paraemter `intervalBetweenAccesses` can be configured to slow down the modbus bridge's access. Specifically, it defines the minimum time (in milliseconds) between 2 modbus accesses, which normally equals to the core cycle time (usually 1000ms).

`intervalBetweenAccesses` is aligned to and calculated from the core cycle time. It's set to be 0 by default and any value between 0 and core cycle time will not affect the original modbus bridge logic. The right amount of core cycles to be skipped will be calculated and applied when `intervalBetweenAccesses` is set to be longer than core cycle time.

Note that the real interval may be longer than configured due to the round off of core cycle time. For example, we set `intervalBetweenAccesses` to 1.5 seconds but core time is 1 second and in this case the final interval is 2 seconds.


io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/BridgeModbusTcpImpl.java
```diff
@@ -72,6 +86,18 @@ private void modified(ComponentContext context, ConfigTcp config) throws Unknown
 	private void applyConfig(ConfigTcp config) {
 		this.setIpAddress(InetAddressUtils.parseOrNull(config.ip()));
 		this.port = config.port();
+        int coreCycleTime = 1000;
+        try {
+			// the default cycleTime is not persisted and won't be returned, only modified cycleTime will!
+			coreCycleTime = (int) (Integer) this.cm.getConfiguration("Core.Cycle").getProperties().get("cycleTime");
```
sfeilmeier, line 92: [Cycle](https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.common/src/io/openems/edge/common/cycle/Cycle.java#L120-L125) has a method `getCycleTime()`. Is there a reason you don't use it directly?

io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/ConfigTcp.java
```diff
@@ -33,5 +33,8 @@
 	@AttributeDefinition(name = "Invalidate elements after how many read Errors?", description = "Increase this value if modbus read errors happen frequently.")
 	int invalidateElementsAfterReadErrors() default 1;
 
+	@AttributeDefinition(name = "Interval between accesses (in milliseconds)", description = "Cycle skipping will be calculated from this value and core cycle time.")
+	int intervalBetweenAccesses() default 0;
```
sfeilmeier, line 37: I would rename this to `minIntervalBetweenAccesses()`.

Description for unit can be simply "[ms]" instead of "(in milliseconds)" - see this example: https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.core/src/io/openems/edge/core/cycle/Config.java#L13

sfeilmeier (commented):
I like the feature 👍

How about this alternative approach for implementation?
- Drop the calculation of `noSkipIdx` in `applyConfig()`
- Instead calculate the passed time manually in `handleEvent()`:

```
public void handleEvent(Event event) {
    [...]
    if (this.skipCycle(event)) {
        return;
    }
    [...]
}

private Instant lastBeforeProcessImageEvent = Instant.MIN;
private boolean _skipCycle;
private boolean skipCycle(Event event) {
	final var now = Instant.now();
    if (event.getTopic() == EdgeEventConstants.TOPIC_CYCLE_BEFORE_PROCESS_IMAGE) {
        if (Duration.between(this.lastBeforeProcessImageEvent, now).toMillis() > this.minIntervalBetweenAccesses) {
        	this._skipCycle = true;
        }
        this.lastBeforeProcessImageEvent = now;
    }
    return this._skipCycle;
}
```

io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/BridgeModbusTcpImpl.java
sfeilmeier: Ah sorry. Just realized, that you asked exactly this question here: https://community.openems.io/t/bridgemodbustcpimpl-doesnt-work-when-cycle-is-referenced/2530/3

io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/ConfigTcp.java
cvabc: Thanks! I will fix them.

cvabc:
> I like the feature 👍
> 
> How about this alternative approach for implementation?
> 
> * Drop the calculation of `noSkipIdx` in `applyConfig()`
> * Instead calculate the passed time manually in `handleEvent()`:
> 
> ```
> public void handleEvent(Event event) {
>     [...]
>     if (this.skipCycle(event)) {
>         return;
>     }
>     [...]
> }
> 
> private Instant lastBeforeProcessImageEvent = Instant.MIN;
> private boolean _skipCycle;
> private boolean skipCycle(Event event) {
> 	final var now = Instant.now();
>     if (event.getTopic() == EdgeEventConstants.TOPIC_CYCLE_BEFORE_PROCESS_IMAGE) {
>         if (Duration.between(this.lastBeforeProcessImageEvent, now).toMillis() > this.minIntervalBetweenAccesses) {
>         	this._skipCycle = true;
>         }
>         this.lastBeforeProcessImageEvent = now;
>     }
>     return this._skipCycle;
> }
> ```

Actually the same idea was used in the initial commit https://github.com/OpenEMS/openems/pull/2615/commits/940fa15b9c2942d652db1b9cd239af3e559311f1 but was later replaced by https://github.com/OpenEMS/openems/pull/2615/commits/4035fb225672ddaab2afc9315418ca5dc964dc65, for edge case reasons.
The test script was written with the concept of cycle, instead of passed time, as can be found in the use of `next()` function: https://github.com/OpenEMS/openems/blob/76f519cf550ad2a03f717dee11b70d76ade6d55c/io.openems.edge.bridge.modbus/test/io/openems/edge/bridge/modbus/BridgeModbusTcpImplTest.java#L77
The problem of using passed time shows when repeating the test script for dozens of time, little imprecision of cycle time will accumulate to large error (edge case), which could potentially invalidate the test. Therefore I tried another way, not rely on time passed, but time relative to core cycle and hence counting how many cycles should be skipped.
Actually current implementation would fail in tests in some conditions as well. For example, the following change of the parameter would invalidate the test
```
--- a/io.openems.edge.bridge.modbus/test/io/openems/edge/bridge/modbus/BridgeModbusTcpImplTest.java
+++ b/io.openems.edge.bridge.modbus/test/io/openems/edge/bridge/modbus/BridgeModbusTcpImplTest.java
@@ -177,7 +177,7 @@ public class BridgeModbusTcpImplTest {
                        }
                        // 0 <= interval < maxInterval
                        numTests = 7;
-                       int maxInterval = CYCLE_TIME * 3;
+                       int maxInterval = CYCLE_TIME * 4;
                        for (int i = 0; i < numTests; i++) {
                                var sut = new BridgeModbusTcpImpl();
                                var device = new MyModbusComponent(DEVICE_ID, sut, UNIT_ID);
```
But I could not explain the error and now it seems to me the error is not related to the implementation of skip cycle function, but the test script itself. Maybe I misunderstand the test script somehow?
Also I noticed the original test script would also fail around here: https://github.com/OpenEMS/openems/blob/76f519cf550ad2a03f717dee11b70d76ade6d55c/io.openems.edge.bridge.modbus/test/io/openems/edge/bridge/modbus/BridgeModbusTcpImplTest.java#L96 Is it expected?

PR #2623, opened 2024-04-28
FEMS Backports 2023-04

sfeilmeier:


ui/src/app/shared/footer/footer.ts
```diff
@@ -18,16 +19,16 @@ import { Role } from "../type/role";
...
       bottom: 0;
       width: 100%;
-      font-size: 14px !important;
-    }
 
-    ion-row {
-      text-align: center;
-    }
+      font-size: 14px !important;
+      ion-row {
```
Sn0w3y, line 24: Causes:

```
Warning: ▲ [WARNING] A nested style rule cannot start with "ion-row" because it looks like the start of a declaration [css-syntax-error]

    C:/Users/User/Desktop/openems-2024.2.0_ORIGINAL/ui/src/app/shared/footer/footer.ts-angular-inline--3.css:13:6:
      13 │       ion-row {
         │       ~~~~~~~
         ╵       :is(ion-row)

  To start a nested style rule with an identifier, you need to wrap the identifier in ":is(...)" to prevent the rule from being parsed as a declaration.
```

ui/src/app/shared/footer/footer.ts
```diff
@@ -18,16 +19,16 @@ import { Role } from "../type/role";
...
+        text-align: center;
+      }
 
-    ion-item {
```
Sn0w3y, line 28: Causes:

`Warning: ▲ [WARNING] A nested style rule cannot start with "ion-item" because it looks like the start of a declaration [css-syntax-error]

    C:/Users/User/Desktop/openems-2024.2.0_ORIGINAL/ui/src/app/shared/footer/footer.ts-angular-inline--3.css:16:6:
      16 │       ion-item {
         │       ~~~~~~~~
         ╵       :is(ion-item)

  To start a nested style rule with an identifier, you need to wrap the identifier in ":is(...)" to prevent the rule from being parsed as a declaration.`

PR #2628, opened 2024-05-01
Only a small Compilation Warning fix

Sn0w3y:
See Comments here:

https://github.com/OpenEMS/openems/pull/2623/files#r1586336975
https://github.com/OpenEMS/openems/pull/2623/files#r1586337561

### DIFF
--- a/ui/src/app/shared/footer/footer.ts
+++ b/ui/src/app/shared/footer/footer.ts
@@ -21,11 +21,11 @@ import { Role } from "../type/role";
       width: 100%;
 
       font-size: 14px !important;
-      ion-row {
+      :is(ion-row) {
         text-align: center;
       }
 
-      ion-item {
+      :is(ion-item) { /* Update the selector here */
         --min-height: initial !important;
         font-size: inherit;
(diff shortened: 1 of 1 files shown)


Sn0w3y:
Here is the Warning we get if we do not Apply the PR:

`Warning: ▲ [WARNING] A nested style rule cannot start with "ion-row" because it looks like the start of a declaration [css-syntax-error]

    C:/Users/User/Desktop/openems-2024.2.0_ORIGINAL/ui/src/app/shared/footer/footer.ts-angular-inline--3.css:13:6:
      13 │       ion-row {
         │       ~~~~~~~
         ╵       :is(ion-row)

  To start a nested style rule with an identifier, you need to wrap the identifier in ":is(...)" to prevent the rule from being parsed as a declaration.
`

Sn0w3y:
@lukasrgr maybe have a look

lukasrgr (approved):
thx for you contribution, this has actually been an issue for some time

sfeilmeier (approved):

PR #2637, opened 2024-05-11
feat: Add config to only run peak shaving controller on peaks

nlamarti:
This is done so that it can be combined with for instance the time of use controller

### DIFF
--- a/io.openems.edge.controller.symmetric.peakshaving/src/io/openems/edge/controller/symmetric/peakshaving/Config.java
+++ b/io.openems.edge.controller.symmetric.peakshaving/src/io/openems/edge/controller/symmetric/peakshaving/Config.java
@@ -25,9 +25,16 @@ import org.osgi.service.metatype.annotations.ObjectClassDefinition;
 
 	@AttributeDefinition(name = "Peak-Shaving power", description = "Grid purchase power above this value is considered a peak and shaved to this value.")
 	int peakShavingPower();
+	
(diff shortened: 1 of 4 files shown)


sfeilmeier:
Thanks for the PR. I've been thinking about a similar functionality, but would prefer a different approach:

- `enableRecharge = false` can already be achieved by setting `rechargePower` to a very small value (e.g. `0` or negative)
- instead of setting a fixed value with `ess.setActivePowerEquals()`, set a minimum (`ActivePowerGreaterOrEquals`) or maximum (`ActivePowerLessOrEquals`) respectively for peak-shaving and recharging.

This way Controller does not have to know if there is another Controller coming afterwards.

What do you think?

nlamarti:
> * `enableRecharge = false` can already be achieved by setting `rechargePower` to a very small value (e.g. `0` or negative)

yes this we can do together with the no longer setting of 0 when both peak-shaving and recharging is not needed


> * instead of setting a fixed value with `ess.setActivePowerEquals()`, set a minimum (`ActivePowerGreaterOrEquals`) or maximum (`ActivePowerLessOrEquals`) respectively for peak-shaving and recharging.

we can and i think should do this, but it will still not be a optional peakshaving if we keep the set to 0 if there is no peak or recharge moment, and if we remove that part it will stay at the last given value if the controller is running stand alone (so no other controller running after it)

PR #2662, opened 2024-06-04
Fix docker ui builds

da-Kai:
Current openems-ui Docker images do not work because the env.js file is missing.


sfeilmeier (approved):
Good solution with `patch`! 👍

PR #2684, opened 2024-06-26
Feature/reverse power relay

DerWahreKlinki:
Adds a reverse power relay (german: Rundsteuerempfänger).

Tested with Siemens LOGO! 8 as network -> modbus interface

### DIFF
--- a/io.openems.edge.controller.pvinverter.reversepowerrelay/src/io/openems/edge/controller/pvinverter/reversepowerrelay/ReversePowerRelayImpl.java
+++ b/io.openems.edge.controller.pvinverter.reversepowerrelay/src/io/openems/edge/controller/pvinverter/reversepowerrelay/ReversePowerRelayImpl.java
@@ -1,0 +1,156 @@
+package io.openems.edge.controller.pvinverter.reversepowerrelay;
+
+import org.osgi.service.component.ComponentContext;
+import org.osgi.service.component.annotations.Activate;
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.ConfigurationPolicy;
+import org.osgi.service.component.annotations.Deactivate;
+import org.osgi.service.component.annotations.Reference;
+import org.osgi.service.metatype.annotations.Designate;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.openems.common.exceptions.OpenemsError.OpenemsNamedException;
+import io.openems.common.types.ChannelAddress;
+import io.openems.edge.common.component.AbstractOpenemsComponent;
+import io.openems.edge.common.component.ComponentManager;
+import io.openems.edge.common.component.OpenemsComponent;
+import io.openems.edge.controller.api.Controller;
+import io.openems.edge.pvinverter.api.ManagedSymmetricPvInverter;
+import io.openems.edge.common.channel.BooleanReadChannel;
+
+@Designate(ocd = Config.class, factory = true)
+@Component(//
+		name = "Controller.PvInverter.ReversePowerRelay", //
+		immediate = true, //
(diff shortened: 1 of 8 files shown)


io.openems.edge.controller.pvinverter.reversepowerrelay/src/io/openems/edge/controller/pvinverter/reversepowerrelay/ReversePowerRelayImpl.java
```diff
@@ -0,0 +1,169 @@
...
+			BooleanReadChannel channel = this.componentManager.getChannel(address);
+			return channel.value().orElse(null);
+		} catch (OpenemsNamedException e) {
+			this.log.error("Error reading channel value", e);
+			return null;
+		}
+	}
+
+	@Override
+	public void run() throws OpenemsNamedException {
```
Sn0w3y, line 122: Could be simplyfied with:

```java
private void setPvLimitBasedOnChannelValues(Boolean value0Percent, Boolean value30Percent, Boolean value60Percent, Boolean value100Percent) {
    if (value0Percent) {
        setPvLimit(0);
    } else if (value30Percent && !value60Percent && !value100Percent) {
        setPvLimit(this.powerLimit30Percent);
    } else if (value60Percent && !value30Percent && !value100Percent) {
        setPvLimit(this.powerLimit60Percent);
    } else if (value100Percent) {
        setPvLimit(null);
    } else {
        setPvLimit(0);
    }
} 
```

io.openems.edge.controller.pvinverter.reversepowerrelay/src/io/openems/edge/controller/pvinverter/reversepowerrelay/ReversePowerRelayImpl.java
```diff
@@ -0,0 +1,169 @@
...
+				ReversePowerRelay.ChannelId.values() //
+		);
+	}
+
+	@Activate
+	private void activate(ComponentContext context, Config config) {
+		super.activate(context, config.id(), config.alias(), config.enabled());
+		this.pvInverterId = config.pvInverter_id();
+
+		this.powerLimit30Percent = config.powerLimit30();
```
Sn0w3y, line 61: Couldn't this be calculated based on the Inverters Power?

Like maybe add one Config field to add the max Power of the Inverter and based on this Limit it can calculate how much is 30% or 60% off of it ? :)

io.openems.edge.controller.pvinverter.reversepowerrelay/src/io/openems/edge/controller/pvinverter/reversepowerrelay/ReversePowerRelayImpl.java
DerWahreKlinki: I think it´s important to handle not (yet) exististing modbus channels. Otherwise limitation kicks in at every restart. The Exception in run() and the early return of the null-checks offers more safety. What do you think?

io.openems.edge.controller.pvinverter.reversepowerrelay/src/io/openems/edge/controller/pvinverter/reversepowerrelay/ReversePowerRelayImpl.java
DerWahreKlinki: Thought about that, too. My intention was to have a little more "freedom" how much the inverter gets limited. But you´re right. It´s a bit more handy to calculate the values.

DerWahreKlinki:
- Made max power of PV inverter configurable (not all inverters have set this value) and the 30/60% are calculated dynamically
- Improved handling of empty values
- Improved JuniTests

PR #2699, opened 2024-07-02
fix: add types to edgeconfig.ts

miettal:
npx tsc --strict: 3646 -> 3622 errors

### DIFF
--- a/ui/tsconfig.json
+++ b/ui/tsconfig.json
@@ -20,7 +20,7 @@
     "module": "es2020",
     "noImplicitOverride": true,
     "lib": [
-      "es2019",
+      "es2022",
       "dom"
     ],
     "resolveJsonModule": true,
--- a/ui/src/app/shared/components/edge/edgeconfig.ts
+++ b/ui/src/app/shared/components/edge/edgeconfig.ts
@@ -1,4 +1,3 @@
-// @ts-strict-ignore
 import { ChannelAddress, Widgets } from '../../shared';
 import { Edge } from './edge';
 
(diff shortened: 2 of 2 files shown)


ui/tsconfig.json
```diff
@@ -20,7 +20,7 @@
     "module": "es2020",
     "noImplicitOverride": true,
     "lib": [
-      "es2018",
+      "es2022",
```
miettal, line 23: lib es version is bumped to es2022 to use `.flat()`.

ui/src/app/shared/edge/edgeconfig.ts
```diff
@@ -725,14 +721,17 @@ export module EdgeConfig {
...
-        public readonly description: string;
-        public readonly isRequired: boolean;
-        public readonly defaultValue: any;
-        public readonly schema: {};
+        public readonly id!: string;
+        public readonly name!: string;
+        public readonly description!: string;
+        public readonly type!: string;
+        public readonly isRequired!: boolean;
+        public readonly isPassword!: boolean;
```
miettal, line 729: it seems type and isPassword is mssing so I added these field.

ui/src/app/shared/edge/edgeconfig.ts
```diff
@@ -413,7 +412,7 @@ export class EdgeConfig {
      * Lists all available Factories, grouped by category.
      */
     public static listAvailableFactories(factories: { [id: string]: EdgeConfig.Factory }): CategorizedFactories[] {
-        const allFactories = [
+        const allFactories: CategorizedFactories[] = [
             {
                 category: { title: 'Simulatoren', icon: 'flask-outline' },
                 factories: Object.entries(factories)
```
miettal, line 418: allFactories.factories was mixing flat([]) list and nested([[]]). this is not type first. I rewrite flat-style because this is easily expected from the name `factories`.

ui/tsconfig.json
miettal: generally target es version is lower than lib es version to compatibility.
but this project target es is higher than lib es version. ??
anyway, it is not change of target es version, so I think no effect browser compatibility.

miettal:
rebased!

sfeilmeier (approved):

PR #2701, opened 2024-07-04
fix: add null type to null initialization variables

miettal:
```
$ npx tsc --strict
Found 3646 errors in 210 files.
↓
Found 3666 errors in 188 files.
```

This change adds null type to null initialization variables.
This change increases the number of typescript strict mode errors, but this typing is based on the fact of the code.


miettal:
rebased.

sfeilmeier (approved):

PR #2731, opened 2024-08-03
Unify icon styling

da-Kai:


ui/src/app/shared/type/widget.ts
```diff
@@ -44,6 +44,11 @@ export type Icon = {
     name: string;
 };
 
+export type ImageIcon = {
+    src: string;
+    large: boolean;
```
lukasrgr, line 49: not sure what you want to provide here, shouldnt this be `size`?

ui/src/global.scss
```diff
@@ -396,3 +396,17 @@ ion-modal.full-width {
...
+    margin-top: 12px;
+    margin-bottom: 12px;
+    --border-radius: 0;
+    max-height: 24px;
+    max-width: 24px;
+}
+
+ion-avatar.iconify.icon-large {
+    max-height: 32px;
+    max-width: 32px;
```
lukasrgr, line 411: im always worried about fixed ratios? are we sure this works out for all display res?

lukasrgr (changes requested):

ui/src/global.scss
da-Kai: i copied these values from ion-icon

lukasrgr (approved):
🚀

sfeilmeier (approved):

PR #2738, opened 2024-08-08
Fix for Configuration & Status not populated Fully

Sn0w3y:
This pull request offers two alternative solutions to fix the issue in the `listActiveComponents` method. 

### Changes
```typescript
components.push(...this.getComponentsByFactory(factory.id));
```

### Explanation
In this approach, we use the `push` method to add elements to the `components` array directly. 

### Advantages
- **Simplicity**: The code is straightforward and easy to understand.
- **Performance**: `push` modifies the array in place, which is generally faster than creating a new array with `concat`.

### Disadvantages
- **In-Place Modification**: `push` modifies the original array, which might not be desirable if you need to maintain immutability.

### Changes
```typescript
public listActiveComponents(ignoreComponentIds: string[] = []): CategorizedComponents[] {
    const allComponents = [];
    const factories = this.listAvailableFactories();
    for (const entry of factories) {
        let components: EdgeConfig.Component[] = [];
        for (const factory of entry.factories) {
            components = components.concat(...this.getComponentsByFactory(factory.id));
        }
        allComponents.push({
            category: entry.category,
            components: components,
        });
    }
    // (rest of the code remains the same)
}
```

### Explanation
In this approach, we use `concat` to combine the current `components` array with new elements and assign the result back to `components`.

### Advantages
- **Immutability**: `concat` does not modify the original array; instead, it returns a new array. This can be useful if you want to keep the original array unchanged.

### Disadvantages
- **Performance**: `concat` creates a new array each time, which may be less efficient than `push` for large arrays.
- **Complexity**: This approach adds a bit more complexity by reassigning the array.

### DIFF
--- a/ui/src/app/shared/components/edge/edgeconfig.ts
+++ b/ui/src/app/shared/components/edge/edgeconfig.ts
@@ -586,9 +586,9 @@ export class EdgeConfig {
         const allComponents = [];
         const factories = this.listAvailableFactories();
         for (const entry of factories) {
-            const components: EdgeConfig.Component[] = [];
+            const components: EdgeConfig.Component[] = []; // Option 2: let components: EdgeConfig.Component[] = [];
             for (const factory of entry.factories) {
-                components.push(...this.getComponentsByFactory(factory.id));
+                components.push(...this.getComponentsByFactory(factory.id)); // Option 2: components = components.concat(...this.getComponentsByFactory(factory.id));
             }
             allComponents.push({
                 category: entry.category,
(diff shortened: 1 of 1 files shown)


lukasrgr:
@Sn0w3y i fell over this issue as well, `Array.prototoype.concat` should be used for merging arrays, with concat we have to write some unneccessary code which doesnt make anything better

Sn0w3y:
@lukasrgr can be merged, removed Option 2 concat :)

lukasrgr (approved):

Sn0w3y:
@sfeilmeier can be merged i guess

PR #2755, opened 2024-08-21
fix: fix system status modal.

miettal:
Sorry past my commit has bug for making component list logic.
in this commit I use `concat` but it's not in-place method so this code doesn't work.
This PR rewrte `concat` to `push`.
https://github.com/OpenEMS/openems/commit/2664b1393177f1739f588d3629455e36dce73e82#diff-18c3e87449e26c6e1913ede0123a3d5aa61209b14e6cfb97fac0c9614553c352R591


Sn0w3y:
Hi @miettal 

already done here:

https://github.com/OpenEMS/openems/pull/2738

PR #2778, opened 2024-09-03
Feature/add import order rule to eslint

miettal:
In a96c720293623679c0db8e2db449b9867b7dc4b6, vscode import organize setting is configured.
This PR add same behavior setting to eslint.

1.  import/order is import order rule
2. in default, import/order categorize import like builtin, external internal... and these are separate by each group.
3. But vscode import organization setting don't care it. and the function make 1 group.
4. in this setting, this rule customized by definition all import in 1 group to fit vscode import organization setting.


miettal:
> Error: src/app/edge/history/shared.ts:3:62 - error TS2305: Module '"date-fns"' has no exported member 'de'.

wow, I just added eslint rules and only did `eslint --fix src/`. but error is happened. I convert this PR to draft. I will check it.

lukasrgr:
@miettal are these settings recommended?

miettal:
I rebased this PR to catch up 2024.10 release.

I'm not familiar with bestpractice for typescript import order rules.
But in fork project, current setting will generate many import order diffs from upstream by using vscode editing. this is bad effect to openems ecosystem. its make difficult to management patchset. this is sure.
So I think unifying import order in all codes and forcing import rule are better.

PR #2781, opened 2024-09-09
Add the state of charge in kWh to storage system widgets

j-eissler:
Currently the state of charge (SOC) of storage systems is only shown in %. This change adds the SOC in kWh to the
flat-widget displayed on the dashboard and the modal-widget that pops up when clicking on the storage system.

To do this, the percentagebar-widget needed to be extended to support a secondary parameter, which is displayed in parenthesis behind the percentage value.


sfeilmeier:
Thanks for the contribution. 

The feature was implemented this way purposely by FENECON. In fact we are trying to put less information on the (first pages of) the UI and not more. For most people this information might be too detailed; i.e. your smartphone does not show you the `Wh` of the battery as well.

I believe we should discuss this in the Community. Can you please start a thread with screenshots and link it here?

PR #2807, opened 2024-09-17
Fix: evcs cluster widget missing alias

Jasonlee6789:
Before
image

After
image


lukasrgr:
@Jasonlee6789 default is a empty string right now,
https://github.com/OpenEMS/openems/blob/7ef6c74fb0f90b8d24f0fcef1d076e04fc53070e/io.openems.edge.evcs.cluster/src/io/openems/edge/evcs/cluster/Config.java#L16
 so it is possible that the title is not set, it would be better if we touch the edge config instead, then we can also get rid of the https://github.com/OpenEMS/openems/blob/7ef6c74fb0f90b8d24f0fcef1d076e04fc53070e/ui/src/app/edge/live/Multiple/Evcs_Api_Cluster/Evcs_Api_Cluster.ts#L63

Jasonlee6789:
> @Jasonlee6789 default is a empty string right now,
> 
> https://github.com/OpenEMS/openems/blob/7ef6c74fb0f90b8d24f0fcef1d076e04fc53070e/io.openems.edge.evcs.cluster/src/io/openems/edge/evcs/cluster/Config.java#L16
> 
> 
> so it is possible that the title is not set, it would be better if we touch the edge config instead, then we can also get rid of the
> https://github.com/OpenEMS/openems/blob/7ef6c74fb0f90b8d24f0fcef1d076e04fc53070e/ui/src/app/edge/live/Multiple/Evcs_Api_Cluster/Evcs_Api_Cluster.ts#L63
> 
> FYI @sebastianasen wasnt there a general change in aliases not being empty but taking the id as default?

If this previous design doesn't work, I find the bug is that we should use || replace ??, I have tried ||, it works, I will make the PR.

Jasonlee6789:
After the latest commit 
image

da-Kai (approved):
@lukasrgr you have the final say here

Jasonlee6789:
Replaced ?? with || to handle empty string aliases, check more strictly. This ensures a fallback value is used when alias is missing or empty, like handling empty strings, which improves consistency and prevents alias display issues.

lukasrgr (approved):

PR #2824, opened 2024-09-30
fix: add initialization variables to fix typescript strict mode errors.

Jasonlee6789:
npx tsc --strict:

Found 1522 errors in 194 files.
↓
Found 1505 errors in 196 files.


ui/src/app/edge/history/abstracthistorychart.ts
```diff
@@ -57,6 +57,10 @@ export abstract class AbstractHistoryChart {
         borderColor: "rgba(128,128,0,1)",
     };
 
+
+    private activeQueryData: string = "";
```
lukasrgr, line 61: needed?

lukasrgr (changes requested):

lukasrgr:
@Jasonlee6789 thx for your contribution, im not sure, if this PR has some purpose

ui/src/app/edge/history/abstracthistorychart.ts
Jasonlee6789: Thank you for your review. In my opinion, this change is necessary to meet TypeScript's strict property initialization rule.
By initializing activeQueryData, it ensures it is always defined, after this change, this place TypeScript strict mode error was reduced. Without this explicit initialization, the compiler generates an error because it cannot guarantee that `activeQueryData` will have a value before it’s used.

Jasonlee6789:
> @Jasonlee6789 thx for your contribution, im not sure, if this PR has some purpose
Thank you for your review. The purpose of this PR is to fix uninitialized variables in the UI code and explicitly add null type annotations to variables initialized as null, in order to fix TypeScript strict mode errors and prevent potential runtime errors. 
The result of this PR is that after running `npx tsc --strict`, the number of errors was reduced from 1522 to 1505, eliminating 17 errors.

PR #2827, opened 2024-10-01
Improve Edge Caching Mechanism in InitializeEdgesWorker

Sn0w3y:
This pull request enhances the InitializeEdgesWorker class by implementing a regular caching mechanism for edges from PostgreSQL. The main improvements are:

Regular edge caching: Edges are now cached immediately upon startup and then every 20 minutes.
Simplified scheduling: Uses a ScheduledExecutorService to manage both immediate and periodic tasks.
Executor management: Ensures that executors are properly shut down when stopping the worker.
Improved error handling and logging: Enhanced logging for better error tracing and debugging.

Benefits
- Regular Edge Caching: Keeps the edge cache updated by regularly fetching data from PostgreSQL.
- Simplified Scheduling: Reduces complexity by using a single scheduler for task execution.
- Resource Management: Ensures executors are properly shut down to prevent resource leaks.
- Improved Logging: Enhanced error messages facilitate easier debugging.

### DIFF
--- a/io.openems.backend.metadata.odoo/src/io/openems/backend/metadata/odoo/postgres/InitializeEdgesWorker.java
+++ b/io.openems.backend.metadata.odoo/src/io/openems/backend/metadata/odoo/postgres/InitializeEdgesWorker.java
@@ -5,6 +5,8 @@ import java.sql.PreparedStatement;
 import java.sql.SQLException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


sfeilmeier:
@michaelgrill: Can you please take over? Should we switch this caching to a library that handles updating internally? Guava provides a nice Cache feature: https://github.com/google/guava/wiki/cachesexplained

sfeilmeier:
> I will add a Configurable Time to the Worker aswell as "Enable and Disable" Option :)

One value is sufficient. Negative or zero means "disable".

sfeilmeier:
I had a quick look and renamed the config property to `cacheRefreshInterval`. But `InitializeEdgesWorker` also has to be renamed, because the name does not make any sense anymore. Also Checkstyle currently fails...

katsuya:
@Sn0w3y Thank you so much! We’ve started testing it internally!

PR #2833, opened 2024-10-08
Add hint to README to fix an error with building a Docker image

j-eissler:
This is just a small commit that should help people trying to build their own Docker image. When running the build command from the readme file the error `failed to solve: error from sender: context canceled` can occur. This can easily be fixed by making sure the Eclipse IDE isn't open. To avoid others having to search for a solution I would like to add this info to the readme.

### DIFF
--- a/tools/docker/backend/README.md
+++ b/tools/docker/backend/README.md
@@ -78,3 +78,9 @@
     ```
 
     *for UI Image see [ui/README.md](../ui/README.md)*
+
+# Common Problems and Solutions
+```
+ERROR: failed to solve: error from sender: context canceled
+```
+When building the Docker image this error may occur because another program is accessing the project files. Try closing these programs (e.g. Eclipse IDE) and run the build command again.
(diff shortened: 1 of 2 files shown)


da-Kai:
@j-eissler 
Thank you for your contribution! There are a few improvements I'd suggest to make it more comprehensive:

1. It would be great to move this content to the end of the README under a new section titled `# Common Problems and Solutions`.

2. The same topic should also be included in the backend Docker README, as it's relevant there too.
[tools/docker/backend/README.md](https://github.com/OpenEMS/openems/blob/develop/tools/docker/backend/README.md)

3. The issue you're describing isn't exclusive to Eclipse IDE. It can occur with any program that locks files. Could you please adjust the wording to something like: "Ensure no program is accessing the files", and use Eclipse IDE as an example?

j-eissler:
@da-Kai Thanks for the feedback! I've updated the description according to your response.

da-Kai (approved):

sfeilmeier (approved):

PR #2868, opened 2024-11-10
Center Refresh Spinner

Sn0w3y:


lukasrgr (changes requested):
this style exists already https://github.com/OpenEMS/openems/blob/d58c769bc8ead9a598bb59893f18d6eb8779e56b/ui/src/global.scss#L31

PR #2879, opened 2024-11-19
Modbus: AbstractReadTask: Enhance Parsing Response; exception catch

DerStoecki:
More generic catch block for more detailed channel assignment errors. E.g. the original code is not able to catch the following kind of errors: 
```
[modbus1 ] ERROR [dbus.api.task.AbstractReadTask] Parsing Response failed. 
IllegalArgumentException: Unable to set value [-4395165653183430656] for Channel [battery0/ChrW]: 
Cannot convert. Long [-4395165653183430656] is not fitting in Integer range. 
```

This kind of messages would have been very helpful for detecting channel mapping errors, but it is not catched by OpenemsException


clehne:
@DerStoecki Could you please add an appropriate Junit Testcase, so that Codecov runs through.

sfeilmeier (approved):
Thank you!

PR #2880, opened 2024-11-19
Edge: Backend Api Controller: Change SendChannelsValueWorker SendAllChannelsAfterSeconds to 4 min

DerStoecki:
Due to resolution request, the UI might display empty values in data charts. (5 min time range)
By reducing the send all channel values time to 4 min on the edge, this can be prevented.

The issue only occurs with components that have the same channel value over a long time period.

E.g. an ess over night, where the ess is neither charging or discharging (look at 1 o clock), this is a reoccuring theme, and the image is a mere example.

5MinIssue

### DIFF
--- a/io.openems.edge.controller.api.backend/src/io/openems/edge/controller/api/backend/SendChannelValuesWorker.java
+++ b/io.openems.edge.controller.api.backend/src/io/openems/edge/controller/api/backend/SendChannelValuesWorker.java
@@ -54,7 +54,7 @@ import io.openems.edge.common.type.TypeUtils;
 public class SendChannelValuesWorker {
 
 	private static final int AGGREGATION_MINUTES = 5;
-	private static final int SEND_VALUES_OF_ALL_CHANNELS_AFTER_SECONDS = 300; /* 5 minutes */
+	private static final int SEND_VALUES_OF_ALL_CHANNELS_AFTER_SECONDS = 240; /* 4 minutes */
 
 	private final Logger log = LoggerFactory.getLogger(SendChannelValuesWorker.class);
 


clehne:
Some additional notes. The current situation before this PR is as follows:
1. Edges SendChannelValuesWorker sends constant data every 5 mins. 
2. Due to 1 the backend database has a datapoint every 5 mins only for "constant channel values
3. UI requests the backend rpc method `queryHistoricTimeseriesEnergy` which delivers back 5min average values and UI shows the values with a 5min resolution.

From time to time there is no data in the database for a 5min window. This happens because sometimes the distance between two datapoints is 5.001 mins. Thus UI shows a history chart like in the screenshot above. We have this behaviour seldom, but the "artefact" is shown over all history widgets. 

Our latest edge version reduces SendChannelValuesWorker to 4minutes. With that version all artefacts have been gone.

Note that we haven't tested this with an Edge UI and the edge side rrd4j database. The rrd4j writes data every 5min also. This may lead to other unexpected side effects.

sfeilmeier:
Note that this would cause quite a performance problem for us. We implemented a randomizer because of too much data sent at once - see: https://github.com/OpenEMS/openems/pull/2882/commits/d21b7b4e0176485b539f0e25ed0fc57af467ca6c

Are you using Timedata Aggregated InfluxDB (https://github.com/OpenEMS/openems/tree/develop/io.openems.backend.timedata.aggregatedinflux)? 

Timestamp `5.001 min` sounds like it could be solvable with a different query, too.

DerStoecki:
Closed - Bc not necessary - use aggregated Influx -

PR #2892, opened 2024-11-29
Fix EssDcCharger nature on modbus

simon-lgb:
The hash of EssDcCharger devices on the Modbus interface currently incorrectly identifies them as SymmetricEss. This PR fixes the problem by replacing the incorrect class passed to the ModbusSlaveNatureTable.of method.


sfeilmeier (approved):
Thanks! 👍

Issue #2906, opened 2024-12-03
Ocpp server fail to send commands to the charging stations when two connections are established


katsuya:
### Description

When two connections are established to the same `charge point A`, the following bug occurs, causing OpenEMS to fail in properly handling the previous connection:

![384244385-f3e38a07-7a33-4a1f-81e1-f9663c18631d](https://github.com/user-attachments/assets/eefdd023-9a1e-48c7-8925-5c530aa36bf9)

- When `connection B` is established, and a `lostSession` is triggered in `activeEvcSessions` for `connection A`, it incorrectly affects `charge point A` as a whole. This results in `connection B` malfunctioning as well.

The following issues should be addressed (esolving the first issue alone might be sufficient):

1. When a new connection is established, the old connection must be correctly removed from `activeEvcSessions`.
  - https://github.com/OpenEMS/openems/blob/263377b6282b9e812c218b045831c2f990ee0c1b/io.openems.edge.evcs.ocpp.server/src/io/openems/edge/evcs/ocpp/server/MyJsonServer.java#L105
3. When the old connection is disconnected, it should ensure that the `lostSession` for the `charge point` referenced by the new connection is not invoked.
  - https://github.com/OpenEMS/openems/blob/263377b6282b9e812c218b045831c2f990ee0c1b/io.openems.edge.evcs.ocpp.server/src/io/openems/edge/evcs/ocpp/server/MyJsonServer.java#L117
5. Disconnection of the old connection should not remove the `charge point` from `ocppSessions` associated with the new connection.
  - https://github.com/OpenEMS/openems/blob/263377b6282b9e812c218b045831c2f990ee0c1b/io.openems.edge.evcs.ocpp.server/src/io/openems/edge/evcs/ocpp/server/MyJsonServer.java#L128

### Operating System

Ubuntu

### How to reproduce the Error?

1. Connect `charge point A` to the OCPP server.  
2. Establish another connection between `charge point A` and the OCPP server without disconnecting the first connection.  
3. Disconnect the first connection.

katsuya:
> @katsuya: OCPP is not actually used right now on our side, so this has no high priority ATM.

@sfeilmeier That's really unfortunate. We've started using it, so if we send a patch from our side, would you consider reviewing it?

sfeilmeier:
> @sfeilmeier That's really unfortunate. We've started using it, so if we send a patch from our side, would you consider reviewing it?

[...]

PR #2908, opened 2024-12-03
Improve Session Management for Charge Point Connections

Sn0w3y:
#### Summary
This PR addresses the issue of improper session management when multiple connections are established to the same charge point. Specifically, it ensures that only one active session per charge point is maintained, and existing sessions are properly terminated before establishing new ones. This update helps prevent unintended side effects where disconnecting an older connection impacts newer connections.

#### Changes Made

1. **Handle Existing Sessions When a New Connection is Established**
   - Updated `newSession` method in `MyJsonServer.java`:
     - Checks if an existing session already exists for a given charge point.
     - If found, gracefully terminates the old session by invoking `lostSession()` and removes it from active sessions.
     - Associates the new session with the charge point and initializes it accordingly.

2. **Prevent `lostSession` from Affecting New Connections**
   - Updated `lostSession` method in `MyJsonServer.java`:
     - Ensures that only the intended session is terminated without affecting other active connections.
     - Before removing a session from `ocppSessions`, it verifies that the session being disconnected is still the active one.

3. **Ensure Proper Cleanup in `removeEvcs`**
   - Reviewed and modified `removeEvcs` method in `EvcsOcppServer.java`:
     - Ensures that removing an EVCS component correctly handles session mappings without affecting other active sessions.
     - Properly removes the EVCS component from session lists and invokes `lostSession()` for cleanup.

#### Additional Enhancements
- **Thread Safety**: Made sure that shared resources such as `activeEvcsSessions` and `ocppSessions` are accessed in a thread-safe manner.
- **Logging Improvements**: Enhanced logging to include more contextual information to facilitate easier debugging and monitoring of session activities.
- **Session Validation**: Added additional checks to validate the integrity of sessions to ensure correct mapping between charge points and sessions.
- **Resource Cleanup**: Ensured proper resource cleanup upon session termination to prevent memory leaks and stale references.

#### Testing
- Please test this Code Improvement as i do not have the ability to test it on my side.
- @katsuya please take this task maybe as you have the issue right now in:
https://github.com/OpenEMS/openems/issues/2906

#### Motivation
These changes address the observed problem where OpenEMS failed to handle multiple simultaneous connections to the same charge point, causing instability in session management. By implementing these changes, the system will now correctly manage and isolate sessions, ensuring reliable operation of charge point connections.

#### Related Issues
- Fixes improper handling of multiple charge point connections.
- Resolves issues with old connections impacting newly established sessions.

#### Notes
If further refinements are needed, or any issues persist with session handling, please feel free to provide feedback.


katsuya:
@Sn0w3y Thanks a lot! We'll give the provided patch a try.

> Thread Safety: Made sure that shared resources such as activeEvcsSessions and ocppSessions are accessed in a thread-safe manner.

This is absolutely true. Occasionally, we also experience erratic behavior, and it's sometimes difficult to determine if this is the cause, which can be quite frustrating.

By the way, here’s the fix we implemented. It mostly resolved the issue, but we’re still experiencing some very rare instances of erratic behavior, which is troubling.

```
From 342832e40892ae147dc020edd37d93fc297b52e1 Mon Sep 17 00:00:00 2001
From: Katsuya Oda <oda.katsuya@pplc.co>
Date: Wed, 4 Dec 2024 09:05:59 +0900
Subject: [PATCH] fix: remove previous session

---
 .../src/io/openems/edge/evcs/ocpp/server/MyJsonServer.java  | 6 +++++-
 1 file changed, 5 insertions(+), 1 deletion(-)

diff --git a/io.openems.edge.evcs.ocpp.server/src/io/openems/edge/evcs/ocpp/server/MyJsonServer.java b/io.openems.edge.evcs.ocpp.server/src/io/openems/edge/evcs/ocpp/server/MyJsonServer.java
index 66d4d78b3..e461755a2 100644
--- a/io.openems.edge.evcs.ocpp.server/src/io/openems/edge/evcs/ocpp/server/MyJsonServer.java
+++ b/io.openems.edge.evcs.ocpp.server/src/io/openems/edge/evcs/ocpp/server/MyJsonServer.java
@@ -90,7 +90,7 @@ public class MyJsonServer {
 
 				var ocppIdentifier = information.getIdentifier().replace("/", "");
 
-				MyJsonServer.this.parent.ocppSessions.put(ocppIdentifier, sessionIndex);
+				var sessionIndexPrevious = MyJsonServer.this.parent.ocppSessions.put(ocppIdentifier, sessionIndex);
 
 				var presentEvcss = MyJsonServer.this.parent.ocppEvcss.get(ocppIdentifier);
 
@@ -103,6 +103,10 @@ public class MyJsonServer {
 					evcs.newSession(MyJsonServer.this.parent, sessionIndex);
 					MyJsonServer.this.sendInitialRequests(sessionIndex, evcs);
 				}
+
+				if (sessionIndexPrevious != null) {
+					MyJsonServer.this.parent.activeEvcsSessions.remove(sessionIndexPrevious);
+				}
 			}
 
 			@Override
-- 
2.39.5 (Apple Git-154)
```

io.openems.edge.evcs.ocpp.server/src/io/openems/edge/evcs/ocpp/server/EvcsOcppServer.java
```diff
@@ -110,21 +110,33 @@ protected void removeEvcs(Evcs evcs) {
...
-			} else {
-				this.activeEvcsSessions.get(ocppEvcs.getSessionId()).remove(ocppEvcs);
+		var sessionId = ocppEvcs.getSessionId();
+		if (sessionId != null) {
+			var evcss = this.activeEvcsSessions.get(sessionId);
+			if (evcss != null) {
+				evcss.remove(ocppEvcs);
+				if (evcss.isEmpty()) {
+					this.activeEvcsSessions.remove(sessionId);
+					// Also remove from ocppSessions if needed
```
katsuya, line 120: In actual production environments, we rarely remove components, so we went ahead with weekend testing without including this commit. However, I don’t see any issues with the code itself.

io.openems.edge.evcs.ocpp.server/src/io/openems/edge/evcs/ocpp/server/MyJsonServer.java
```diff
@@ -90,6 +90,23 @@ public void newSession(UUID sessionIndex, SessionInformation information) {
 
 				var ocppIdentifier = information.getIdentifier().replace("/", "");
 
+				// Check if there is already a session for this ocppIdentifier
```
katsuya, line 93: Over the weekend, we tested both our patch(https://github.com/OpenEMS/openems/pull/2908#issuecomment-2515850362) and this one, and both appeared to work properly. However, I’m concerned about the following point in this patch. What do you think?

- The old session is deleted before establishing a new one(https://github.com/OpenEMS/openems/pull/2908/files#diff-a9d06149a614c84998d6bcad6d51f36ca40834f2e200120f2a2e9bfb290601eaR119). If something goes wrong during that process, wouldn’t we end up unable to use both the old and new sessions?

io.openems.edge.evcs.ocpp.server/src/io/openems/edge/evcs/ocpp/server/MyJsonServer.java
```diff
@@ -122,10 +139,16 @@ public void lostSession(UUID sessionIndex) {
 				for (Entry<String, UUID> session : MyJsonServer.this.parent.ocppSessions.entrySet()) {
```
katsuya, line 139: To address the issue mentioned below, we added the following changes. However, these modifications might not be necessary if our only concern is the brief window between establishing the new session and discarding the old one, as they make the code more complex.

> When the old connection is disconnected, it should ensure that the lostSession for the charge point referenced by the new connection is not invoked.

```java
        if (sessionEvcss != null) {
          for (AbstractManagedOcppEvcsComponent ocppEvcs : sessionEvcss) {
            var ocppId = ocppEvcs.getConfiguredOcppId();
            UUID latest = MyJsonServer.this.parent.ocppSessions.get(ocppId);
            /**
             * Since multiple sessions are connected with the same ocppId, if
             * `lostSession` is executed with an old session, the `sessionId` of the
             * component corresponding to the latest session would be set to null.
             */
            if (latest != sessionIndex) {
              MyJsonServer.this.logWarn("skip lostSession of old session [" + sessionIndex +
                  "] for ocppIdentifier [" + ocppId + "]");
              continue;
            }
            ocppEvcs.lostSession();
            MyJsonServer.this.logWarn("lostSession [" + sessionIndex + "] for ocppIdentifier [" + ocppId + "]");
          }
          MyJsonServer.this.parent.activeEvcsSessions.remove(sessionIndex);
          MyJsonServer.this.logWarn("remove activeEvcsSessions [" + sessionIndex + "]");
        }
```

io.openems.edge.evcs.ocpp.server/src/io/openems/edge/evcs/ocpp/server/MyJsonServer.java
```diff
@@ -122,10 +139,16 @@ public void lostSession(UUID sessionIndex) {
...
 					if (session.getValue().equals(sessionIndex)) {
 						ocppId = session.getKey();
+						break;
 					}
 				}
 
-				MyJsonServer.this.parent.ocppSessions.remove(ocppId);
+				// Before removing, check if the sessionIndex is still the one mapped to ocppId
```
katsuya, line 146: This code also deals with the window time between establishing the new session and discarding the old one inside `newSession`, so it might not be necessary after all. During our weekend tests, we couldn’t hit this code path without artificially introducing sleeps, which suggests we may not need it.

katsuya (commented):
@Sn0w3y It seems that the fix for the following issue has been removed and replaced with the changes from https://github.com/OpenEMS/openems/pull/2912. Was this intentional? Since the root causes are different, I think it would be better to address them separately.

> When a new connection is established, the old connection must be correctly removed from activeEvcSessions.

PR #2930, opened 2024-12-17
fix: Use correct modbus addresses for Janitza UMG96RM(E) meter

parapluplu:
Implementation used 8xx addresses which are programmable. Switched to 190xx addresses, which cannot change. See https://www.janitza.com/files/download/manuals/current/UMG96RM/E/janitza-mal-umg96rme-en.pdf page 19.


sfeilmeier:
@pooran-c: Could you please review?

parapluplu:
To check the other ones I need them as hardware devices, which I don't have. So it will be hard to verify it. For this specific device I was able to verify the fix works

PR #2935, opened 2024-12-20
UI: Overview search: trigger on every input change

Sn0w3y:
### **Description:**
This PR updates the search functionality in the `overview.component.html` by modifying the event that triggers the search action when the user types in the `ion-searchbar`. 

### **Changes:**
- Replaced the `(ionChange)` event with the `(input)` event on the `ion-searchbar`. 
- This change ensures that the `searchOnChange()` method is called immediately when the input value changes, allowing the search to trigger as the user types, rather than only after pressing enter or losing focus.

### **Reason for Change:**
- The `(ionChange)` event would only fire when the input field loses focus or the user presses the Enter key. This made the search feel unresponsive for real-time input. By using `(input)`, the search updates dynamically with every keystroke.
  
### **Impact:**
- This change enhances the user experience by providing immediate feedback on search input.
- It ensures a more responsive and seamless interaction with the search bar.

### DIFF
--- a/ui/src/app/index/overview/overview.component.html
+++ b/ui/src/app/index/overview/overview.component.html
@@ -42,7 +42,7 @@
                                 <ion-col>
                                     <ion-searchbar
                                         style="--background: var(--ion-color-searchbar-background); color:var(--ion-color-searchbar-text);"
-                                        [(ngModel)]="query" (ionChange)="searchOnChange()">
+                                        [(ngModel)]="query" (input)="searchOnChange()">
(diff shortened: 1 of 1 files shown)


lukasrgr:
@Sn0w3y good that you caught this, apparently ionic decided to change their behavior of `ionChange`

shouldnt we use [ionInput](https://ionicframework.com/docs/api/searchbar#events) instead of input?

sfeilmeier (approved):

PR #2946, opened 2024-12-25
Fix resend historic data function

cvabc:
- Backend's write function is implememted
- RRD4j's floating number data is converted accordingly in Edge


michaelgrill (changes requested):
I hope my comments are understandable, feel free to ask if they are not.

PR #2958, opened 2025-01-07
Fix component.update View

Sn0w3y:
Before:

image

After:

image

Issue:

In some ways the Text overwrites the input container (See Delay before TX):

image

This change is just an example for a "Quick Fix"

PR #2962, opened 2025-01-13
UI: fix formly form styles

lukasrgr:
follow up to: #2958 

**Changes**

fixing styles permanently for `component-update` and `component-install` and all other components, which are using the default formlyfield type `input`


venu-sagar (approved):

lukasrgr:
@sfeilmeier there would be a permanent solution to this issue, if using labelPlacement floating, this would change the style completely, what do you think?

sfeilmeier:
I could live with the change... https://ionicframework.com/docs/api/input#label-placement

PR #2963, opened 2025-01-13
Switch from `ionChange` to `ionInput` for Real-Time Filtering in Search Bar

Sn0w3y:
This PR updates the `<ion-searchbar>` event from `(ionChange)` to `(ionInput)`. By listening to each keystroke, the filter is applied in real-time as the user types. This delivers immediate feedback on search queries, which can enhance the overall user experience for quick lookups or rapidly changing query inputs.

**Why This Makes More Sense**  
1. **Real-Time Results**: Users see the updated list as they type, making searches feel more interactive and giving them instant feedback.  
2. **Improved UX for Quick Searches**: When dealing with smaller data sets or faster filter operations, immediate updates can be more intuitive than having to confirm or leave the input field first.  
3. **Common Pattern in Live Searches**: Many search experiences provide real-time suggestions or filtering, aligning with modern UX expectations.

---

**Changes**  
- Replaced `(ionChange)="updateFilter($event.detail.value)"` with `(ionInput)="updateFilter($event.detail.value)"` in the `<ion-searchbar>`.  
- Kept existing styling and filter logic intact.  

Please review and let me know if there are any additional improvements or considerations needed!

### DIFF
--- a/ui/src/app/edge/settings/component/install/index.component.html
+++ b/ui/src/app/edge/settings/component/install/index.component.html
@@ -1,6 +1,6 @@
 <ion-content>
   <ion-searchbar style="--background: var(--ion-color-searchbar-background); color:var(--ion-color-searchbar-text);"
-    (ionChange)="updateFilter($event.detail.value)"></ion-searchbar>
+    (ionInput)="updateFilter($event.detail.value)"></ion-searchbar>
   <ion-list>
     <ng-container *ngFor="let entry of list">
       <ion-item-group>


lukasrgr (approved):

sfeilmeier (approved):

lukasrgr:
@Sn0w3y if you got time, theres still an issue with overloads of the jsonrpc responses for getEdges, 

In terms of developing mobile first, the `onEnter` key event makes completely sense with everybody pressing enter or search on their keyboard, on desktop this might be not to obvious.

The overloads for the jsonrpc requests would be definitely solved with the onEnter event.

Are there use cases where you want to filter for multiple ems`s at once and is there a need to see the results per character typed?

PR #2265, opened 2023-07-11
Bug in Edge2Edge in ModbusType Float64 fixed.

tsicking:
Fixed Typo: Changed UnsignedQuadrupleWordElement to FloatQuadrupleWordElement.

PR #2972, opened 2025-01-17
Add active charge energy and active discharge energy to modbus table

Femumme:
Add active charge energy and active discharge energy to modbus nature table of SymmetricEss

This will allow the Edge2EdgeEss to read those channels and export correct historical data.

Since this change is intended to work with a FEMS, I would love to see this PR accepted and imported into FEMS soon :-)


Femumme:
Hint: there is another PR open that fixes a Bug in the FLOAT64 Type that is used here:
#2265

tsicking:
Note: This has been handled in #3234, the Edge2EdgeEss calculates the energy from its power now.

Issue #2984, opened 2025-01-21
FEMS Backports 2025-01-20 breaks local deployment with empty config


Lamarqe:
### Description

See https://community.openems.io/t/fehler-beim-lokalen-start-von-openems/5609.

https://github.com/OpenEMS/openems/commit/878f7e776a6787b49b24da773f131e0b7ed633ac breaks a fresh installation as described [here](https://openems.github.io/openems.io/openems/latest/gettingstarted.html).

```
FrameworkEvent ERROR

org.osgi.framework.ServiceException: Service factory returned null

```
caused by

```
java.lang.VerifyError: Inconsistent stackmap frames at branch target 273
Exception Details:
  Location:
    io/openems/edge/core/componentmanager/EdgeConfigWorker.readConfigurations(Lio/openems/common/types/EdgeConfig$ActualEdgeConfig$Builder;Ljava/lang/String;)Z @273: aload
  Reason:
    Type 'java/lang/Object' (current frame, locals[10]) is not assignable to 'java/lang/String' (stack map, locals[10])
  Current Frame:
    bci: @203
    flags: { }
    locals: { 'io/openems/edge/core/componentmanager/EdgeConfigWorker', 'io/openems/common/types/EdgeConfig$ActualEdgeConfig$Builder', 'java/lang/String', '[Lorg/osgi/service/cm/Configuration;', 'java/util/Set', 'org/osgi/service/cm/Configuration', integer, integer, '[Lorg/osgi/service/cm/Configuration;', 'java/util/Dictionary', 'java/lang/Object' }
    stack: { 'java/lang/Object' }
  Stackmap Frame:
    bci: @273
    flags: { }
    locals: { 'io/openems/edge/core/componentmanager/EdgeConfigWorker', 'io/openems/common/types/EdgeConfig$ActualEdgeConfig$Builder', 'java/lang/String', '[Lorg/osgi/service/cm/Configuration;', 'java/util/Set', 'org/osgi/service/cm/Configuration', integer, integer, '[Lorg/osgi/service/cm/Configuration;', 'java/util/Dictionary', 'java/lang/String' }
    stack: { }

```

### Operating System

Ubuntu Linux 24.04

### How to reproduce the Error?

Follow https://openems.github.io/openems.io/openems/latest/gettingstarted.html

Lamarqe:
Thanks for the prompt reaction.

PR #2987, opened 2025-01-22
Fix VerifyError in EdgeConfigWorker

sfeilmeier:
Fix java.lang.VerifyError: Inconsistent stackmap frames at branch target 273 that appeared after #2976.

I don't know why, but this fixes the problem.

See #2984

### DIFF
--- a/io.openems.edge.core/src/io/openems/edge/core/componentmanager/EdgeConfigWorker.java
+++ b/io.openems.edge.core/src/io/openems/edge/core/componentmanager/EdgeConfigWorker.java
@@ -9,6 +9,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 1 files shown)


vision1727:
Jän 23 21:16:00 debian java[3193041]: 2025-01-23T21:16:00,527 [artLevel] ERROR [ntmanager.ComponentManagerImpl] bundle io.openems.edge.core:1.0.0.202501231943 (86)[io.openems.edge.core.componentmanager.ComponentManagerImpl(149)] : Error during instantiation of the implementation object

is this coming from the same issue?

vision1727:
Loaded the actual repo from github. Build it an then started OpenEms in commandline. then i got this error. in configmananger. 

happend in a clean fresh installation

vision1727:
It was all cleaned up before starting eclipse and before loading repo. I did this 3 times already every time the same.

Issue #2907, opened 2024-12-03
Select date


cf99:
### Component

UI

### Description

It would be great to compare different times e.g. week 51 2024 to week 51 2023 (also day, month). But only the bars not the curvs..

machimozor:
hi, i woul like to work on this issue @cf99

PR #2988, opened 2025-01-23
Feature/graphic compare bydate

machimozor:
This Pull Request addresses issue #2907. In the Chart component, I propose a solution to compare dates between two periods.

Changes Made
Added logic to allow comparison of dates between two specified periods within the Chart component.


machimozor:
Hello, Thank you for your comments. I am trying to fix it ..

Greetings

PR #3012, opened 2025-02-11
fix(backend): display user role as desired by message

parapluplu:
When access is denied due to missing role privileges, currently the desired role is added to the error message. However, the error message indicates that the current user role is printed.


sfeilmeier (approved):

sfeilmeier:
Thank you!

PR #3020, opened 2025-02-20
Supply allowed string values of @AttributeDefinition channels

Lamarqe:
Until now, AttributeDefinition channels (channels, whose names start with _Property) which are based on Enums do not provide the allowed values in their type defintions retrieved via edge call `getChannelsOfComponent()`.

For example the channel ctrlEvcs0/_PropertyChargeMode provides:
```
    {
        "id": "_PropertyChargeMode",
        "accessMode": "RO",
        "persistencePriority": "HIGH",
        "text": "",
        "type": "STRING",
        "unit": "",
        "category": "OPENEMS_TYPE"
    },
```

With this pull request, the allowed enum values (which were stored internally already previously) are published correctly:

```
    {
        "id": "_PropertyChargeMode",
        "accessMode": "RO",
        "persistencePriority": "HIGH",
        "text": "",
        "type": "STRING",
        "unit": "",
        "category": "OPENEMS_TYPE",
        "options": [
            "FORCE_CHARGE",
            "EXCESS_POWER"
        ]
    },
```

### DIFF
--- a/io.openems.common/src/io/openems/common/jsonrpc/request/GetChannelsOfComponent.java
+++ b/io.openems.common/src/io/openems/common/jsonrpc/request/GetChannelsOfComponent.java
@@ -7,8 +7,14 @@ import static java.util.stream.Collectors.mapping;
 import static java.util.stream.Collectors.toList;
 
 import java.util.List;
+import java.util.ListIterator;
 import java.util.stream.Stream;
+import java.lang.reflect.Type;
+import com.google.gson.reflect.TypeToken;
 
+import com.google.gson.Gson;
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
 import com.google.gson.JsonPrimitive;
 
 import io.openems.common.channel.AccessMode;
@@ -104,6 +110,10 @@ public class GetChannelsOfComponent implements EndpointRequestType<Request, Resp
 			// category = ChannelCategory.STATE
 			Level level, //
 
+			// category = ChannelCategory.OPENEMS_TYPE
+			// type = OpenemsType.STRING
+			List<String> stringOptions,
+
 			// category = ChannelCategory.ENUM
 			List<OptionsEnumEntry> options //
 	) {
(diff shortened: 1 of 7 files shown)


Lamarqe:
No idea. 
I don't see any connection to this pull request.

Lamarqe:
Is there anything which I need to do in order to receive a review or other feedback about how to proceed?

sfeilmeier:
Thank you for the contribution. I'll need some time to review it in detail. Ideally I would like to not add it to the parent Doc directly as it relates only to Enums. Also the code does not fully align with our coding guidlines (https://openems.github.io/openems.io/openems/latest/contribute/coding-guidelines.html)

Lamarqe:
Regarding the code formatting related warnings:
If you are generally fine with merging the changes as they are, I can take care of these. 

Is the assumption correct that there is a tool and some central config somewhere for eclipse available which applies automatically the correct code style?
Any directions would be highly appreciated, as its a bit error prone to correct the style without immediate feedback and verification of the changes.

Lamarqe:
Thank you for having a first look and taking the time for a full review.
I guess with "parent doc", you refer to Doc.java? I did not find a way to modify only AbstractDoc.java.
Logically, its an enum channel, but technically, it is not. (Its ChannelCategory.OPENEMS_TYPE using OpenemsType.STRING. The only difference is the schema it comes with.) 

Regarding the the style warnings: During the last hour, I applied the Eclipse code formatter to all modified files and updated the branch accordingly. PR should be clean now.

Lamarqe:
@sfeilmeier : do you think you will find the time to do a full review anytime soon?

Lamarqe:
Is there any chance to get this MR merged?
If not, please let me know and share a short explanation  

Then I would just close it and stop investing time in doing regularly rebases and retests.

Lamarqe:
Yes, good question.

Having this information available in the channel definition will allow for a generic integration into other smart home systems.
Instead of hard-coding behaviors and options, integrations to connected systems can use the information retrieved via the `GetChannelsOfComponent` reponse to define possible user and system interactions.

One example of such a connected system could be Home Assistant. But this MR is fully independent of the concrete connected system.

sfeilmeier:
Sorry, I hardly find time to work on this feature. The two reasons why I don't immediately accept your current implementation:
- In `EdgeConfig` we already parse Enums correctly, when creating the Schema (https://github.com/OpenEMS/openems/blob/develop/io.openems.common/src/io/openems/common/types/EdgeConfig.java#L821). This info would have to be propagated to where the Channel is created (https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.common/src/io/openems/edge/common/component/AbstractOpenemsComponent.java#L306)
- There are no JUnit tests to validate this new behaviour. We just had a similar issue with our API for external home automation systems which also only happend because nobody was aware a dependency update could break it (see https://github.com/OpenEMS/openems/pull/3068)

Lamarqe:
Hi @sfeilmeier,

As this would be a really valuable addition, I went over it again with the intention to adjust the implementation according to your feedback and create unit test coverage. Regarding your implementation comment:

> In `EdgeConfig` we already parse Enums correctly, when creating the Schema (https://github.com/OpenEMS/openems/blob/develop/io.openems.common/src/io/openems/common/types/EdgeConfig.java#L821). This info would have to be propagated to where the Channel is created (https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.common/src/io/openems/edge/common/component/AbstractOpenemsComponent.java#L306)

This is exactly what my implementation does. The EdgeConfig enum parsing stores the options in the schema field templateOptions->options. My implementation gets this information exactly in the point you recommend ("where the Channel is created") See (https://github.com/Lamarqe/openems/blob/c28f59b9c4c127731a7f05f459cef21a6466a7d9/io.openems.edge.common/src/io/openems/edge/common/component/AbstractOpenemsComponent.java#L359-L364)

If this is not adequate: Could you give a few more details of how you would like to have the implementation changed?

Lamarqe:
As part of re-basing, I noticed a recent [code change](https://github.com/OpenEMS/openems/commit/8a4fc7e74d8bd2a1b4baeee0f9476fce2ce9074c) which is similar to the extension made with my PR. This gives me better confidence that my changes are aligned with the existing implementation.

Also, I added today the unit tests requested by Stefan. 
With that, I am sending it again for review with the request to merge.

Lamarqe:
@sfeilmeier : Can we please move on with this PR? Earlier review comments have been incorporated / answered.

PR #3039, opened 2025-03-05
MQTT: avoid duplicate connect attempts

MurselKhan:
synchronize reconnect logic to avoid duplicate attempts, update logWarn to use proper log level


sfeilmeier (approved):

sfeilmeier:
Thank you for the contribution!

PR #3049, opened 2025-03-14
Implement Shelly Plus Add-On

sthhs:
Implementation of requesting the input values from a Shelly Plus Add-On.
There are different input types:

- Temperature
- Humidity (always in combination with temperature)
- Voltage
- Analog
- Digital

Each input is configured as a separate component. The type and index (100 - 104) have to correspond with the ones for this input in the Shelly configuration page.
For avoiding errors, the type and the index can only be selected by drop down menus.

The request is the same as of the io.openems.edge.io.shelly.shellyplus1pm component: http://ip-address/rpc/Shelly.GetStatus
In the resulting JSON the corresponding objects are evaluated:

`,
"input:100": {
  "id": 100,
  "state": false
},
"temperature:100": {
  "id": 100,
  "tC": 48.2,
  "tF": 118.9
},
  "input:100": {
    "id": 100,
    "percent": 47.1
},
"voltmeter:100": {
  "id": 100,
  "voltage": 4.71
},
`
For temperature only the Celsius value ("tC") is  read.
The channels are of type Integer, except for the input type "Digital", which is of type Boolean.

The values are stores as mV (voltage), dC (temperature), percent (humidity and analog) and true/false (digital).

The corresponding console output looks like this:
`ioShellyInput101[516 dC] ioShellyInput3100[271 dC] ioShellyInput3101[411 dC] ioShellyInput3102[321 dC] ioShellyInput3103[445 dC] ioShellyInput3104[401 dC] ioShellyInput4100[487 dC] ioShellyInputDigital100[false] ioShellyInputVolt100[4710 mV]
`

### DIFF
--- a/io.openems.edge.io.shelly/src/io/openems/edge/io/shelly/shellyplus1pmaddon/AddOnEnums.java
+++ b/io.openems.edge.io.shelly/src/io/openems/edge/io/shelly/shellyplus1pmaddon/AddOnEnums.java
@@ -1,0 +1,32 @@
+package io.openems.edge.io.shelly.shellyplus1pmaddon;
+
+public class AddOnInputType {
+	
+	public enum InputType {
+		TEMPERATURE,
+		TEMPERATURE_AND_HUMIDITY,
+		VOLTAGE,
+		DIGITAL_INPUT,
+		ANALOG_INPUT,
+		NONE;
+	}
+
+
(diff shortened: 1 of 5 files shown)


NiklasTuepker:
I like the idea behind this topic. I'm not a developer. :( How can it be advanced?

PR #3050, opened 2025-03-15
Implementation: Variant of ShellyPlus1PM for more than one Shelly model

sthhs:
This variant of the ShellyPlus1PM component is capable of covering more than one Shelly model.
The Shelly models can be defined in the enum io.openems.edge.io.shelly.common.Utils.ShellyModel with name, power channel and a flag indicating if it has a relay.
Tested with the Shelly Plus PM Mini and Shelly Plus 1 PM..

### DIFF
--- a/io.openems.edge.io.shelly/src/io/openems/edge/io/shelly/common/Utils.java
+++ b/io.openems.edge.io.shelly/src/io/openems/edge/io/shelly/common/Utils.java
@@ -15,7 +15,48 @@ import io.openems.edge.common.component.OpenemsComponent;
 
 public class Utils {
 
+	public enum ShellyModel {
+		/**
+		 * Defines a Shelly Plus 1 PM with relay
+		 * power can be read from the object "switch:0"
+		 */
+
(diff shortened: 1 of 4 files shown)


sthhs:
@Sn0w3y surely it would make more sense to use another class name.

sfeilmeier:
Thanks for the contribution. I think we can combine your ideas. I think it is feasibly to combine implementations that are very similar and really only differ in the name inside the JSON response (`switch:0` vs. `pm1:0`). If the available channels differ, it might be better to split it in components - i.e. I would distinguish between Shellys with and without internal Relay.

This also makes some advanced internals like choosing a relay output in the App Center easier (see https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.core/src/io/openems/edge/app/common/props/RelayProps.java)

 Of course, when splitting we should try to reuse as much code as possible with helper classes and try to have as much JUnit test coverage as possible - to avoid breaking code, when later adding new compatibilities.

PR #3052, opened 2025-03-17
Improve querylimit handling to avoid unintended increment due to type conflicts

ikegam:
Previously, the backend incremented querylimit even for type conflicts in InfluxDB queries. Since querylimit should only reflect actual overload cases, this change modifies the handling logic to prevent increments caused solely by type conflicts.

I have tested this PR, and it appears to be working as expected. However, since this affects a core part of the backend functionality, I'm opening it initially as a Draft PR for further review and discussion.

### DIFF
--- a/io.openems.backend.timedata.influx/src/io/openems/backend/timedata/influx/FieldTypeConflictHandler.java
+++ b/io.openems.backend.timedata.influx/src/io/openems/backend/timedata/influx/FieldTypeConflictHandler.java
@@ -37,9 +37,10 @@ public class FieldTypeConflictHandler {
 	 * fields that already exist in the database.
 	 *
 	 * @param e the {@link FieldTypeConflictException}
+	 * @return true means handled successfully (data written to influxdb)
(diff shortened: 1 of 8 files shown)


io.openems.backend.timedata.influx/src/io/openems/backend/timedata/influx/FieldTypeConflictHandler.java
ikegam, line 1: https://github.com/OpenEMS/openems/pull/3052/files#diff-600033dd350337b45882b26cb482d1a8827faab63f1866da00b505641852c437L45-R71

This "handleExceptionMessage" return true and false and it seems it successfully adds handler then true so I suppose this can be used for conditions to make this PR.

sfeilmeier:
@michaelgrill: Could you have a look please?

ikegam:
We have operated our backend several months and it works well as expected with this patch. so I have updated the PR to ready.

PR #3067, opened 2025-03-25
Adds limiting EVCS Cluster, ESS optional and minimum guaranteed charge power

chpoulter:
Added a channel to allow limiting maximum power that is distributed among the EVCS. This is required when reducing power because of §14a EnWG. Am planing to write a Controller to maintain this value. Hope I did not miss some other way... 

While implementing I made the minimum guaranteed charge power configurable (as the hard coded 4500 can be a problem for §14a EnWG). I also made ESS optional, so there is no need anymore to setup a simulated ESS.

Hope this contribution is useful and of interest.


chpoulter:
sorry ... did not know that renaming the branch will close this PR ... created a new one as #3084

PR #3084, opened 2025-04-07
Adds ManagedEvcsCluster with Limit and Evcs count, makes ESS optional and minimum guaranteed charge power

chpoulter:
Adds ManagedEvcsCluster with an active power limit and evcs count channel to allow limiting maximum power that is distributed among the EVCS. This is required when reducing power because of §14a EnWG or may be useful for other things. I am writing on a controller to maintain this limit. Hope I did not miss some other way.

While implementing I made the minimum guaranteed charge power configurable (as the hard coded 4500 can be a problem for §14a EnWG). I also made ESS optional, so there is no need anymore to setup a simulated ESS.

Sorry for creating a new PR ... I've renamed the branch and did not know it would close the existing PR.


chpoulter:
I've only tested two AC stations, as that is my scenario. But basically I did not change the existing distribution method, so it should work as before (->unit tests are still all running). Only the maximum available power to distribute is lowered to the value on the new channel.

To be honest I do not really see any different handling for AC / DC stations in EvcsClusterPeakShaving. I think for your case it will just distribute requested/guaranteed power in the sorted order to the stations until the limit is hit. So if there is not enough power, some stations will be switched off. If there is power left over (which might happen, if not all stations are running) it seems to be additionally distributed.

io.openems.edge.evcs.cluster/src/io/openems/edge/evcs/cluster/EvcsClusterPeakShavingImpl.java
```diff
@@ -224,16 +218,33 @@ public void handleEvent(Event event) {
 			return;
 		}
 		switch (event.getTopic()) {
+		case EdgeEventConstants.TOPIC_CYCLE_EXECUTE_WRITE:
```
sfeilmeier, line 222: Why not call this AFTER_PROCESS_IMAGE?

chpoulter: As it is quite a long time ago, I am not really sure anymore. I have another Controller that writes on MAXIMUM_ALLOWED_POWER_TO_DISTRIBUTE in AFTER_CONTROLLERS and when I remember that value got lost or I had some trouble setting/overriding it from the UI.

Rereading the cycle concept makes me now thinking that AFTER_PROCESS_IMAGE would be a better place, as MAXIMUM_ALLOWED_POWER_TO_DISTRIBUTE is actually an input for EvcsClusterPeakShaving. 

If you suggest that too, I would change it and have a look, how my system behaves.

By further thinking of it: why would I need to call _setMaximumAllowedPowerToDistribute by my own? Wouldn't do nextProcessImage in the channel do the job for me anyway when run in Switch Process Image?

chpoulter: After further investigation, I found that I am setting MAXIMUM_ALLOWED_POWER_TO_DISTRIBUTE by using setNextWriteValue. This value needs to be moved to the READ side in TOPIC_CYCLE_EXECUTE_WRITE otherwise IntegerWriteChannel.value() does not return anything. I guess this is not the correct usage? ... I should be using setNextValue to set a new value from another controller and the WRITE side stays empty?

sfeilmeier: Honestly I would have to dig into the details here, to properly answer. My focus with regard to e-mobility is on the newer EVSE architecture.

chpoulter: should I switch to IntegerReadChannel as the value of this channel is not written to any device and is just used to exchange a value between two components?

sfeilmeier: Yes

chpoulter: I have change it to IntegerReadChannel and removed the logic in TOPIC_CYCLE_EXECUTE_WRITE.

PR #3085, opened 2025-04-07
Implements VirtualMeter on PvInverterCluster

chpoulter:
Added VirtualMeter interface to PvInverterCluster. This allows to configure, if values are added to the sum. Otherwise you end up with doubled values, as the sum will take the cluster as a normal pvInverter.

Made MeterType on the way configurable ... maybe useful for something.

### DIFF
--- a/io.openems.edge.pvinverter.cluster/src/io/openems/edge/pvinverter/cluster/Config.java
+++ b/io.openems.edge.pvinverter.cluster/src/io/openems/edge/pvinverter/cluster/Config.java
@@ -3,6 +3,8 @@ package io.openems.edge.pvinverter.cluster;
 import org.osgi.service.metatype.annotations.AttributeDefinition;
 import org.osgi.service.metatype.annotations.ObjectClassDefinition;
 
+import io.openems.common.types.MeterType;
+
 @ObjectClassDefinition(//
 		name = "PV-Inverter Cluster", //
 		description = "Combines several PV-Inverters to one.")
@@ -16,6 +18,12 @@ import org.osgi.service.metatype.annotations.ObjectClassDefinition;
 
 	@AttributeDefinition(name = "Is enabled?", description = "Is this Component enabled?")
 	boolean enabled() default true;
+
+	@AttributeDefinition(name = "Meter-Type", description = "Production meter (=default)")
+	MeterType meterType() default MeterType.PRODUCTION;
+
+	@AttributeDefinition(name = "Add to Sum?", description = "Should the data of this cluster be added to the Sum?")
+	boolean addToSum() default true;
 
 	@AttributeDefinition(name = "PV-Inverter-IDs", description = "IDs of PvInverter devices.")
 	String[] pvInverter_ids();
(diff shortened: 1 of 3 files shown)


io.openems.edge.pvinverter.cluster/src/io/openems/edge/pvinverter/cluster/Config.java
```diff
@@ -17,6 +19,12 @@
 	@AttributeDefinition(name = "Is enabled?", description = "Is this Component enabled?")
 	boolean enabled() default true;
 
+	@AttributeDefinition(name = "Meter-Type", description = "Production meter (=default)")
+	MeterType meterType() default MeterType.PRODUCTION;
```
sfeilmeier, line 23: Could this ever be anything else than PRODUCTION? If we don't set it here, we might have to add it in UI `isProducer()`, though: https://github.com/OpenEMS/openems/blob/develop/ui/src/app/shared/components/edge/edgeconfig.ts#L582

io.openems.edge.pvinverter.cluster/src/io/openems/edge/pvinverter/cluster/Config.java
```diff
@@ -17,6 +19,12 @@
...
+
+	@AttributeDefinition(name = "Add to Sum?", description = "Should the data of this cluster be added to the Sum?")
+	boolean addToSum() default true;
```
sfeilmeier: For Energy Storage Systems this is identified via `MetaEss` Interface. We should establish similar approaches for the same logic throughout OpenEMS, don't you think? Maybe `addToSum()` is in fact the most flexible approach - see https://github.com/OpenEMS/openems/pull/3238. I am open for discussion.

chpoulter: I looks like the VirtualMeter interface was just created for this use case. I would suggest to implement it in SymmetricEss (and everywhere else needed) and test for it before the switch once bailing out if needed. I'm not really sure about the other PR. As the result value is hardcoded in the interface to true, there is nothing won for existing meters. Of course you could overwrite it and return false in an own meter, but then you could also just implement existing VirtualMeter interface and no change of SumImpl is needed. But maybe I did not get the idea.

EDIT: Maybe the name should be changed or a new interface added, as it seems to refer to the MeterVirtualAdd and MeterVirtualSubstract.

io.openems.edge.pvinverter.cluster/src/io/openems/edge/pvinverter/cluster/Config.java
chpoulter: okay true ... I guess in most cases this will be PRODUCTION. I think there is PRODUCTION returned as default in ManagedSymmetricPvInverter interface anyway, so we could drop this completely. To be honest, I cannot remember anymore, why I added it.

chpoulter:
I did not see further discussion, so I just added the interface and adopted sum calculation. Was this the solution, which you had in mind?

PR #3086, opened 2025-04-07
Max active and reactive power on ManagedSymmetricPvInverter

chpoulter:
Adds two channels on ManagedSymmetricPvInverter: MAX_ACTIVE_POWER and MAX_REACTIVE_POWER. I also implemented maintaining these values on PvInverterCluster and SunSpec inverters.

Note: there is already a channel MAX_APPARENT_POWER on ManagedSymmetricPvInverter, which seems to be used for max active power. I swapped its usage to MAX_ACTIVE_POWER for distribution the pv limit in PvInverterCluster and also adopted the mapping in SunSpec (W_RTG -> V_A_RTG) ... unfortunately my SolarEdge does not support S120 or S702 and uses some other registers for all three MAX ACTIVE/REACTIVE/APPARENT POWER but maybe there is a background for the current implementation?


sfeilmeier (approved):
Thank you!

PR #3091, opened 2025-04-08
Update Documentation for Log Extraction Command

Sn0w3y:
## Overview

This PR updates the documented systemd journal extraction command in the `RunOptimizerFromLogApp` to make it more precise and aligned with current usage.

## Changes

- **Enhanced Log Filtering**: Added a `sed` command to filter log lines so that output always starts with `OPTIMIZER`.

### Before

```bash
journalctl -lu openems --since="20 minutes ago" | grep OPTIMIZER
```

### After

```bash
journalctl -lu openems --since="10 minutes ago" | grep OPTIMIZER | sed 's/.*OPTIMIZER/OPTIMIZER/'
```

## Motivation

The change:
- Ensures that the output is cleaner and focused by removing unnecessary log prefixes, making it easier to use the log data for simulation or debugging.

## Testing

- Verified that the new command returns only the relevant log entries starting with `OPTIMIZER`.
- Confirmed that the updated time window captures the intended log entries during recent activity.

## Conclusion

This is a minor documentation change that enhances clarity and usability for developers working with optimizer log data.

### DIFF
--- a/io.openems.edge.energy/test/io/openems/edge/energy/optimizer/app/RunOptimizerFromLogApp.java
+++ b/io.openems.edge.energy/test/io/openems/edge/energy/optimizer/app/RunOptimizerFromLogApp.java
@@ -13,7 +13,7 @@ import static io.openems.edge.energy.optimizer.app.AppUtils.simulateFromLog;
  * 
  * <p>
  * <code>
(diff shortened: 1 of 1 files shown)


sfeilmeier:
Ok... if it makes your live easier :-) Should I review + merge?

sfeilmeier (approved):

PR #3099, opened 2025-04-16
Adds throttling modbus tasks and config parameters for finetuning

chpoulter:
After adding a second device to my modbus I noticed some read timeouts. I'm using a ethernet-to-modbus interface with four modbus ports. Running the devices on separate ports (only one device per bus) worked fine. Also combining them on one bus work generally, but there occured read timeouts now and then. More worse over night the device got marked as defective, I guess because of too much timeouts in a row.

Debugging that issues I found that OpenEMS can run modbus reads and writes (tasks) without any delay on the modbus-thread. So in some cases a read followed just 1 ms after a successful write. I'm not sure, if one of the devices or the interface cannot handle that, but after adding some code to ensure that there was a fixed delay of 25ms between tasks, all the timeouts were gone. 

So I made this little patch, which adds two configuration values to a modbus device:
- cycle_delay (was hardcoded to 0, not sure for what this is really needed)
- minSleepTime (minimum delay between two tasks, eg. a read or write)

This now allows to fine tune modbus access, if someone has trouble with timeouts. Setting both values to zero will make OpenEMS to work the old way.

On the way I fixed wrong name in FC2ReadInputsTask.


chpoulter:
Hi ... is this patch not of interest?

timo-schlegel:
There should be a planned delay between write and read tasks in the Modbus implementation. This has been fixed in [#3547](https://github.com/OpenEMS/openems/pull/3547). Please verify whether this fixes your issue.

chpoulter:
I just removed my patch and I cannot see a configuration setting that allows to set a minimum delay between any two modbus actions on a "Bridge Modbus/TCP". Having a very short look in the new code, I did not spot something like it. Anyhow I left the edge running sometime without my patch but now get timeouts:

`edge-1  | 2026-02-07T00:21:13,397 [modbus2 ] ERROR [e.modbus.api.task.AbstractTask]   Execute failed FC3ReadHoldingRegisters [meter1;unitid=1;priority=HIGH;ref=1024/0x400;length=58] ModbusIOException: Executing transaction 04 30 00 00 00 06 01 03 04 00 00 3A failed (tried 1 times) Socket timeout reading response - Read timed out`

Adding back my patch prevents those timeouts, so unfortunately, it looks like #3547 does not fix my problem.

timo-schlegel:
Do you require a delay only between write and read operations, or between all tasks?
The delay between write and read should already be handled by the mentioned PR. A delay between all tasks would require a separate PR. In that case, you would need to ensure that the planned delay calculation still works correctly.

If you only require a delay between write and read operations, which should be fixed already, could you please provide Modbus debug logs with the READS_AND_WRITES_DURATION log level, using the mentioned PR and your fix?

chpoulter:
Yes - I think it needs a delay between any bus access.

`edge-1  | 2026-02-07T19:38:03,477 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC1ReadCoils [io3;unitid=10;priority=LOW;ref=0/0x0;length=8;response=00000000]
edge-1  | 2026-02-07T19:38:03,500 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC2ReadCoils [io3;unitid=10;priority=HIGH;ref=0/0x0;length=8;response=00000000]
edge-1  | 2026-02-07T19:38:06,507 [modbus2 ] ERROR [e.modbus.api.task.AbstractTask]   Execute failed FC3ReadHoldingRegisters [meter1;unitid=1;priority=HIGH;ref=1024/0x400;length=58] ModbusIOException: Executing transaction 00 04 00 00 00 06 01 03 04 00 00 3A failed (tried 1 times) Socket timeout reading response - Read timed out
edge-1  | 2026-02-07T19:38:06,605 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC3ReadHoldingRegisters [meter1;unitid=1;priority=HIGH;ref=1024/0x400;length=58;response=0003 8bbc 0003 90b7 0003 89de 0005 c42b 0006 2040 0006 6c6f 0003 8cc5 0006 1af3 0000 109a 0000 0b0e 0000 049c 0000 0000 0000 09a6 0000 0350 0000 0230 0000 00e1 0000 0661 0000 0398 0000 0256 0000 00f6 0000 06e5 ffff fe9a ffff ff2d ffff ff9d ffff fd63 01f4 01f4 01f3 01f4 0399 03a7 0392 039d]`

Here you see two accesses in 3ms ... maybe this works on other devices, but I found I need an idle delay of about 25ms between each modbus frame to prevent timeouts. I'm not a modbus expert, but modbus messages on RTU require to be separated by an idle periods of 3,5 chars. With 9600baud this is slightly more than 4ms. Maybe my TCP->RTU device is not smart enough to handle bursts and just forwards frames to the bus flooding it.

chpoulter:
Without my patch
`
edge-1  | 2026-02-07T20:16:41,762 [artLevel] INFO  [onent.AbstractOpenemsComponent] [modbus2] Activate Bridge.Modbus.Tcp
edge-1  | modbus2[CycleDelay:5520 ms]
edge-1  | modbus2[CycleDelay:0 ms]
edge-1  | modbus2[CycleDelay:0 ms]
edge-1  | modbus2[CycleDelay:3435 ms]
edge-1  | modbus2[CycleDelay:0 ms]
edge-1  | 2026-02-07T20:16:58,261 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC1ReadCoils [io3;unitid=10;priority=LOW;ref=0/0x0;length=8] Elapsed [21ms]
edge-1  | 2026-02-07T20:16:58,283 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC2ReadCoils [io3;unitid=10;priority=HIGH;ref=0/0x0;length=8] Elapsed [19ms]
edge-1  | modbus2[CycleDelay:0 ms]
edge-1  | 2026-02-07T20:17:01,287 [modbus2 ] ERROR [e.modbus.api.task.AbstractTask]   Execute failed FC3ReadHoldingRegisters [meter1;unitid=1;priority=HIGH;ref=1024/0x400;length=58] Elapsed [3001ms] ModbusIOException: Executing transaction 00 04 00 00 00 06 01 03 04 00 00 3A failed (tried 1 times) Socket timeout reading response - Read timed out
edge-1  | 2026-02-07T20:17:01,386 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC3ReadHoldingRegisters [meter1;unitid=1;priority=HIGH;ref=1024/0x400;length=58] Elapsed [94ms]
edge-1  | modbus2[CycleDelay:2218 ms]
edge-1  | 2026-02-07T20:17:04,592 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC3ReadHoldingRegisters [meter1;unitid=1;priority=LOW;ref=270/0x10e;length=52] Elapsed [74ms]
edge-1  | modbus2[CycleDelay:0 ms]
edge-1  | modbus2[CycleDelay:0 ms]
`

With my patch:
`edge-1  | modbus2[CycleDelay:3387 ms]
edge-1  | modbus2[CycleDelay:0 ms]
edge-1  | modbus2[CycleDelay:2101 ms]
edge-1  | 2026-02-07T20:24:30,397 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC1ReadCoils [io3;unitid=10;priority=LOW;ref=0/0x0;length=8] Elapsed [20ms]
edge-1  | 2026-02-07T20:24:30,454 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC2ReadInputsTask [io3;unitid=10;priority=HIGH;ref=0/0x0;length=8] Elapsed [20ms]
edge-1  | 2026-02-07T20:24:30,586 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC3ReadHoldingRegisters [meter1;unitid=1;priority=HIGH;ref=1024/0x400;length=58] Elapsed [94ms]
edge-1  | modbus2[CycleDelay:0 ms]
edge-1  | modbus2[CycleDelay:1351 ms]
edge-1  | 2026-02-07T20:24:34,697 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC3ReadHoldingRegisters [meter1;unitid=1;priority=LOW;ref=270/0x10e;length=52] Elapsed [75ms]
edge-1  | 2026-02-07T20:24:34,753 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC2ReadInputsTask [io3;unitid=10;priority=HIGH;ref=0/0x0;length=8] Elapsed [19ms]
edge-1  | 2026-02-07T20:24:34,884 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC3ReadHoldingRegisters [meter1;unitid=1;priority=HIGH;ref=1024/0x400;length=58] Elapsed [94ms]
edge-1  | modbus2[CycleDelay:1512 ms]
edge-1  | 2026-02-07T20:24:37,361 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC3ReadHoldingRegisters [meter1;unitid=1;priority=LOW;ref=1280/0x500;length=52] Elapsed [77ms]
edge-1  | 2026-02-07T20:24:37,418 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC2ReadInputsTask [io3;unitid=10;priority=HIGH;ref=0/0x0;length=8] Elapsed [19ms]
edge-1  | 2026-02-07T20:24:37,568 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC3ReadHoldingRegisters [meter1;unitid=1;priority=HIGH;ref=1024/0x400;length=58] Elapsed [113ms]
edge-1  | modbus2[CycleDelay:1673 ms]
edge-1  | 2026-02-07T20:24:40,029 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC3ReadHoldingRegisters [meter1;unitid=1;priority=LOW;ref=1380/0x564;length=52] Elapsed [78ms]
edge-1  | 2026-02-07T20:24:40,085 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC2ReadInputsTask [io3;unitid=10;priority=HIGH;ref=0/0x0;length=8] Elapsed [20ms]
edge-1  | 2026-02-07T20:24:40,215 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC3ReadHoldingRegisters [meter1;unitid=1;priority=HIGH;ref=1024/0x400;length=58] Elapsed [94ms]
edge-1  | modbus2[CycleDelay:1733 ms]
edge-1  | 2026-02-07T20:24:42,587 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC3ReadHoldingRegisters [meter1;unitid=1;priority=LOW;ref=1480/0x5c8;length=52] Elapsed [82ms]
edge-1  | 2026-02-07T20:24:42,643 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC2ReadInputsTask [io3;unitid=10;priority=HIGH;ref=0/0x0;length=8] Elapsed [19ms]
edge-1  | 2026-02-07T20:24:42,773 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC3ReadHoldingRegisters [meter1;unitid=1;priority=HIGH;ref=1024/0x400;length=58] Elapsed [94ms]
edge-1  | modbus2[CycleDelay:1794 ms]
edge-1  | 2026-02-07T20:24:45,087 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC1ReadCoils [io3;unitid=10;priority=LOW;ref=0/0x0;length=8] Elapsed [19ms]
edge-1  | 2026-02-07T20:24:45,143 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC2ReadInputsTask [io3;unitid=10;priority=HIGH;ref=0/0x0;length=8] Elapsed [19ms]
edge-1  | 2026-02-07T20:24:45,273 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC3ReadHoldingRegisters [meter1;unitid=1;priority=HIGH;ref=1024/0x400;length=58] Elapsed [93ms]
edge-1  | modbus2[CycleDelay:1855 ms]
edge-1  | 2026-02-07T20:24:47,731 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC3ReadHoldingRegisters [meter1;unitid=1;priority=LOW;ref=270/0x10e;length=52] Elapsed [74ms]
edge-1  | 2026-02-07T20:24:47,786 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC2ReadInputsTask [io3;unitid=10;priority=HIGH;ref=0/0x0;length=8] Elapsed [19ms]
edge-1  | 2026-02-07T20:24:47,916 [modbus2 ] INFO  [e.modbus.api.task.AbstractTask]   Execute FC3ReadHoldingRegisters [meter1;unitid=1;priority=HIGH;ref=1024/0x400;length=58] Elapsed [94ms]`

Sorry ... formating seems not to work good ...

timo-schlegel:
Okay, in this case your patch seems to be required, as this is not covered by the mentioned PR.

Could you please test that, with your patch, the write tasks are still executed as early as possible (directly after the EXECUTE_WRITE event) and that the read tasks are still executed as late as possible (i.e., just before the BEFORE_PROCESS_IMAGE event), as described in the Modbus bridge [readme.adoc](https://github.com/OpenEMS/openems/tree/develop/io.openems.edge.bridge.modbus)?

io.openems.common/src/io/openems/common/worker/AbstractWorker.java
```diff
@@ -115,6 +115,17 @@ public void deactivate() {
 	 */
 	protected abstract int getCycleTime();
 
+	/**
+	 * Returns the minimum sleep time between two cycles. This avoids tasks
+	 * to run immediately after each other, eg. forces a break between two
+	 * modbus reads.
+	 *
+	 * @return sleep time in ms
+	 */
```
da-Kai, line 123: As this worker is Generic, i would not mention Modbus specifically.

io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/ConfigTcp.java
```diff
@@ -33,5 +33,11 @@
 	@AttributeDefinition(name = "Invalidate elements after how many read Errors?", description = "Increase this value if modbus read errors happen frequently.")
 	int invalidateElementsAfterReadErrors() default 1;
 
+	@AttributeDefinition(name = "Duration of a cycle on the modbus", description = "Defines how long a cylce on the modbus thread should take in ms. 0 means next task executes immediately. A value below 0 requires external triggering to run the next task.")
+	int cycleTime() default 0;
+
```
da-Kai: if you do not to set cycleTime i would leave it hardcoded at 0

da-Kai (changes requested):
Hi @chpoulter, thanks for this PR 👍

I added some inline review comments. Beyond those, I have two additional requests

Please revert the version changes in:
- `OpenemsConstants.java`
- `ui/package.json` 
- `ui/package-lock.json`
- `changelog.constants.ts`

and To keep the "immediate" behavior explicit please add 
```Java
	@Override
	protected final int getMinSleepTime() {
		return 0;
	}
```
to `io.openems.common.worker.AbstractImmediateWorker`

chpoulter:
> Okay, in this case your patch seems to be required, as this is not covered by the mentioned PR.
> 
> Could you please test that, with your patch, the write tasks are still executed as early as possible (directly after the EXECUTE_WRITE event) and that the read tasks are still executed as late as possible (i.e., just before the BEFORE_PROCESS_IMAGE event), as described in the Modbus bridge [readme.adoc](https://github.com/OpenEMS/openems/tree/develop/io.openems.edge.bridge.modbus)?

I had a look into the CycleTasksManager. From my understanding it filters available Tasks according to the current OpenEMS cycle state and adds a delay after all WRITES before proceeding with the READS. The delay time is based on information from the last cycles. Honestly this code is quite complex and I am not sure, if I understood it fully.

With my change a forced delay between any two Tasks on the Worker Thread is added. If you set minSleepTime to 0 there is no change to the old behavior, so CycleTasksManager should work the same. If you set a positive value, the worker thread will work "slower", the forever() is not called as fast as possible anymore, cuz there are delays between the calls. From my understanding this could be imagined as throttling the worker thread, what basically could happen for other reasons (eg. CPU overload) at any time too. As the CycleTasksManager "learns" it's delay from prior runs, it should take the slower worker automatically into account and adjusts it's own delay, like it has different run times on different devices anyway. Of course - if you set a high minSleepTime or have a too much Tasks at some point CycleTasksManager cannot reduce it's delay any further. But this is a general problem, if the OpenEMS cycle is to short for processing all devices. Also as we have now fewer calls to forever() it is harder to hit the optimal moment for executing READS or WRITES - so maybe they are a bit to late, eg. when onExecuteWrite is called, there could be a delay up to minSleepTime before ModbusWorker gets the chance to run the WRITES. But as minSleepTime is addressing a Modbus timing problem, it should be choosen with a reasonable value that would be enforced otherwise at some other point in the architecture (like the TCP->MODBUS hardware device should actually implement it).

If you set minSleepTime to 0 (which is the default) there is no change in worker behavior and I see no risk for anybody. For me I've been running with minSleepTime set to 35ms since I opened this PR and did not notice any trouble (I'm currently running 3 Modbus with 2-3 devices on them).

chpoulter:
> Hi @chpoulter, thanks for this PR 👍
> 
> I added some inline review comments. Beyond those, I have two additional requests
> 
> Please revert the version changes in:
> 
>     * `OpenemsConstants.java`
> 
>     * `ui/package.json`
> 
>     * `ui/package-lock.json`
> 
>     * `changelog.constants.ts`
> 
> 
> and To keep the "immediate" behavior explicit please add
> 
> ```java
> 	@Override
> 	protected final int getMinSleepTime() {
> 		return 0;
> 	}
> ```
> 
> to `io.openems.common.worker.AbstractImmediateWorker`

I'm not sure where this changes came from, maybe cuz I used main and not develop. I merged develop back into the branch and hope the patch looks better for you. I also incorporated your suggestions.

io.openems.common/src/io/openems/common/worker/AbstractWorker.java
chpoulter: changed

timo-schlegel:
@chpoulter
Thanks for getting back to me on my comment. I’ve taken a closer look at the topic and would like to provide some additional context and clarification to help ensure we’re all aligned.

There is a stopwatch that measures if additional delay is possible (it starts with a zero delay on startup).
The possible delay will be calculated at [this line](https://github.com/OpenEMS/openems/blob/b3df4f93b66b2a3a8b838fd8e299efc4b6ec2f04/io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/api/worker/internal/WaitDelayHandler.java#L97-L98) (if it can be increased) or at [this line](https://github.com/OpenEMS/openems/blob/b3df4f93b66b2a3a8b838fd8e299efc4b6ec2f04/io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/api/worker/internal/WaitDelayHandler.java#L107-L117) (if it needs to be decreased).
The stopwatch will be restarted at [this line](https://github.com/OpenEMS/openems/blob/b3df4f93b66b2a3a8b838fd8e299efc4b6ec2f04/io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/api/worker/internal/WaitDelayHandler.java#L155).

Thus, the possible wait time will be the last delay time plus the measured time between `FINISHED` and `ON_BEFORE_PROCESS_IMAGE` event.

The `WaitTask` will run in the `WAIT_BEFORE_READ` state (see [this reference](https://github.com/OpenEMS/openems/blob/b3df4f93b66b2a3a8b838fd8e299efc4b6ec2f04/io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/api/worker/internal/CycleTasksManager.java#L151-L153)).

Therefore, as long as there are no changes in timing within `WAIT_BEFORE_READ` and between `FINISHED` and `ON_BEFORE_PROCESS_IMAGE`, the timing should not be disrupted.

If I understand correctly, you're introducing a sleep time at [this line](https://github.com/OpenEMS/openems/blob/573baf7ca2420a1f016c9bfa66f4cd7a17e46edb/io.openems.common/src/io/openems/common/worker/AbstractWorker.java#L151-L153), before calling `forever()` at [this line](https://github.com/OpenEMS/openems/blob/573baf7ca2420a1f016c9bfa66f4cd7a17e46edb/io.openems.common/src/io/openems/common/worker/AbstractWorker.java#L175) which will then call `forever()` at [this line](https://github.com/OpenEMS/openems/blob/573baf7ca2420a1f016c9bfa66f4cd7a17e46edb/io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/api/worker/ModbusWorker.java#L75), which will execute the code at [this line](https://github.com/OpenEMS/openems/blob/b3df4f93b66b2a3a8b838fd8e299efc4b6ec2f04/io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/api/worker/internal/CycleTasksManager.java#L126).

This means you are not changing the timing within `WAIT_BEFORE_READ`. Additionally, the timing in the `FINISHED` task (where we wait for `BEFORE_PROCESS_IMAGE`) should not be affected.

As a result, the following timing should still hold:

* Execute Write-Tasks as early as possible (directly after the `EXECUTE_WRITE` event).
* Execute Read-Tasks  as late as possible to have values available exactly when they are needed (i.e. just before the `BEFORE_PROCESS_IMAGE` event).

To verify the theory, you could add temporary logging for the `EXECUTE_WRITE` and `BEFORE_PROCESS_IMAGE` events (e.g., in an ESS implementation in the `handleEvent()` method). You could also set the Modbus log verbosity to `READS_AND_WRITES` to see exactly when reads and writes are executed, and when the `EXECUTE_WRITE` and `BEFORE_PROCESS_IMAGE` events are triggered.

---

I hope that helps! I understand that your change will only affect behavior when configured, but based on the analysis above, I believe the essential timing within OpenEMS should still work as expected in those cases. I also think it makes sense that the timing will be correct in your scenario as well.

Since it's not trivial, timing discrepancies might not be noticed immediately when something goes wrong. Consider the timing of my PR and when the Modbus rework was released — there was quite a long period where the timing wasn't quite optimal.

da-Kai (changes requested):
@chpoulter 
I would still suggest the following changes:
1. the `ModbusWorker` should extend `AbstractWorker` instead of `AbstractImmediateWorker` as it states that it never waits
2. `AbstractWorker` should define `protected abstract int getMinSleepTime();` and let each worker implement it.
3. `AbstractImmediateWorker` should override `getMinSleepTime()` and explicitly return 0.

Are there reasons against any of these changes?

chpoulter:
@da-Kai I actually did not want to change all the usages of AbstractWorker ... but I am fine with your suggestion.

chpoulter:
@timo-schlegel I think basically you are right. Every call to forever() will be delayed, but because of the structure of getNextTask() and reacting on external events not every step of the state machine is run in a separate call of forever(). IMHO it should look like this:

INITIAL_WAIT (triggered by onBeforeProcessImage)
  sleep
  waitMutexTask
WRITE (triggered by onExecuteWrite)
  sleep
  writeTask
  sleep
  ...
  writeTask
  sleep
WAIT_BEFORE_READ (no write tasks anymore)
  waitDelayTask
READ (triggerd by onWaitDelayTaskFinished)
  sleep
  readTask
  sleep
  ...
  readTask
  sleep
FINISHED (no read tasks anymore)
   waitMutexTask

Sleeps are mainly in the WRITE and READ state. As forever() is called after waiting() the first write / read is slightly delayed. waitDelayTask takes time from start of FINISHED until onBeforeProcessImage, which has no sleep in it. The sleeps mainly extend WRITE and READ state. So if there is time left at the end, it can be safely waited in waitDelayTask on the next cycle.

da-Kai (approved):
Your latest merge from main reintroduced some version updates and changes to `package-lock.json` again.
Aside from that, looks good to me 👍

chpoulter:
I have switched to another modbus gateway that works out of the box. I won't maintain this code anymore. But if you like it anyway, feel free to take over it.

Howaner:
I debugged the modbus bridge implementation in the last days and discovered the same problem. OpenEMS is sending requests way too fast after another response/request. Modbus requires a delay of at least 3.5 characters between requests.

Normally, j2mod should be responsible to handle this. It's already implemented there in ModbusSerialTransport: https://github.com/steveohara/j2mod/blob/master/src/main/java/com/ghgande/j2mod/modbus/io/ModbusSerialTransaction.java#L131C85-L131C109

But this does not work because every request is currently using a new ModbusSerialTransport object and therefore the lastTransactionTimestamp is always 0.

Howaner:
@chpoulter  Are you sure that a 25ms delay is required? With a baudrate of 19200, a delay of 1.6ms should be already enough to reach the 3.5 character delay period.

chpoulter:
Hello @Howaner – I have just choose a rather high value to avoid any kind of congestion. I only have two devices on the bus, so it's not really busy.

I was not aware that the lib handles it too - then I might have fixed the lib handling instead  of implementing it myself. But now I do not need it anymore. I guess, it would be the best to fix the handling of the lib in OpenEMS. Feel free 😉 ...

Check ghcr.io/tma/mbproxy - it is a tcp modproxy that you can run in docker. It allows throttling and caching - and also logs everything. Can be helpful.

PR #3122, opened 2025-05-11
Improve Docker container

da-Kai:
* Improve Container creation; Reduce size by moving more work to the build container.
* Use Temurin JDK
* Add JVM settings
* Edge: Implement OperatingSystemDocker
* Add more core components to default setup


sfeilmeier (approved):
Thanks!

PR #3125, opened 2025-05-15
LSTM predictor with dropout regularization

Sn0w3y:
- Implement dropout regularization in LSTM predictor for better model generalization
- Add scaling factors for dropout to maintain output magnitude during training


sfeilmeier (approved):

sfeilmeier:
We are currently not working actively on the LSTM, so no, cannot test locally at the moment.

PR #3127, opened 2025-05-15
Update docker-compose.yml: removed "version" attribute

sjjh:
As docker was complaining:
# docker compose up -d
WARN[0000] docker-compose.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion

cf: https://docs.docker.com/reference/compose-file/version-and-name/#version-top-level-element-obsolete

### DIFF
--- a/tools/docker/backend/docker-compose.yml
+++ b/tools/docker/backend/docker-compose.yml
@@ -1,5 +1,3 @@
-version: "3"
-
 services:
   openems_backend:
     image: openems/backend:latest


da-Kai:
@sjjh Thank you!
Could you update the edge docker-compose file as well? Thanks!
https://github.com/OpenEMS/openems/blob/develop/tools/docker/edge/docker-compose.yml

sfeilmeier (approved):

PR #3157, opened 2025-06-02
feat: add edge-list to siderbar in UI

miettal:


lukasrgr:
@miettal Hi, thanks for your contribution. Could you write down, what motivated you to do this change, and what use cases this will suit, 

In my opinion, this would make most sense, to show this only for `owner`s

miettal:
@lukasrgr This is a usecase for user who manages a lot of edges, like administrator.
Current implementation, user needs to go overview for switching to other edge.
This PR provides siwitching interface from sidebar to go other edge directly.

miettal:
> In my opinion, this would make most sense, to show this only for owners

I think so, I will revise.

lukasrgr:
@miettal i think we should make this optional, if it works for your use case, 

@sfeilmeier what do you think, is this something that should be merged and shown to everyone?

miettal:
both is ok, I will follow.

if optionalization is needed, I wii add this commit.
https://github.com/girasolenergy/openems-fork/commit/dfed40f287ecb1d8f69b1046c0f7c167092fb37f

lukasrgr (approved):
Thx for making it optional, great work 🚀

miettal:
// I added optionalization change to this PR.

miettal:
rebased

PR #3182, opened 2025-06-11
feat: change reconnection timer implementation unix time -> cpu clock time

miettal:
Currently reconnection timer use unix timestamp and it feed from system. But this time is not monotonic increasing, So there is possibility of time go away to far future, in this case the timer doesn't work correctly. Especially, In case of use RTC and edge is in booting time, If System recoginize wrong time, Sometime edge set wrong timer period and wait forever. // Actually We face this problem recently.

This change provice to use monotonic timer it based on cpu clock to resolve ↑ problem.
Cpu clock has no external dependencies and It suitable reconnection timer use.

Co-author: @cvabc, claude code, chatgpt codex, github copilot

### DIFF
--- a/io.openems.common/src/io/openems/common/websocket/ClientReconnectorWorker.java
+++ b/io.openems.common/src/io/openems/common/websocket/ClientReconnectorWorker.java
@@ -38,7 +38,7 @@ public class ClientReconnectorWorker extends AbstractWorker {
 		this.parent = parent;
 		this.config = config;
 		this.minWaitSecondsBetweenRetries = new Random().nextInt(config.maxWaitSeconds()) + config.minWaitSeconds();
-		this.lastTry = System.nanoTime() - this.minWaitSecondsBetweenRetries * 1_000_000_000L;
+		this.lastTry = 0;
 	}
 
 	public ClientReconnectorWorker(AbstractWebsocketClient<?> parent) {


io.openems.common/src/io/openems/common/websocket/ClientReconnectorWorker.java
```diff
@@ -33,13 +31,14 @@ public record Config(int connectTimeoutSeconds, int maxWaitSeconds, int minWaitS
...
 	private final long minWaitSecondsBetweenRetries;
-	private Instant lastTry = Instant.MIN;
+	private long lastTry;
 	private String debugLog = null;
 
 	public ClientReconnectorWorker(AbstractWebsocketClient<?> parent, Config config) {
 		this.parent = parent;
 		this.config = config;
 		this.minWaitSecondsBetweenRetries = new Random().nextInt(config.maxWaitSeconds()) + config.minWaitSeconds();
+		this.lastTry = System.nanoTime() - this.minWaitSecondsBetweenRetries * 1_000_000_000L;
```
sfeilmeier: `1_000_000_000L` seems arbitrary. Can we not set `lastTry` to zero instead?

sfeilmeier (commented):
Looks good. Sorry, I did not find time to review earlier. Now it has to be rebased to current develop.

PR #3183, opened 2025-06-11
feat: add listen ip parameter to components that have socket and set 127.0.0.1 as default

miettal:
Current implementation has no parameter of bind ip address and So openems listen 0.0.0.0 (all interface of the computer). This will be problem in usecase of using in global network that can be acceced by anyone. Because some component doesn't provide authentication/authorization/encryption, So if these component is used without IP Firewall in global network, vulnerable endpoint will be exposed.

Therefore this PR introduce bind ip address parameter to these components and provide option of lisning interface.

Today, HTTP/TLS become very complex protocol, So implementing this by application-side itself is very very difficut. Almost case, HTTP/TLS server in global network is provided by combination application and reverse proxy server(like apache, nginx). This PR is mainly targeting lisning localhost in application and lisning 0.0.0.0 in reverse proxy use.

Co-Author: @cvabc, chagpt codex

### DIFF
--- a/io.openems.backend.b2brest/src/io/openems/backend/b2brest/Backend2BackendRest.java
+++ b/io.openems.backend.b2brest/src/io/openems/backend/b2brest/Backend2BackendRest.java
@@ -1,5 +1,6 @@
 package io.openems.backend.b2brest;
 
+import java.net.InetSocketAddress;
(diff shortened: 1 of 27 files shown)


miettal:
rebased

miettal:
rebased

PR #3197, opened 2025-06-19
Fix JSCalendar without recurrence rule

tsicking:
Previously, a `NullPointerException` was thrown when parsing a `JSCalendar`-Entry without a recurrence rule. This is fixed here, and JUnit tests are added.

### DIFF
--- a/io.openems.common/src/io/openems/common/jscalendar/JSCalendar.java
+++ b/io.openems.common/src/io/openems/common/jscalendar/JSCalendar.java
@@ -27,6 +27,7 @@ import java.time.LocalTime;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
+import java.util.Collections;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.UUID;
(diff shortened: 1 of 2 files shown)


clehne (approved):

sfeilmeier:
@michaelgrill Can you please validate? Shouldn't the `mapIfPresent` variant also work?

tsicking:
> @michaelgrill Can you please validate? Shouldn't the `mapIfPresent` variant also work?

It doesn't work because  `mapIfPresent` returns `null`, and therefore `forEach` throws a `NullPointerException`. Here is the result of the JUnit test with the old code:

image

michaelgrill:
@sfeilmeier 
@tsicking is right `mapIfPresent` returns null if not included in the json (as mentioned in the javadoc). Use `mapToOptional` or the shortest way would be something like this 
`json.getOptionalList("recurrenceRules", RecurrenceRule.serializer()).orElse(Collections.emptyList())`

tsicking:
> @sfeilmeier @tsicking is right `mapIfPresent` returns null if not included in the json (as mentioned in the javadoc). Use `mapToOptional` or the shortest way would be something like this `json.getOptionalList("recurrenceRules", RecurrenceRule.serializer()).orElse(Collections.emptyList())`

Thanks for your quick response. Your suggestion also works and is slightly more elegant, so I changed it accordingly.

sfeilmeier:
Thank you all! 👍🚀

sfeilmeier (approved):

PR #3199, opened 2025-06-20
Update README.md: removed link to felix, as it is neither accessible nor needed

sjjh:
According to my testing and feedback from the community, the felix web console of the edge is not accessible and is not needed, as OpenEMS is configured via the UI. cf: https://community.openems.io/t/openems-release-2025-6-0-backend-docker-deployment-fails/10186/5


sjjh:
not necessary, with updated pr #3200

PR #3201, opened 2025-06-20
Update docker.adoc: added hint & link to influxdb deployment

sjjh:
link to influxdb setup in backend deployment


Sn0w3y:
IMHO Every User which sets up OpenEMS should read and is advised to read the Getting Started, where one finds exactly this Information...

OpenEMS is a Powerful Tool and if one needs a Backend (mostly Commercial Users) one would definately know or read the Getting Started i guess. Or at least they should..

sjjh:
I added it to the Edge part of the getting started -- so the use case here is that you do _not_ need a backend, but still want to save the data persistent in an InfluxDB.  Thus the link towards the other part of the getting started guide (to not duplicate the contents).

sjjh:
@da-Kai probably you can evaluate best if it's worth merging, otherwise feel free to just close it.

PR #3203, opened 2025-06-21
Update IoShellyPro3Em.java: included remedy for communication error msg

sjjh:
Good error messages should contain information how to solve the issue, cf. https://www.nngroup.com/articles/error-message-guidelines/


Sn0w3y:
As stated in https://github.com/OpenEMS/openems/issues/3205#issuecomment-2993788158 i would not just Inform the User in the Generic "Communication Failed" Channel as this is too Generic. It coul basically be EVERYTHING why it fails - so it woul be better to create a new Channel for this if we get a Status Code 401 from the Shelly Device. See the /shelly Endpoint

sjjh:
obsolete due to #3248

PR #3204, opened 2025-06-21
Update readme.adoc: added limitation regarding UI pwd

sjjh:
https://community.openems.io/t/how-to-debug-communication-error-with-shelly-pro-3em/10229/4

### DIFF
--- a/io.openems.edge.io.shelly/readme.adoc
+++ b/io.openems.edge.io.shelly/readme.adoc
@@ -11,4 +11,6 @@ Compatible with
 - https://www.shelly.com/de/products/shop/shelly-plus-1-pm[Shelly Plus 1PM]
 - https://www.shelly.com/de/products/shop/shelly-pro-3-1[Shelly Pro 3]
 
+_Hint:_ Connecting to devices having a password set in their UI is currently not supported.
+
 https://github.com/OpenEMS/openems/tree/develop/io.openems.edge.io.shelly[Source Code icon:github[]]


sjjh:
@sfeilmeier this is IMHO trivial, would you mind merging it?

michaelgrill (approved):
Thanks for the contribution. 🙂👍

Even if you added the community post in the PR description write a short summary not everyone may want to read thru the whole story.

Some next steps could be:
 - Add a different warning channel if authentication fails, documentation is good but not everyone may read it.
  - And as you mentioned in the community post the possibility to even authenticate

sjjh:
thx @michaelgrill for the review

> *    Add a different warning channel if authentication fails, documentation is good but not everyone may read it.

I belive @Sn0w3y has implemented this already: https://github.com/OpenEMS/openems/pull/3248

> *    And as you mentioned in the community post the possibility to even authenticate

I think that this would be the best out of a users perspective, but as far as I understood @Sn0w3y it's not the best implementation wise.

Sn0w3y:
> thx @michaelgrill for the review
> 
> > *    Add a different warning channel if authentication fails, documentation is good but not everyone may read it.
> 
> I belive @Sn0w3y has implemented this already: https://github.com/OpenEMS/openems/pull/3248
> 
> > *    And as you mentioned in the community post the possibility to even authenticate
> 
> I think that this would be the best out of a users perspective, but as far as I understood @Sn0w3y it's not the best implementation wise.


Well basically it would be just better to Implement the Auth Flow - but in this Case tbh it is easier to just inform the User. So we should stick to the "Info" first imho.

So "implementationwise" we should keep my approach but in the long run we could  Implement the Auth aswell.

Greetings!

sjjh:
@michaelgrill how can we progress here? :)

sjjh:
@michaelgrill any update on this? Or do you maybe have an idea, @da-Kai ? I'd like to get this PR merged (or delete it) to be able to delete my messy fork and create a new one.

PR #3208, opened 2025-06-23
[Edge] EVCS Mennekes: replace handling of phase-rotated wiring

Sn0w3y:
1. Replacing the deprecated mapLongToPhaseRotatedCurrentChannel calls with direct channel mapping using this.getPhaseRotation().channelCurrentL1(), channelCurrentL2(), and channelCurrentL3()
  2. Removing the unused import for mapLongToPhaseRotatedCurrentChannel

  The new approach directly maps the Modbus register values to the appropriate current channels based on the configured phase rotation, which is cleaner and follows the recommended pattern.


sfeilmeier (approved):
Thanks. I did some more cleanup

PR #3196, opened 2025-06-18
[wip]feat: enable adjustable websocket compression

miettal:
From TooTallNate/Java-WebSocket 1.6, the websocket library support specification of compression level.
So this PR introduce compression level adjusable websocket compression function.

But this repo still use 1.5.4 because it has a blocker (compatibility problem, blocking reconnection was not supported). 
https://github.com/OpenEMS/openems/pull/2926
But recently blocking reconnection is supported! This freature will released in 1.6.1, I think.
https://github.com/TooTallNate/Java-WebSocket/pull/1465

Currently this PR is in WIP, because waiting 1.6.1 release.

PR #3212, opened 2025-06-25
Fix implementation for resetWebSocketClient method to allow websocket version bump

Yuval-Roth:
Hello, i am the author of https://github.com/TooTallNate/Java-WebSocket/pull/1465
I'm here to offer you a solution for https://github.com/OpenEMS/openems/pull/2462#issuecomment-1862381632 to allow you to bump the websocket client version from 1.5.4 to 1.6.0 and continue this https://github.com/OpenEMS/openems/pull/3196

As i understand it, the issue you had with closeBlocking() hanging indefinitely was [fixed in version 1.5.5](https://github.com/TooTallNate/Java-WebSocket/issues/1365).
So, the implementation for resetting the websocket client is trivial - just call the private reset method.

This works in 1.6.0 as I have used it myself.

### DIFF
--- a/io.openems.backend.application/BackendApp.bndrun
+++ b/io.openems.backend.application/BackendApp.bndrun
@@ -61,7 +61,7 @@
 	bnd.identity;id='io.openems.backend.uiwebsocket',\
 
 -runbundles: \
-	Java-WebSocket;version='[1.5.4,1.5.5)',\
+	Java-WebSocket;version='[1.6.0,1.6.1)',\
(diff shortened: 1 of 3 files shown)


sfeilmeier (approved):
Awesome! 🚀 Thank you for your active support - and the great library! Java-WebSocket is really a super important core library for OpenEMS, as all the communication between Edge, Backend and UI is transferred via JSON/WebSocket.

PR #3237, opened 2025-07-13
Implementation OpenWB Series2 Meter

sthhs:
This components requests the actual power, voltage and current values from an OpenWB Series 2 with one internal charge point in primary mode, which is commonly used. In this case the wallbox can only be monitored from other systems and not be controlled.
The purpose is considering the current charging power in the total consumption and showing and archiving the retrieved values in OpenEMS. If no more information is needed in OpenEMS a meter component seems to be a suitable solution compared to an EVCS or EVSE component.
The pictures show the corresponding values in the OpenWB GUI and in OpenEMS at nearly the same moment.
Bildschirmfoto 2025-07-13 um 14 46 57
Bildschirmfoto 2025-07-13 um 14 46 54

These channels can be requested from the internal charge point:
Bildschirmfoto 2025-07-13 um 17 22 14

### DIFF
--- a/io.openems.edge.meter.openwb/src/io/openems/edge/meter/openwb/MeterOpenWbImpl.java
+++ b/io.openems.edge.meter.openwb/src/io/openems/edge/meter/openwb/MeterOpenWbImpl.java
@@ -17,6 +17,7 @@ import org.osgi.service.event.EventHandler;
 import org.osgi.service.event.propertytypes.EventTopics;
 import org.osgi.service.metatype.annotations.Designate;
 
+import io.openems.common.channel.AccessMode;
 import io.openems.common.exceptions.InvalidValueException;
 import io.openems.common.exceptions.OpenemsException;
 import io.openems.common.types.MeterType;
@@ -24,6 +25,9 @@ import io.openems.common.utils.InetAddressUtils;
 import io.openems.edge.common.component.AbstractOpenemsComponent;
 import io.openems.edge.common.component.OpenemsComponent;
 import io.openems.edge.common.event.EdgeEventConstants;
+import io.openems.edge.common.modbusslave.ModbusSlave;
+import io.openems.edge.common.modbusslave.ModbusSlaveNatureTable;
+import io.openems.edge.common.modbusslave.ModbusSlaveTable;
 import io.openems.edge.meter.api.ElectricityMeter;
 import io.openems.edge.timedata.api.Timedata;
 import io.openems.edge.timedata.api.TimedataProvider;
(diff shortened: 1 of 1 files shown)


io.openems.edge.meter.openwb/test/io/openems/edge/meter/openwb/MeterOpenWbImplTest.java
```diff
@@ -0,0 +1,24 @@
...
+
+import static io.openems.common.types.MeterType.CONSUMPTION_METERED;
+import org.junit.Test;
+import io.openems.edge.common.test.AbstractComponentTest.TestCase;
+import io.openems.edge.common.test.ComponentTest;
+
+
+public class MeterOpenWbImplTest {
+
+	@Test
```
Sn0w3y, line 11: Is it possible to add a "working" test, which not only tests the activation of the Component, but also the actual behaviour? You can see how we do it in the Shelly's aswell :)

Would improve Code Coverage.

sthhs: Since this component has been copied from io.openems.edge.tesla.powerwall2.core, the tests are detailed the same way. 
Is there a chance for this PR being accepted? If so, please give me some time to check how they could be improved.

sthhs: Sorry, I can't find any tests in similar components that are using HttpsURLConnection. All test cases are empty and I can't offer a solution since I need at least a template which I can adapt.

sfeilmeier:
Sorry... it's hard to find the time to review those PRs in detail. 

IMHO, even if this implementation is read-only it would still be preferable to implement EVCS, because OpenEMS then shows its own Live Widget with additional information. See other 'read-only' EVCS implementations.

How does this PR relate to https://github.com/OpenEMS/openems/pull/3058?

sthhs:
> IMHO, even if this implementation is read-only it would still be preferable to implement EVCS, because OpenEMS then shows its own Live Widget with additional information. See other 'read-only' EVCS implementations.

Since this implementation only acts as a meter and requests no more than power, current and voltage, I wasn't sure if it's a true EVCS component. But it could be extended to request plug state and charging state, so that EVCS maybe would suit better.
> 
> How does this PR relate to #3058?

This PR (#3237) is intended to show the status of an OpenWB in **primary** mode, where it's totally controlled by itself. In this mode only http requests are accepted by the box and no control is possible from the outside.

PR #3058 uses Modbus to request data from an OpenWB that is in **secondary** mode. This means, that the box is totally controlled by another system like another OpenWB or this OpenEMS component. But since it is not very likely that someone uses an OpenWB in secondary mode without a primary OpenWB, #3058 can be ignored. Maybe I should close it to reduce the amount of open PR's.

sthhs:
Replaced by PR #3299

PR #3250, opened 2025-07-25
Do not bind felix http host to localhost in docker backend

simon-lgb:
Binding to localhost when running on Docker means the Apache Felix web console will only be accessible from inside the docker container, not from the host machine or other containers. This is almost certainly unintended and also conflicts with the provided docker-compose.yml, which suggests exposing the web console port. However, this will not work with the current configuration of felix.


sfeilmeier (approved):
Thank you!

PR #3260, opened 2025-07-30
Modbus Bridge: fix delay for Write-Task

sfeilmeier:
As discussed and suggested in https://community.openems.io/t/asking-for-feedback-new-modbusworker-implementation/1747/5


Sn0w3y:
See https://github.com/OpenEMS/openems/pull/3261

PR #3261, opened 2025-07-31
Implement interruptible wait in Modbus tasks

Sn0w3y:
This update introduces an interruptible wait mechanism to the Modbus bridge's WaitTask and CycleTasksManager classes, allowing for more responsive task switching and reducing unnecessary delays in task execution. Specifically:

1. Added a Mutex-based interrupt mechanism in WaitTask to replace the fixed Thread.sleep() delay, enabling the task to be interrupted and proceed immediately with the next operation. This change allows for more dynamic task management and reduces latency in processing Modbus operations.

2. Modified CycleTasksManager to interrupt the current WaitDelayTask upon executing a write operation, ensuring that write tasks are executed immediately without waiting for the delay to complete. This enhancement addresses a specific issue where write tasks were delayed by up to the Cycle Time due to unexpired wait times, thereby improving the system's responsiveness and throughput.

3. Extended CycleTasksManagerTest to include a test case verifying the immediate execution of write tasks following an onExecuteWrite() call, demonstrating the elimination of the previously observed delay and validating the effectiveness of the interrupt mechanism.

Discussed [here](https://community.openems.io/t/asking-for-feedback-new-modbusworker-implementation/1747/9)

  === TEST: Write task executes immediately after onExecuteWrite ===
  Starting cycle with onBeforeProcessImage()
  Got WaitDelayTask with delay: 0ms
  Delay task thread started, beginning wait...
  Delay task completed normally
  Calling onExecuteWrite() to interrupt the delay
  Got write task immediately: DummyWriteTask [name=WT_1, delay=90]
  Test passed: Write task was available immediately without waiting for delay!
  === END TEST ===

  The test confirms that:
  1. The cycle starts with a WaitDelayTask (0ms in this test case)
  2. When onExecuteWrite() is called, it interrupts the delay
  3. The write task WT_1 is available immediately
  4. The implementation successfully fixes the delay issue

### DIFF
--- a/io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/api/task/WaitTask.java
+++ b/io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/api/task/WaitTask.java
@@ -89,7 +89,6 @@ public abstract non-sealed class WaitTask implements Task {
 		public final long initialDelay;
 
 		private final Runnable onFinished;
-		private final io.openems.common.utils.Mutex interruptMutex = new io.openems.common.utils.Mutex(false);
 
 		private long delay;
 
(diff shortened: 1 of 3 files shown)


Sn0w3y:
@sfeilmeier could you please re-run the Test as this is a false-positive test run :D It actually succeeds locally and also on my CI.

PR #3289, opened 2025-08-25
Fixed enum property in SellToGridLimitState

noctarius:
Since this is a public enum property, renaming might not be as easy.

Please feel free to close the PR if not possible.


noctarius:
Not exactly sure why the build broke, but it seems not related to my change but a missing library in the build process 🤔

PR #3299, opened 2025-08-29
Implement OpenWB Series 2 EVCS

sthhs:
This PR extends and replaces #3237. 
The component requests the actual power, voltage, current, plug state, charge state and energy values from an OpenWB Series 2 with one or two internal charge points in primary mode, which is commonly used. In this case the wallbox can only be monitored from other systems with http requests and can not be controlled.
The purpose is displaying and archiving the most important values in OpenEMS. 
The pictures show the corresponding energy meter values in the OpenWB GUI and in OpenEMS at nearly the same moment.
Bildschirmfoto 2025-07-13 um 14 46 57
Bildschirmfoto 2025-07-13 um 14 46 54

These channels can be requested from the internal charge point:
Bildschirmfoto 2025-07-13 um 17 22 14

Bildschirmfoto 2025-08-29 um 20 29 21

The charged energy is calculated from the difference of the actual imported energy from the OpenWB and the value at the time when the car is plugged in. The values from the OpenWB monitoring and this component are the same: 
Bildschirmfoto 2025-08-29 um 20 29 13

The evcs modal widget:
Bildschirmfoto 2025-08-28 um 20 06 07


io.openems.edge.evcs.openwb/src/io/openems/edge/evcs/openwb/Config.java
```diff
@@ -0,0 +1,43 @@
...
+import org.osgi.service.metatype.annotations.AttributeDefinition;
+import org.osgi.service.metatype.annotations.ObjectClassDefinition;
+
+
+
+@ObjectClassDefinition(name = "Evcs OpenWB", //
+		description = "Implements the evcs component for OpenWB Series2 with internal chargepoints via HTTP API")
+@interface Config {
+	
+	   enum ChargePoint{
```
Sn0w3y: Maybe add this to an "enum" Package as in other Implementations? :)

sthhs: Done

Sn0w3y:
Please still fix your checkstyle Issues

Sn0w3y:
https://openems.github.io/openems.io/openems/latest/contribute/coding-guidelines.html
You do not have only one Checkstyle Issue - it just exits after one..

Sn0w3y:
May i ask, why we can not use Modbus on this EVSE?
https://openwb.de/main/wp-content/uploads/2020/10/ModbusTCP-openWB.pdf
https://openwb.de/main/wp-content/uploads/2023/10/ModbusTCP-openWB-series2-Pro-1.pdf

sthhs:
> May i ask, why we can not use Modbus on this EVSE?

This component is for monitoring the OpenWB in primary mode, which means it is controlled by it's own software.
Modbus can only be activated and used in secondary mode (controlled by another system):
https://wiki.openwb.de/doku.php?id=openwb:api
Bildschirmfoto 2025-11-04 um 07 57 18
Bildschirmfoto 2025-11-04 um 07 57 09

sfeilmeier (approved):
Thanks for continuously updating this PR. I took some time to review and cleanup the code and it is merge ready now.

Some thoughts or future TODOs, still:
- I am not sure HTTP is really the best protocol for OpenWB. There support for MQTT seems to be better and would use much less requests. Currently there is one request per datapoint per Cycle or second
- The `ReadWorker` approach is rather old in OpenEMS. Nowadays we would use the HTTP-Bridge
- `PhaseRotation` should be configurable
- I refactored the part in ReadWorker where you are setting the values in such a way, that the Channels would get reset if there was a read error
- I deleted unused Channels
- You are calculating Energy from Power values; but Session Energy is calculated from the original Energy read from OpenWB. Thats insonsistent.

Feel free to open up new followup PRs. I am merging this one now.

PR #3314, opened 2025-09-17
feat(simulated ess): allow configuration of dis/charge limit

parapluplu:
Currently the dis/charge limit was based on the capacity. With this change it's configurable.

Testing: Executed `java -jar build/openems-edge.jar` and configured simulated ess with custom dis/charge limits and verified them via using the fixed state of charge controller


io.openems.edge.simulator/test/io/openems/edge/simulator/ess/symmetric/reacting/MyConfig.java
```diff
@@ -68,6 +68,16 @@ public int maxApparentPower() {
 		return this.builder.maxApparentPower;
 	}
 
+	@Override
+	public int maxChargePower() {
+		return this.capacity();
```
sfeilmeier: Should be explicitely set as well.

sfeilmeier (approved):

PR #3315, opened 2025-09-17
Update MyModbusDeviceImpl.java

sjjh:
removed exception, as method signature didn't match abstract base class anymore


sjjh:
Creating a new component via Eclipse leads to an error.

sjjh:
@sfeilmeier this (and #3318) should be a quick one. :)

sfeilmeier (approved):

PR #3313, opened 2025-09-17
feat(simulation): Allow to set simulation data via channel

parapluplu:
For testing it's helpful to set production/consumption data externally via channel updates. This controller adds this feature

Testing:
```
> curl -d '{"value":6000000}' http://admin:admin@localhost:8084/rest/channel/simulateConsumption/Data
{}
> curl http://admin:admin@localhost:8084/rest/channel/_sum/ConsumptionActivePower
{"address":"_sum/ConsumptionActivePower","type":"INTEGER","accessMode":"RO","text":"","unit":"W","value":6000000}
> curl -d '{"value":1}' http://admin:admin@localhost:8084/rest/channel/simulateConsumption/Data
{}
> curl http://admin:admin@localhost:8084/rest/channel/_sum/ConsumptionActivePower
{"address":"_sum/ConsumptionActivePower","type":"INTEGER","accessMode":"RO","text":"","unit":"W","value":1}
```

PR #3319, opened 2025-09-18
Introduce integration tests

parapluplu:
This pull requests adds a folder `integration_tests` that enables us to execute integration tests. The framework automatically spins up an openems edge instance before each test, loads desired configuration files and executes the test.

More infos about how to use the integration tests are in the `integration_tests/README.md` file.

Prerequisites of this pull request are two other pull requests. For now they are included as commit in here, but since these are changes in OpenEMS components I think it makes sense to review them separately. Once they are merged I'll remove them from this pull request

https://github.com/OpenEMS/openems/pull/3314
https://github.com/OpenEMS/openems/pull/3313


da-Kai:
@parapluplu did you continue working on this? 
I would suggest to move `integration_tests` into the `tools` directory and then let's get this merged

parapluplu:
> @parapluplu did you continue working on this? I would suggest to move `integration_tests` into the `tools` directory and then let's get this merged

Yep, we're using the integration tests as part of our own pipeline for a while now.
I've moved the directory into the tools directory

da-Kai (approved):

sfeilmeier:
Thank you! This sounds like a very useful feature. We should cover it in the official documentation as well - and probably discuss use-cases in the Community. 👍

PR #3321, opened 2025-09-20
Updated the MQTT component to optionally filter its messages by topic paths

da-Kai:
> The current implementation of MQTT can turn in to a bit of a fire hose of data for situations where you only want some of the fields possible. This patch adds a filter function which inhibits certain topics from actually sending to the broker, instead they silently complete, doing nothing.
> 
> A single new configuration string has been added, into which a semicolon-delimited set of topic filters can be written allowing only topics which match these specifications to pass.
> 
> The topic filters use the same syntax as normal MQTT client subscriptions, so can use the + and # specifiers to indicate single, or multi-level paths.
> 
> An example path might be edge/fems33/channel/meter+/+ which would pull all metrics in meter0, meter1, etc. paths, but exclude any other topics.
> 
> THIS NEEDS TESTING - Please can someone with an otherwise working setup please check that this works, as I'm currently unable to, as I don't have access to our test hardware.

Fork of #1503 by @JohnVidler


io.openems.edge.controller.api.mqtt/src/io/openems/edge/controller/api/mqtt/Config.java
```diff
@@ -43,6 +43,9 @@
 
 	@AttributeDefinition(name = "Trust Store", description = "The trust store in PEM format")
 	String trustStorePem();
+	
+	@AttributeDefinition(name = "Filter Spec", description = "A list of valid topic paths. Follows the MQTT topic subscription rules for globbing.")
+	String filterSpec();
```
sfeilmeier: If it's a list, it should be an array - instead of a ";"-separated string. Also an example should be given here and in the README.

io.openems.edge.controller.api.mqtt/src/io/openems/edge/controller/api/mqtt/ControllerApiMqttImpl.java
```diff
@@ -87,12 +96,41 @@ public ControllerApiMqttImpl() {
...
 
 		super.activate(context, config.id(), config.alias(), config.enabled());
 
-		this.scheduleReconnect();
+		if (this.isEnabled()) {
+			this.scheduleReconnect();
+		}
+	}
+
+	private Pattern[] getFilter(Config config) {
```
sfeilmeier: Either make this static or provide `Config` as parameter - not both mixed.

io.openems.edge.controller.api.mqtt/src/io/openems/edge/controller/api/mqtt/ControllerApiMqttImpl.java
```diff
@@ -87,12 +96,41 @@ public ControllerApiMqttImpl() {
...
+		String[] buildList = config.filterSpec().split(";");
+		final var newFilter = new Pattern[buildList.length];
+		for (int i = 0; i < buildList.length; i++) {
+			try {
+				newFilter[i] = Pattern.compile(//
+						"^" + buildList[i].replace("+", "[^/]+").replace("#", ".+") + "$"//
+				);
+			} catch (PatternSyntaxException e) {
+				this.logWarn(this.log, "Unable to parse filter pattern: '" + buildList[i] + "'! Filter disabled!");
+				this.log.warn(e.getMessage(), e);
```
sfeilmeier: Inconsistent usage of `log.warn` vs `logWarn

io.openems.edge.controller.api.mqtt/test/io/openems/edge/controller/api/mqtt/MyConfig.java
```diff
@@ -147,4 +147,9 @@ public String privateKeyPem() {
 	public String trustStorePem() {
 		return this.builder.trustStorePem;
 	}
+
+	@Override
+	public String filterSpec() {
+		return "";
```
sfeilmeier: Should be `return this.builder.filterSpec;` and used in the JUnit test

sfeilmeier (approved):
Nice solution 👍

PR #3341, opened 2025-10-03
Fix TypeError: ctx_r1.subscribeChannels is not a function

Sn0w3y:


Sn0w3y:
Sorry, i found out that the Issue is a little bit deeper - i will close this, as this is not the intended behaviour

PR #3364, opened 2025-10-15
change console output format from plan to colored in gradle build of ci

miettal:
from gradle 9.1, it support colord option (I contributed this). --console=rich break ci output, because it designed rich terminal(VT100). but this option is for ci, not break ci output.


miettal:
Oh!!

This project output is hooked! this flag is not needed!!

https://github.com/OpenEMS/openems/blame/799ee37a0a380da5867da71ceb91df5e4647a378/build.gradle#L55

da-Kai (approved):

PR #3368, opened 2025-10-20
🛠️ Fix: Prevent "Interrupted system call" when writing to `/etc/apk/repositories`

Sn0w3y:
## 📌 Summary
This PR fixes a sporadic build failure during multi-arch Docker builds (`linux/amd64`, `linux/arm64`) caused by a race condition when writing to `/etc/apk/repositories` using `echo`. The error message was:
```
/bin/sh: can't create /etc/apk/repositories: Interrupted system call
```

## ✅ Solution
Replaced the fragile `echo` approach with a more robust sequence that ensures directories and files exist and uses `printf` instead:

### Before
```dockerfile
RUN wget -O /etc/apk/keys/adoptium.rsa.pub https://packages.adoptium.net/artifactory/api/security/keypair/public/repositories/apk && \
    echo 'https://packages.adoptium.net/artifactory/apk/alpine/main' >> /etc/apk/repositories
```

### After
```dockerfile
RUN mkdir -p /etc/apk/keys && \
    wget -O /etc/apk/keys/adoptium.rsa.pub https://packages.adoptium.net/artifactory/api/security/keypair/public/repositories/apk && \
    touch /etc/apk/repositories && \
    printf '%s\n' 'https://packages.adoptium.net/artifactory/apk/alpine/main' >> /etc/apk/repositories
```

## 📦 Result
- Fixed build failure on multi-architecture builds
- Improved reliability and reproducibility of Docker image builds

As seen in:

https://github.com/OpenEMS/openems/actions/runs/18647900625/job/53164486257

### DIFF
--- a/tools/docker/edge/Dockerfile
+++ b/tools/docker/edge/Dockerfile
@@ -22,8 +22,10 @@ FROM ghcr.io/linuxserver/baseimage-alpine:edge AS base_container
 
 ARG JAVA_VERSION
 
-RUN wget -O /etc/apk/keys/adoptium.rsa.pub https://packages.adoptium.net/artifactory/api/security/keypair/public/repositories/apk && \
-    echo 'https://packages.adoptium.net/artifactory/apk/alpine/main' >> /etc/apk/repositories
+RUN mkdir -p /etc/apk/keys && \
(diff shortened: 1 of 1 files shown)


da-Kai (changes requested):
You should then add this to the edge Dockerfile aswell.

Sn0w3y:
@da-Kai done ! :)

PR #3376, opened 2025-10-23
Fix little Bug in Enforcement of tranlsations to Upper Case

Sn0w3y:


ui/src/app/user/user.component.html
```diff
@@ -39,7 +39,7 @@
                   {{"MENU.ACCESS_LEVEL" | translate}}
                 </ion-label>
                 <ion-label>
-                  {{ 'Role.' + user.globalRole | translate }}
+                  {{ ('ROLE.' + (user.globalRole?.toUpperCase() || '')) | translate }}
```
sfeilmeier, line 42: Function calls should be avoided, because they get executed all the time

Issue #3387, opened 2025-10-28
UI: dashboard energy monitor: gauge diagram does not match textual values


sjjh:
### Description

Dashboard energy monitor widget: There is a mismatch between the bars within the gauge diagram and the textual values. See below screenshot, there the bars of the gauge show approx. following fractions:
- 1/3 from gird
- 2/3 from pv inverters
- 3/3 consumption
Looking at the textual values, there is a total different split, approx.:
- 4/5 from grid
- 1/5 from pv inverters
- 5/5 consumption
Looking at the clouds outside, the textual values seem to be correct, and the diagram seems to be wrong. The arrows seem to follow the textual values.

### Screenshots

<img width="598" height="658" alt="Image" src="https://github.com/user-attachments/assets/c9ba3b88-5718-4fd4-8dfd-a04b9c7328cf" />

### Operating System

Edge/UI: docker; Client: Linux, Firefox

### How to reproduce the Error?

Install Edge, UI, add some pv inverters and a grid meter.

Sn0w3y:
The bars in the gauge are not based on the current power distribution but on the maximum values ever recorded for each source.
So they do not represent the actual live ratio between grid, PV, and consumption. Instead, they show how the current values compare to their respective historical maximums.
That’s why the textual values and arrows (which reflect the real-time power flow) can differ significantly from the bar lengths, which are only scaled to the peak values seen so far.

@da-Kai am I right here? So i do not see this as an "Issue"

da-Kai:
@Sn0w3y you are correct. So its not a functional issue but maybe a UX flaw.

We could probably clarify this behavior in the UI or add a tooltip.

sjjh:
Thanks for your comments and eagerness to improve the situation. I'm tempted to respond with
> What's the commonality between a joke and a UI? -- If you have to explain it, it's not good.

IMHO #3392 is a hot-fix, that improves the situation, but probably not the best solution. Not knowing what is possible, I would suggest to conduct user testing to find out if I'm the only person miss-reading the diagram, and if not, redesign the visualization. Tbh, I don't have a better design off the top of my head, ideally a UX designer with a graphical background would be involved.

Sn0w3y:
ideally a UX designer with a graphical background would be involved.

- you mean the Graphical Designers which already Designes the one how it is right now? I guess the Energy Monitor is one of the Core Components which has the same exact Style and Functionality since OpenEMSes Start. Changing it seems not to be a solution in mz Opinion.

[...]

PR #3392, opened 2025-11-02
[UI] Add tooltip to Energymonitor

Sn0w3y:
This "fix" Adresses [Issue ](https://github.com/OpenEMS/openems/issues/3387)


sfeilmeier:
I'm marking this as draft for now till we finish discussion at https://github.com/OpenEMS/openems/issues/3387, because I agree with
>  What's the commonality between a joke and a UI? -- If you have to explain it, it's not good.

janklostermann:
Wouldn't it be helpful to just merge this PR as a quick fix to instantly help people having this question?
Of cause this should not stop the started discussion nor prevent a better solution! (I do agree to the UX issue.)

janklostermann:
What would be needed to get this PR merged (apart from the now needed update)?
Is there anything speaking against it? Could I help fixing?

PR #3366, opened 2025-10-19
Improve UI docker

da-Kai:
* Split into `Dockerfile.backend` and `Dockerfile.edge`
* Proxy websocket connection via container
* Unify url setup in environments
* Remove docker anchor in `index.html`
* Add Rest-Api endpoint
* Split `UI_WEBSOCKET` env variable

PR #3400, opened 2025-11-07
Fix WebSocket backend URL when running UI via openems-edge-dev configuration

Venkata-Sai-Vishwanath-robo:
### Description
This PR fixes a WebSocket connection issue encountered when following the Getting Started documentation to run the UI with:
`ng serve -c openems-edge-dev`

### Problem
The current environment.openems-edge-dev.ts uses:
`` url: `${getWebsocketScheme()}://${location.host}/backend`, ``

When running the UI dev server, location.host resolves to localhost:4200, so the UI tries to connect to:
ws://localhost:4200/backend
However, the OpenEMS Edge WebSocket Bridge runs on port 8085, not 4200.
This results in a failed WebSocket handshake and the UI does not connect to Edge:
image

### Solution
Explicitly point the WebSocket URL to port 8085 while keeping hostname dynamic:
`` url: `${getWebsocketScheme()}://${location.hostname}:8085`, ``

### DIFF
--- a/ui/src/themes/openems/environments/edge-dev.ts
+++ b/ui/src/themes/openems/environments/edge-dev.ts
@@ -5,7 +5,7 @@ export const environment: Environment = {
     ...theme, ...{
 
         backend: "OpenEMS Edge",
-        url: `${getWebsocketScheme()}://${location.host}/backend`,
+        url: `${getWebsocketScheme()}://${location.hostname}:8085`,
 
         production: false,
         debugMode: true,


Sn0w3y:
Why don't we just return to the initial Usage:

`    url: getWebsocketScheme() + location.hostname + ":8085",
`

Venkata-Sai-Vishwanath-robo:
Hello @Sn0w3y,

I actually tried doing that, but it doesn't work. 
After some investigation, I found out that the `getWebsocketScheme()` function previously returned the WebSocket protocol including the `://` separator, but the current version it doesn't include the `://`.  (The changes were made in #3366 to make the urls unified).

da-Kai:
@Venkata-Sai-Vishwanath-robo thanks, your solution looks right to me. 
With suburls only works for docker containers. I obviously made a mistake in #3366

sfeilmeier (approved):

PR #3410, opened 2025-11-12
fix: move injected property assignments to constructor to avoid TS2729

Jasonlee6789:
Description:
This PR fixes TS2729 errors: Property 'xxx' is used before its initialization, refactors all class property initializations that previously used injected services (such as TranslateService, Service, PlatFormService, etc.) at the field declaration level. 
In strict TypeScript mode, these injected properties are not available until the constructor runs, which causes TS2729 errors.

Key changes:
Moved all assignments that depend on injected services from field declarations to the constructor.

Benefits:
Prevents runtime and compile-time errors in strict TypeScript environments.
Improves compatibility with future Angular and TypeScript upgrades.
No functional changes; only initialization timing is affected.
Codebase is now safer and more robust for teams using strict typing.

### DIFF
--- a/ui/src/app/edge/history/Controller/EnerixControl/overview/overview.ts
+++ b/ui/src/app/edge/history/Controller/EnerixControl/overview/overview.ts
@@ -14,10 +14,7 @@ import tr from "./translation.json";
     standalone: false,
 })
 export class OverviewComponent extends AbstractHistoryChartOverview {
-    protected readonly STATES: string = `
-    1.${this.translate.instant("Edge.Index.Widgets.ENERIX_CONTROL.NO_INPUT")}
-    2.${this.translate.instant("Edge.Index.Widgets.ENERIX_CONTROL.NO_DISCHARGE")} 
-    `;
+  protected readonly STATES: string;
 
     // disabled till next release
     // 3.${this.translate.instant("Edge.Index.Widgets.ENERIX_CONTROL.FORCE_CHARGE")}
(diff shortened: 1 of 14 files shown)


Sn0w3y:
Hi @Jasonlee6789,

  Thank you for this PR! We reviewed the changes and have a question about the TS2729 errors you encountered.

  ## Our Setup (identical to upstream `develop`)
  - **TypeScript**: 5.8.3
  - **Angular**: 20.3.6
  - **tsconfig.json**: `"strict": false`

  When we compile the codebase on both `develop` and our branch, we **don't get any TS2729 errors** - neither in the console nor when running `npx tsc --noEmit --strict`.

  ## Questions
  1. Are you testing against a different branch or configuration?
  2. Are you using any additional TypeScript compiler flags or IDE settings that might trigger this error?
  3. Can you share steps to reproduce the TS2729 error with the current `develop` configuration?

Jasonlee6789:
Hi @Sn0w3y , Thank you for reviewing the PR and for your feedback.

Let me answer them directly and then provide a more detailed explanation of my findings.

  1. Are you testing against a different branch or configuration?

  I was indeed testing with the same configuration as the upstream 2025.11.0 develop branch. I did not test against any other branches.

  2. Are you using any additional TypeScript compiler flags or IDE settings that might trigger this error?

  I am not using any additional TypeScript compiler flags or IDE settings. My tsconfig.json file is consistent with the upstream configuration, and the strict option is set to false. 

  3. Can you share steps to reproduce the TS2729 error with the current develop configuration?

  This is the most complex part. Currently, I can no longer reliably reproduce the TS2729 error after a full reboot of my local PC system. This leads me to believe the error was triggered by a temporary state in my local environment. 

I am sorry about this confusing part, I suggest to close this PR for now

PR #3414, opened 2025-11-14
EEBUS Bridge

sfeilmeier:


clehne:
Looks like a good starting point..

PR #3419, opened 2025-11-16
Enhance OAuthBackendImpl with dynamic backend reference

Sn0w3y:
1. Made the backend reference optional - Added cardinality = ReferenceCardinality.OPTIONAL so the component can activate even when Controller.Api.Backend is not configured

2. Added null safety checks - All three methods now check if backend is available before using it

3. Added proper error handling - Returns a failed CompletableFuture with a descriptive error message when backend is unavailable

4. Made the reference dynamic and volatile - Follows the same pattern as UserServiceImpl and AppCenterBackendUtilImpl

### DIFF
--- a/io.openems.edge.core/src/io/openems/edge/core/componentmanager/OsgiValidateWorker.java
+++ b/io.openems.edge.core/src/io/openems/edge/core/componentmanager/OsgiValidateWorker.java
@@ -15,7 +15,6 @@ import org.osgi.service.cm.ConfigurationAdmin;
 import org.osgi.service.cm.ConfigurationEvent;
 import org.osgi.service.component.runtime.ServiceComponentRuntime;
 import org.osgi.service.component.runtime.dto.ComponentConfigurationDTO;
-import org.osgi.service.component.runtime.dto.ComponentDescriptionDTO;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


sfeilmeier:
@michaelgrill: What do you think? This currently causes the issue reported here: https://community.openems.io/t/fehler-und-wie-finde-ich-den-componentmanager-defective-null-unsatisfied-reference-for-backend/10609/2

vision1727:
tested the changes of @Sn0w3y works now without this error "unsatisfied reference"

michaelgrill:
i would prefer a solution where we dont have to touch the `OAuthBackendImpl` class. IMO it just adds a lot of unnecessary code to it. (and future components like this)

Can we change the validation to only validate classes, which implement `OpenemsComponent`? Because in this case its only required if its used by `TimeOfUseTariffLuoxImpl` and in future i believe the log is only required if you have a `OpenemsComponent` using it.

io.openems.edge.core/src/io/openems/edge/core/componentmanager/OsgiValidateWorker.java
```diff
@@ -143,6 +143,11 @@ private static void updateInactiveComponentsUsingScr(Map<String, String> defecti
 			ServiceComponentRuntime scr) {
 		var descriptions = scr.getComponentDescriptionDTOs();
 		for (ComponentDescriptionDTO description : descriptions) {
+			// Only validate components that implement OpenemsComponent
+			if (!Stream.of(description.serviceInterfaces)
```
iseeberg79: looks good! can't description.serviceInterfaces be null?
equivalent implementation is running fine here

iseeberg79:
could you update the PR information accordingly? Latest changes are quite different ;)

sfeilmeier:
I'll work on this to get rid of this issue for next release

Sn0w3y:
> I'll work on this to get rid of this issue for next release

Shall i close this PR or would you work in this PR for it?

sfeilmeier (approved):
Thank you! 👍

PR #3412, opened 2025-11-12
Fix: updated Docker deployment docs (issue #3411)

akshith30388:
This PR updates the Docker deployment documentation to:
- Replace deprecated UI_WEBSOCKET with WEBSOCKET_HOST and WEBSOCKET_PORT
- Add a “Changing Language in UI” section
- Add a “Next Steps” section for clarity

Fixes: #3411

PR #3427, opened 2025-11-18
CLAUDE.md file as context store for AI support

janklostermann:
CLAUDE.md to provide context on the project for AI assistants.


Sn0w3y:
To be honest I do not know if @sfeilmeier wants AI's to even commit to the Repo (he has the last word on this) but IMHO AI hallucinates a little bit too much for beeing helpful on OpenEMS. (See https://github.com/OpenEMS/openems/pull/3412#discussion_r2537076987)

Also if one wants to use AI on this Project one should and could generate his own Claude.md File or Context file to his own extend. (Do not know if this is helpful for OpenEMS and should actually be included)..

janklostermann:
The Kostal bug I was able to fix using AI (tested in reality) - thus: not just halluzination.
In the end it is all about testing and quality assurance. Humans halluzinate quite a bit as well...

janklostermann:
The purpose of having a joint CLAUDE.md in the project is to have a starting point for an AI about the context it is living in (providing it context about the project). 
Therefore I think it could be helpful to have a joint effort to improve such a context-giving document. For private preferences other files can then serve, that everyone keeps private. 
The name CLAUDE.md could be changed, to avoid the misunderstanding that it only serves claude.ai.

Sn0w3y:
I’m critical about this idea. A central CLAUDE.md or any generic “AI context file” inside the OpenEMS repository is, in my opinion, out of place.

Why?

The repository is not the right place for AI instructions.
It contains code, documentation, and architecture descriptions - but not personal prompt setups or context collections for individual tools.

Everyone uses AI differently.
Some use ChatGPT, others use Claude, others don’t use AI at all. A shared context file adds more confusion than value.

Risk of outdated or incorrect information.
Such a file will not be maintained the same way code is. It would quickly become outdated and potentially misleading for newcomers.

AI assistance is an individual tool.
Anyone who wants to use AI can create their own context file (Claude.md, chatgpt.md, etc.), just like everyone maintains their own editor or IDE configurations.

OpenEMS should remain neutral.
Having an official AI context file in the repository gives the impression that it is recommended or part of the development process - which it isn’t.

In short: In my Opinion this does not belong in the project - but as already said Stefan has the last word on these Decisions :)

da-Kai:
Hi @janklostermann,

Thanks for your PR. 

After going through your CLAUDE.md file, i am not sure, if adding it in this form really helps the project or AI. The file is quite large and repeats a lot of things we already have in the repo. It combines info from a lot of sources like README, gradle-files, pom.xml, package.json, etc. Realistically, it won’t stay in sync with the actual state of the project and will drift over time. That risks confusing AI agents rather than helping them, thus becoming bloat instead.

What would be genuinely useful is improving the existing files, like extending the README.md and maybe adding an ARCHITECTURE.md ([examples](https://github.com/noahbald/awesome-architecture-md)) that explains the projects structure. That kind of documentation would give both developers and AI assistants real value without duplicating what we already have.

sfeilmeier:
I fully supported @da-Kai's view on this

janklostermann:
I will withdraw this PR. It was premature.

PR #3443, opened 2025-11-26
[UI] Only apply additional filter when filter values are selected

Sn0w3y:
Previously, when opening and closing the SumState filter dropdown without selecting any values, the `isOnline: true` additional filter was still being applied. This caused the device list to show only online devices even when no SumState filter was selected.

The issue was in `filter.component.ts` where `additionalFilter` was set unconditionally after the if/else block. Now it's only set when actual filter values are selected.

Fixes: Opening SumState filter and closing it without selection no longer filters to online-only devices.

### DIFF
--- a/ui/src/app/index/filter/filter.component.ts
+++ b/ui/src/app/index/filter/filter.component.ts
@@ -190,9 +190,9 @@ export class FilterComponent {
         } else {
             this.searchParams.set(filter.category, values);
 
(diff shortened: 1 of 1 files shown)


Sn0w3y:
@sfeilmeier could we add this aswell before the next Release? :)

sfeilmeier:
@lukasrgr FYI

PR #3447, opened 2025-11-27
fix(near equals power solver): Invalid behavior for limits

parapluplu:
This fixes an error where a charge limit was set for a single ess inside a cluster. The power solver disrespected this, and set a value outside of the limit. For details see unit test part `#4` of `testNearEqualDistribution in io.openems.edge.ess.core/test/io/openems/edge/ess/core/power/PowerComponentTest.java. Without this fix, this test fails and charges higher than the allowed -1900.

Testing: Unit tests, also tested in a setup with a power plant with 12 ESS in total stacked in two layers of ESS clusters. First layer is an cluster of 3 Edge2Edge ESS, second layer is a Cluster of 4 EssGenericManagedSymmetric each.

### DIFF
--- a/io.openems.edge.ess.core/src/io/openems/edge/ess/core/power/optimizers/KeepAllNearEqual.java
+++ b/io.openems.edge.ess.core/src/io/openems/edge/ess/core/power/optimizers/KeepAllNearEqual.java
@@ -48,17 +48,10 @@ public class KeepAllNearEqual {
 		var reactivePowerSolved = solvePowerIfNotNaN(setReactivePower, essList, Pwr.REACTIVE, direction);
 
 		var mergedResult = mergeResults(coefficients, esss, activePowerSolved, reactivePowerSolved);
-		if (mergedResult == null) {
+		if (mergedResult == null || mergedResult.length == 0) {
 			return null;
 		}
-
-		var result = Arrays.stream(mergedResult)//
-				.map(d -> reverseAbsoluteData(d, direction))//
-				.toArray();
-		if (result.length == 0) {
-			return null;
-		}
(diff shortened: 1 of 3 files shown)


parapluplu:
Could somebody explain to me we codecov/patch is failing with a 50% diff target? I've basically removed code and added tests. The contrary should happen and 50% is waaaaay out of line

Sn0w3y:
> Could somebody explain to me we codecov/patch is failing with a 50% diff target? I've basically removed code and added tests. The contrary should happen and 50% is waaaaay out of line

TBH i never understood codecov here :D

parapluplu:
I'd like to publish another merge request, that improves the null handling of the power solver in general. However, as it builds on top of this change, I would like to make my life easier and merge this one before publishing the other one.
Also they are different kind of changes, why I want to keep them seperated as commits. Is there any chance the change in this merge request gets merged?

parapluplu:
https://github.com/OpenEMS/openems/commit/701d7b839d8ed6023bddfb3d4f15f4a8bc8f50ba

This was merged 7 hours ago with the back port merge and creates conflicts for this pull request. So if this merge request does not get any further traction I will drop this topic and will keep the changes and follow up changes in our internal repo since this now creates significant overhead.

I've also changed the null handling to optionals, which makes it much more explicit and enforces correct handling whenever touching the values. Seems like Fenecon decided to go for a approach with using a constant instead of Optionals https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.ess.core/src/io/openems/edge/ess/core/power/optimizers/KeepAllNearEqual.java#L27

I would prefer a more modern approach provided by java

PR #3456, opened 2025-12-01
[EDGE] Refactor OpenWB to use MQTT

Sn0w3y:


sfeilmeier:
As discussed in private, I would prefer a separate "MQTT Bridge"

Sn0w3y:
> As discussed in private, I would prefer a separate "MQTT Bridge"

I will close this PR and after i finished the MQTT Bridge Implementation i need to Refactor it to use it :) Then i will open another one again :)

PR #3457, opened 2025-12-01
Add Victron Package

Sn0w3y:
As Discussed here: https://community.openems.io/t/verknupfung-ess-controller-in-beide-richtungen/3170/14

@DerWahreKlinki
@sfeilmeier

### DIFF
--- a/io.openems.edge.victron/test/io/openems/edge/victron/battery/MyConfig.java
+++ b/io.openems.edge.victron/test/io/openems/edge/victron/battery/MyConfig.java
@@ -1,8 +1,7 @@
-package io.openems.edge.battery.victron;
+package io.openems.edge.victron.battery;
 
 import io.openems.common.test.AbstractComponentConfig;
 import io.openems.common.utils.ConfigUtils;
-import io.openems.edge.victron.battery.Config;
 
 @SuppressWarnings("all")
 public class MyConfig extends AbstractComponentConfig implements Config {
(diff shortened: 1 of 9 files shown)


sfeilmeier (approved):
First of all, I can see that a lot of effort has gone into writing this code—thank you all for your work! I want to be honest in my feedback: if this code were intended for FEMS, there are some core concepts of OpenEMS and OSGi that I would approach differently. Some patterns here seem to stem from a misunderstanding of certain core ideas.

That said, I also recognize that if it works for OpenEMS users and doesn’t break other parts of the code, it can still be valuable to merge. My only caution is that I cannot guarantee future changes in OpenEMS won’t affect this Victron implementation.

Overall, great effort and thanks again for contributing!

Sn0w3y:
> First of all, I can see that a lot of effort has gone into writing this code—thank you all for your work! I want to be honest in my feedback: if this code were intended for FEMS, there are some core concepts of OpenEMS and OSGi that I would approach differently. Some patterns here seem to stem from a misunderstanding of certain core ideas.
> 
> That said, I also recognize that if it works for OpenEMS users and doesn’t break other parts of the code, it can still be valuable to merge. My only caution is that I cannot guarantee future changes in OpenEMS won’t affect this Victron implementation.
> 
> Overall, great effort and thanks again for contributing!


Would be great if you would tell me/us, what things could be changed to align :) I would love to sit down to Improve the Code for ongoing things ;)

sfeilmeier:
Ok, fair enough :-)

- In some files there are multiple OSGi @Reference-Tags that internally change the same field. It's hard to guess what will actually happen
- There are quite some circular dependencies between the components; that's always a sign for bad software architecture.
- The Battery-Inverter and Battery are not the same kind of universal Battery and Battery-Inverter that are used with Generic-ESS. It would have maybe been better to not reuse the Nature, as it is misleading -> this is my biggest concern. If we ever change the Battery/Battery-Inverter nature I will likely just remove these `implements` from the Victron components
- `handleEvent()` always has to validate that the Component is enabled
- There are not JUnit tests for all Components

PR #3465, opened 2025-12-05
fixed Kostal Piko HTML parsing bug: kWh for Total Energy in HTML, instead of expected Wh in ActiveProductionEnergy

janklostermann:
As the parsed HTML from the Kostal Piko delivers kWh, the value should be multiplied by 1000 before it is stored into ActiveProductionEnergy, that expects Wh, by `this._setActiveProductionEnergy(totalYield);` (line 321)
This PR adds exactly this adjustment.

### DIFF
--- a/io.openems.edge.pvinverter.kostal/src/io/openems/edge/pvinverter/kostal/piko/PvInverterKostalPikoImpl.java
+++ b/io.openems.edge.pvinverter.kostal/src/io/openems/edge/pvinverter/kostal/piko/PvInverterKostalPikoImpl.java
@@ -193,10 +193,13 @@ public class PvInverterKostalPikoImpl extends AbstractOpenemsComponent
 			acPower = value != null ? value : 0;
 		}
 
-		// Total Energy - second value
+		// Total Energy - second value (HTML provides kWh, need to convert to Wh)
 		Long totalYield = null;
 		if (index < valueCells.size()) {
 			totalYield = this.parseLongValue(valueCells.get(index++).text());
+			if (totalYield != null) {
+          		totalYield = totalYield * 1000; // Convert kWh to Wh
+      		}
 		}
 
 		// Day Energy - third value


Sn0w3y:
Can confirm! This is Indeed a Bug :)

image

io.openems.edge.pvinverter.kostal/src/io/openems/edge/pvinverter/kostal/piko/PvInverterKostalPikoImpl.java
```diff
@@ -193,10 +193,13 @@ private void parseHtmlResponse(String html) throws OpenemsException {
 			acPower = value != null ? value : 0;
 		}
 
-		// Total Energy - second value
+		// Total Energy - second value (HTML provides kWh, need to convert to Wh)
 		Long totalYield = null;
 		if (index < valueCells.size()) {
 			totalYield = this.parseLongValue(valueCells.get(index++).text());
```
sfeilmeier, line 199: Just as a note: you could solve this a lttle bit shorter with 
```
Optional.ofNullable(this.parseLongValue(valueCells.get(index++).text()))
    .map(v -> v * 1000) // Convert kWh to Wh
    .orElse(null);
```

janklostermann: :-) But not with my depth of Java knowledge...

janklostermann:
Beyond that: I get really strange effects applying this bugfix, as I described in https://community.openems.io/t/connecting-kostal-piko-pv-inverter-to-openems/10231/47
Therefore I am not sure if this is all that needs to be fixed...

sjjh:
> Can confirm! This is Indeed a Bug :)
> 
> image

Sorry for beeing late to the party, yes can confirm as well, that this fixes the bug.

PR #3474, opened 2025-12-15
fix simulator performance on global cache

iseeberg79:
I use a Docker installation on a VM with very low-performance hardware.
The simulator consumes a lot of resources on my system. The changes reduce system utilization by 70%. By implementing these changes, I was able to significantly improve problems with memory and CPU utilization without any loss of functionality.

There may be another solution, but I would like to document the improvements and suggest changes to the simulator.

@sfeilmeier we exchanged already about - just for reference


iseeberg79:
ready to review but waiting for comments

iseeberg79:
cache removed by df93e36

PR #3479, opened 2025-12-16
Edge2Edge: filter undefined integer modbus values

tcdrop:
This change fixes an issue with the internal OpenEMS Edge-2-Edge communication. It has been discussed in [this forum post](https://community.openems.io/t/edge-2-edge-ess-ungultige-channel-werte/10654/7).

Short Summary: when an OpenEMS edge has an `undefined` integer channel value, it sets the value to the max. int value (e.g. 65535 for 16bit unsigned integer) in the Modbus API. The master edge then reads this Modbus value, but doesn't recognize it as `undefined`. Instead, the value gets used just like any other value. This can lead to problems like SOC of 65535%.

This fix adds a converter for integer values to the `AbstractEdge2Edge` class, that sets channel values to `null`, if the Modbus value is `undefined`.

### DIFF
--- a/io.openems.edge.edge2edge/test/io/openems/edge/edge2edge/common/AbstractEdge2EdgeTest.java
+++ b/io.openems.edge.edge2edge/test/io/openems/edge/edge2edge/common/AbstractEdge2EdgeTest.java
@@ -30,38 +30,44 @@ public class AbstractEdge2EdgeTest {
 	public void testGetConverterForType() {
 		{
 			final var converter = AbstractEdge2Edge.getConverterForType(ModbusType.FLOAT32);
+			assertNull(converter.elementToChannel(null));
 			assertNull(converter.elementToChannel(Float.NaN));
 			assertNull(converter.elementToChannel(ModbusRecordFloat32.UNDEFINED_VALUE));
 			assertEquals(123.456F, (float) converter.elementToChannel(123.456), 0.001);
 		}
 		{
 			final var converter = AbstractEdge2Edge.getConverterForType(ModbusType.FLOAT64);
+			assertNull(converter.elementToChannel(null));
 			assertNull(converter.elementToChannel(Double.NaN));
 			assertNull(converter.elementToChannel(ModbusRecordFloat64.UNDEFINED_VALUE));
 			assertEquals(123.456, (double) converter.elementToChannel(123.456), 0.001);
 		}
 		{
 			final var converter = AbstractEdge2Edge.getConverterForType(ModbusType.STRING16);
+			assertNull(converter.elementToChannel(null));
 			assertEquals(16, ((String) converter.elementToChannel(ModbusRecordString16.UNDEFINED_VALUE)).length());
 			assertEquals("OpenEMS", converter.elementToChannel("OpenEMS"));
 		}
 		{
 			final var converter = AbstractEdge2Edge.getConverterForType(ModbusType.UINT16);
-			assertNull(converter.elementToChannel((short) 0xFFFF));
+			assertNull(converter.elementToChannel(null));
+			assertNull(converter.elementToChannel(0xFFFF));
 			assertNull(converter.elementToChannel(ModbusRecordUint16.UNDEFINED_VALUE));
 			assertEquals(0, converter.elementToChannel(0x0000));
 		}
 		{
 			final var converter = AbstractEdge2Edge.getConverterForType(ModbusType.UINT32);
-			assertNull(converter.elementToChannel(0xFFFFFFFF));
+			assertNull(converter.elementToChannel(null));
+			assertNull(converter.elementToChannel(0xFFFFFFFFL));
 			assertNull(converter.elementToChannel(ModbusRecordUint32.UNDEFINED_VALUE));
-			assertEquals(0, (int) converter.elementToChannel(0x00000000));
+			assertEquals(0L, converter.elementToChannel(0x00000000));
 		}
 		{
 			final var converter = AbstractEdge2Edge.getConverterForType(ModbusType.UINT64);
+			assertNull(converter.elementToChannel(null));
 			assertNull(converter.elementToChannel(0xFFFFFFFFFFFFFFFFL));
 			assertNull(converter.elementToChannel(ModbusRecordUint64.UNDEFINED_VALUE));
-			assertEquals(0, (long) converter.elementToChannel(0x0000000000000000));
+			assertEquals(0L, converter.elementToChannel(0x0000000000000000));
 		}
 	}
 }
--- a/io.openems.wrapper/bnd.bnd
+++ b/io.openems.wrapper/bnd.bnd
@@ -5,10 +5,10 @@ Bundle-Description: This wraps external java libraries that do not have OSGi hea
 -buildpath: \
 	com.github.rzymek:opczip;version='1.2.0',\
 	com.google.gson;version='2.13.2',\
-	com.influxdb:flux-dsl;version='7.4.0',\
-	com.influxdb:influxdb-client-core;version='7.4.0',\
-	com.influxdb:influxdb-client-java;version='7.4.0',\
-	com.influxdb:influxdb-client-utils;version='7.4.0',\
+	com.influxdb:flux-dsl;version='7.5.0',\
+	com.influxdb:influxdb-client-core;version='7.5.0',\
+	com.influxdb:influxdb-client-java;version='7.5.0',\
+	com.influxdb:influxdb-client-utils;version='7.5.0',\
 	com.oracle.labs.olcut:olcut-config-protobuf;version='5.3',\
 	com.oracle.labs.olcut:olcut-core;version='5.3',\
 	com.squareup.okhttp3;version='5.3',\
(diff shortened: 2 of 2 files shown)


io.openems.edge.edge2edge/src/io/openems/edge/edge2edge/common/AbstractEdge2Edge.java
```diff
@@ -443,4 +444,18 @@ private void _readNatureStartAddresses(CompletableFuture<TreeMap<Integer, Short>
 							});
 				});
 	}
+
+	/**
+	 * Selects the appropriate converter for a given ModbusType.
+	 * @param type the type of the Modbus element
+	 * @return the converter
+	 */
+	private ElementToChannelConverter getConverterForType(ModbusType type) {
```
tsicking: You can make this method static:

```suggestion
	private static ElementToChannelConverter getConverterForType(ModbusType type) {
```

I would also put this method after `generateModbusElement()`.

io.openems.edge.edge2edge/src/io/openems/edge/edge2edge/common/AbstractEdge2Edge.java
```diff
@@ -443,4 +444,18 @@ private void _readNatureStartAddresses(CompletableFuture<TreeMap<Integer, Short>
 							});
 				});
 	}
+
+	/**
+	 * Selects the appropriate converter for a given ModbusType.
+	 * @param type the type of the Modbus element
+	 * @return the converter
+	 */
```
tsicking: The empty line before the `@param` tag is convention.

```suggestion
	/**
	 * Selects the appropriate converter for a given ModbusType.
	 * 
	 * @param type the type of the Modbus element
	 * @return the converter
	 */
```

io.openems.edge.edge2edge/src/io/openems/edge/edge2edge/common/AbstractEdge2Edge.java
```diff
@@ -246,7 +247,7 @@ private void mapRemoteChannels(TreeMap<Integer, Short> natureStartAddresses) thr
 					}
 
 					if (record instanceof ModbusRecordChannel r) {
-						m(r.getChannelId(), element);
+						m(r.getChannelId(), element, this.getConverterForType(record.getType()));
```
tsicking: If you make the method static as suggested below, you should remove the `this.`

```suggestion
						m(r.getChannelId(), element, getConverterForType(record.getType()));
```

io.openems.edge.edge2edge/src/io/openems/edge/edge2edge/common/AbstractEdge2Edge.java
```diff
@@ -443,4 +444,18 @@ private void _readNatureStartAddresses(CompletableFuture<TreeMap<Integer, Short>
...
 	}
+
+	/**
+	 * Selects the appropriate converter for a given ModbusType.
+	 * @param type the type of the Modbus element
+	 * @return the converter
+	 */
+	private ElementToChannelConverter getConverterForType(ModbusType type) {
+		return switch (type) {
+			case UINT16 -> ElementToChannelConverter.SET_NULL_FOR_DEFAULT(0xFFFF);				// 65535
```
tsicking: You can add the converters as static import, as it is mostly done in OpemEMS recently.

tsicking: I wonder if there is any way to use `ModbusRecordUint16.UNDEFINED_VALUE` instead of `0xFFFF` as I suggested in the forum, but it's not too important.

tsicking (changes requested):
Thanks for the helpful PR. I have some comments, mostly of code style nature.

io.openems.edge.edge2edge/src/io/openems/edge/edge2edge/common/AbstractEdge2Edge.java
tcdrop: done

io.openems.edge.edge2edge/src/io/openems/edge/edge2edge/common/AbstractEdge2Edge.java
tcdrop: done

io.openems.edge.edge2edge/src/io/openems/edge/edge2edge/common/AbstractEdge2Edge.java
tcdrop: done

io.openems.edge.edge2edge/src/io/openems/edge/edge2edge/common/AbstractEdge2Edge.java
tcdrop: done

io.openems.edge.edge2edge/src/io/openems/edge/edge2edge/common/AbstractEdge2Edge.java
tcdrop: thanks for your feedback! I tried to do that but I couldn't find a nice way of getting the byte array `ModbusRecordUint16.UNDEFINED_VALUE` to work with this code. 
Also I ran into problems with casting between unsigned types Uint16-64 and signed Java types Short, Int, Long. Having it as 0xFFFF worked.

parapluplu:
Is it feasible to add some kind of unit test for this?

sfeilmeier:
I used a different approach, reusing the existing UNDEFINED_VALUEs in https://github.com/OpenEMS/openems/pull/3479/commits/41db62190553d539d4ebed1e78c9ae725131b7fb.

I added JUnit tests, but did not test it in production yet.

Could anybody test it?

tsicking:
> I used a different approach, reusing the existing UNDEFINED_VALUEs in [41db621](https://github.com/OpenEMS/openems/commit/41db62190553d539d4ebed1e78c9ae725131b7fb).
> 
> I added JUnit tests, but did not test it in production yet.
> 
> Could anybody test it?

Hi, I've just tested it locally on my computer the following way:
1. Changed some records in `SymmetricEss.getModbusSlaveNatureTable()` to `UINT16`, `UINT32`, `UINT64`, namely Min/Max Cell Voltage/Temperature: 
```java
public static ModbusSlaveNatureTable getModbusSlaveNatureTable(AccessMode accessMode) {
		return ModbusSlaveNatureTable.of(SymmetricEss.class, accessMode, 100) //
				.channel(0, ChannelId.SOC, ModbusType.UINT16) //
				.channel(1, ChannelId.GRID_MODE, ModbusType.UINT16) //
				.channel(2, ChannelId.ACTIVE_POWER, ModbusType.FLOAT32) //
				.channel(4, ChannelId.REACTIVE_POWER, ModbusType.FLOAT32) //
				.channel(6, ChannelId.MIN_CELL_VOLTAGE, ModbusType.FLOAT32) //
				.channel(8, ChannelId.MAX_CELL_VOLTAGE, ModbusType.UINT64) //
				.channel(12, ChannelId.MIN_CELL_TEMPERATURE, ModbusType.UINT32) //
				.channel(14, ChannelId.MAX_CELL_TEMPERATURE, ModbusType.UINT16) //
				.channel(15, ChannelId.CAPACITY, ModbusType.FLOAT32) //
				.channel(17, ChannelId.MAX_APPARENT_POWER, ModbusType.FLOAT32) //
				.build();
	}
```
2. Activate a simulated ESS, where these channels remain undefined.
3. Activate Controller Modbus TCP with the simulated ESS (on Port 10502).
4. Activate Edge2EdgeEss to myself, i.e. with Modbus bridge to 127.0.0.1:10502.

Strangely, it worked for `UINT16` and `UINT64`, but not for `UINT32`. I got this log messsage:

```
2026-01-14T11:23:02,531 [modbus0 ] ERROR [dbus.api.task.AbstractReadTask] Parsing Response failed. IllegalArgumentException: Conversion for [MIN_CELL_TEMPERATURE] failed
```

The same happened in my previous test, as I described here: https://community.openems.io/t/edge-2-edge-ess-ungultige-channel-werte/10654/6. I have no idea why it doesn't work for 32 bit, but for 16 and 64 bit it works. Maybe it has to do with the fact that Java doesn't have unsigned data types?

Best regards,
Thomas

sfeilmeier:
> I have no idea why it doesn't work for 32 bit, but for 16 and 64 bit it works. Maybe it has to do with the fact that Java doesn't have unsigned data types?

Thank you @tsicking. I built for myself the same test environment and was able to fix the issues this way. It was indeed because of some wrong handling of unsigned types. The test is now successful for me, so I am merging it.

Sn0w3y:
Just FYI:

Java has **no unsigned primitive types**, that’s basically the whole problem here.

| Modbus | Max value | Stored as | Max signed value | Result |
|--------|-----------|-----------|------------------|--------|
| UINT16 | 65535 | `short` | 32767 | ❌ overflow > 32767 |
| UINT32 | 4294967295 | `int` | 2147483647 | ❌ overflow > ~2.1B |
| UINT64 | 18.4e18 | `long` | 9.2e18 | ⚠️ only breaks for huge values |

### Why UINT16 looked ok
Test values (SOC, temps, etc) were just small enough to never trigger overflow.  
`0xFFFF` (65535) actually turns into `-1` when stored as signed short, but we didn’t hit that in tests.

### Why UINT32 fails
`0xFFFFFFFF` stored in signed `int` → `-1`  
This then blows up later with `IllegalArgumentException`.

### Why UINT64 “works”
Already using `Long` and values are way below `Long.MAX_VALUE`.  
`0xFFFFFFFFFFFFFFFF` also becomes `-1`, but this case was already handled.

**TL;DR:** UINT32 was the first one where the overflow actually mattered.

PR #3483, opened 2025-12-17
[EDGE] Adds new Power-Control-Unit (PCU) nature

mbreier-sw:
Just to provide an interface to "click in" a power-plant-controller beside a grid-meter, an ESS and a PV-Inverter into an OpenEMS controller


parapluplu:
could you elaborate what you're trying to achieve with this component? We've integrated an Phoenix EZA Regler into our system as well, but mainly did that via the modbus api controller, as all desired data is exchangeable via that interface with an ESS configured as part of it. We've also thought about an API, but didn't followed up on that yet, as we first wanted to finalize the integration to see, what we can derive from that.

I guess your intention is monitoring of the EZA controller? If so I guess the current active Q mode is interesting as well.
All other values are basically either already in the meter API or the ESS API.

Additionally there are some values that are interesting, depending what you want to do with the EZA regler, which is also the reason, why I'm uncertain if it makes sense to put them into an API:
 - third party power (e.g. from a trader)
 - FCR power
 - AFFR power
 - power limitations given by the grid operator

mbreier-sw:
Dear @parapluplu,

it's not only monitoring the EZA controller. In our base scenario we'd like to run three EZA controller running on a PLCnext device of Phoenix Contact. The EZA controllers are used like follows:

* One EZA Controller is monitoring the grid feed-in point, listening to power limitations of network operator and/or direct marketers.
* Another EZA Controller is monitoring one or more BESS, listening to commands from the EMS and providing feedback to the EMS.
* Third EZA Controller is managing one or more PV-Inverter, listening to command from the EMS. 

We have build a modularized driver communicating with the PLCnext platform over HTTPS and the Phoenix REST-API. We'd like to connect these three EZA controllers within OpenEMS to manage a complex setup using OpenEMS including direct marketers and network operator. Each EZA Controller should be connected to network operator / direct marketing, or BESS or PV-Inverter.

Of course there will be other interesting values, only the defined channels got fixed up to now. We are trying to get community feedback as fast as possible to adjust the plan when necessary.

sfeilmeier:
There is a discussion on that topic in the Community forum: https://community.openems.io/t/nature-api-fur-eza-regler/10668/3. Let's continue over there. I am setting this PR to draft for now.

mbreier-sw:
> Just to provide an interface to "click in" a power-plant-controller beside a grid-meter, an ESS and a PV-Inverter into an OpenEMS controller

During Q1/2026 hackathon we noticed that it might be good to have a Power-Control-Unit API to simplify the handling of more than five different interfaces used by different Grid-Ops to limit power import/export or reactive power. It provides more flexibility than integrating the functionality into the core `Meta` component.

PR #3490, opened 2025-12-19
Update architecture.adoc: fixed ul within section process image

sjjh:
added newline to fix unordered list in section "6. Process Image". Also typographically corrected the dashes in the last sentence of the section.

Should be a fix trivial to merge. :)


sfeilmeier (approved):
Thank you! 👍

PR #3491, opened 2025-12-24
[Edge] Add support for SDM72D-M-2 meter

Sn0w3y:
This commit updates the Eastron SDM630 meter implementation to include support for the SDM72D-M-2 meter. Changes were made in both the readme documentation to list the SDM72D-M-2 as a compatible device and in the Config.java to reflect this compatibility in the description. This enhancement allows users of the SDM72D-M-2 meter to utilize the existing SDM630 implementation, broadening the applicability of the driver.


sfeilmeier:
It looks like the protocol could be the same for all three-phase meters?! If that's true, we could rename the packages and Config descriptions accordingly - see SOCOMEC implementation.

For now this is ok here and I'll merge. Thanks! 👍

PR #3499, opened 2026-01-04
pvinverter: Kostal Piko Corrected dayYield from kWh to Wh and corrected Units of totalYield a…

sjjh:
…nd dayYield


sjjh:
Currently the daily sum in OpenEMS UI is cut off as int. For example, a daily yield of *0.16 kWh* is displayed as *0 kWh*. 
The webUI of the inverter shows the daily yield correct:
grafik

I recognized in the code base, that `totalYield` was multiplied with '1000', but `dailyYield` was not. Thus changed this. I then refactored the cast-method to include the multiplication to prevent duplicated code and be consistent with parsing the miliamps.
I further noticed, that the `ChanelId`-`Enum` still used `KILOWATT_HOURS`, but in my understanding we are using `WATT_HOURS`, thus changed it there as well.
@Sn0w3y can you please check if it makes sense what I did.

sjjh:
Thx for the hint. Corrected it. But messed up my local git, thus need some time to push the commits. Will notify once it is ready.

sjjh:
I tried to sort it out and hope to not have messed up the complete git history. The build is still failing, but this should not be related to my commits?

sjjh:
As I messed up the commit history here, I'll close this and opened the new PR #3538.

PR #3505, opened 2026-01-07
[EDGE] Add tests and update Victron components

Sn0w3y:
This commit introduces several changes across the Victron components and their tests. Specifically, it adds new test configurations and test cases for Victron battery inverter, AC in/out meters, and PV inverter via DC charger components. Additionally, it updates the VictronBatteryInverterImpl and VictronEssImpl classes to improve event handling and reference policy configurations.

See: https://github.com/OpenEMS/openems/pull/3457#issuecomment-3694298949
and: https://github.com/OpenEMS/openems/pull/3457


Sn0w3y:
@GrisuBerlin and @DerWahreKlinki could you check if everything still works?

DerWahreKlinki:
BatteryInverter never gets a reference to battery so (battery-)limits cannot be calculated. Maybe you/we should transfer hardware checks to the Ess component?

PR #3509, opened 2026-01-08
GoodWe: Fix battery charging with PV when constraint is active

iseeberg79:
**Problem**

  When an external controller (e.g., evcc) sets a discharge constraint via setActivePowerLessOrEquals(0) to keep the battery neutral, the GoodWe handler incorrectly charged the battery with available PV power instead of allowing PV to supply household consumption first.

  **Root Cause**

  In handleRemoteMode(), when activePowerSetPoint == 0 and PV was available, the condition pvProduction >= activePowerSetPoint (e.g., 300 >= 0) triggered CHARGE_BAT mode, routing PV to battery instead of household loads.

  **Fix**

  Added explicit handling for constraint case (activePowerSetPoint == 0) to return DISCHARGE_BAT, 0 (neutral), allowing PV to serve household consumption before any battery interaction.

  **Test Coverage**

  - Fixed semantically incorrect test case
  - Added CHARGE_BAT test with proper PV surplus calculation
  - Added constraint test case (activePowerSetPoint == 0)
  - Added real-world evcc scenario test (GitHub evcc#22827)

  **Reference**

  - evcc GitHub Discussion: https://github.com/evcc-io/evcc/discussions/22827

### DIFF
--- a/io.openems.edge.goodwe/src/io/openems/edge/goodwe/common/ApplyPowerHandler.java
+++ b/io.openems.edge.goodwe/src/io/openems/edge/goodwe/common/ApplyPowerHandler.java
@@ -138,6 +138,12 @@ public final class ApplyPowerHandler {
 	private static Result handleRemoteMode(int activePowerSetPoint, int pvProduction) {
 		// TODO PV curtail: (surplus power == setpoint && battery soc == 100% => PV
 		// curtail)
+
+		// Constraint case: keep battery neutral (e.g., external controller constraint)
+		if (activePowerSetPoint == 0) {
+			return new Result(EmsPowerMode.DISCHARGE_BAT, 0);
+		}
+
 		if (activePowerSetPoint < 0) {
 			return new Result(EmsPowerMode.CHARGE_BAT, activePowerSetPoint * -1 + pvProduction);
 		}
--- a/io.openems.edge.goodwe/test/io/openems/edge/goodwe/common/ApplyPowerHandlerTest.java
+++ b/io.openems.edge.goodwe/test/io/openems/edge/goodwe/common/ApplyPowerHandlerTest.java
@@ -96,9 +96,9 @@ public class ApplyPowerHandlerTest {
 				/* maxAcExport */ 5000, //
 				/* surplusPower */ 105));
 
-		// SMART - CHARGE_BAT
-		assertResult(EmsPowerMode.CHARGE_BAT, 300, ApplyPowerHandler.calculate(//
-				/* activePowerSetPoint */ 0, //
+		// SMART - Surplus: activePowerSetPoint == surplusPower → AUTO (balanced)
+		assertResult(EmsPowerMode.AUTO, 0, ApplyPowerHandler.calculate(//
+				/* activePowerSetPoint */ 105,
 				/* pvProduction */ 300, //
 				ControlMode.SMART, //
 				/* gridActivePower */ 250, //
@@ -107,7 +107,29 @@ public class ApplyPowerHandlerTest {
 				/* maxAcExport */ 5000, //
 				/* surplusPower */ 105));
 
-		// SMART - null
+		// SMART - CHARGE_BAT: Battery charges with PV surplus
+		assertResult(EmsPowerMode.CHARGE_BAT, 2000, ApplyPowerHandler.calculate(//
+				/* activePowerSetPoint */ 1000,
+				/* pvProduction */ 3000,
+				ControlMode.SMART, //
+				/* gridActivePower */ 500,
+				/* essActivePower */ 300, //
+				/* maxAcImport */ 5000, //
+				/* maxAcExport */ 5000, //
+				/* surplusPower */ 0));
+
+		// SMART - Constraint: activePowerSetPoint == 0 → battery neutral
+		assertResult(EmsPowerMode.DISCHARGE_BAT, 0, ApplyPowerHandler.calculate(//
+				/* activePowerSetPoint */ 0,
+				/* pvProduction */ 300, //
+				ControlMode.SMART, //
+				/* gridActivePower */ 350, //
+				/* essActivePower */ 420, //
+				/* maxAcImport */ 5000, //
+				/* maxAcExport */ 5000, //
+				/* surplusPower */ 0));
+
+		// SMART - Missing data falls back to AUTO
 		assertResult(EmsPowerMode.AUTO, 0, ApplyPowerHandler.calculate(//
 				/* activePowerSetPoint */ 1000, //
 				/* pvProduction */ 300, //
@@ -236,5 +258,33 @@ public class ApplyPowerHandlerTest {
 		assertTrue(noSmartMeterDetected.get());
 		assertEquals(5300, emsPowerSet.get());
 		assertEquals(EmsPowerMode.CHARGE_BAT, emsPowerMode.get());
+	}
+
+	@Test
+	public void testConstraintedRealWorldScenario() throws OpenemsNamedException {
+		// Real-world scenario: evcc storage lock with EV charging (GitHub evcc-io/evcc #22827)
+		// PV: 5000W, Consumption: 12200W, Loadpoint-Consumption: 7200, Constraint: battery not discharging
+
+		// REMOTE mode: battery stays neutral
+		assertResult(EmsPowerMode.DISCHARGE_BAT, 0, ApplyPowerHandler.calculate(//
+				/* activePowerSetPoint */ 0,
+				/* pvProduction */ 5000, //
+				ControlMode.REMOTE, //
+				/* gridActivePower */ 12200, //
+				/* essActivePower */ -5000, //
+				/* maxAcImport */ 20000, //
+				/* maxAcExport */ 20000, //
+				/* surplusPower */ 0));
+
+		// SMART mode: falls through to REMOTE, same result
+		assertResult(EmsPowerMode.DISCHARGE_BAT, 0, ApplyPowerHandler.calculate(//
+				/* activePowerSetPoint */ 0,
+				/* pvProduction */ 5000, //
+				ControlMode.SMART, //
+				/* gridActivePower */ 12200, //
+				/* essActivePower */ -5000, //
+				/* maxAcImport */ 20000, //
+				/* maxAcExport */ 20000, //
+				/* surplusPower */ 0));
 	}
 }


iseeberg79:
@sfeilmeier, I would appreciate if you could review this. I've analyzed the issues related to evcc usage, referencing both the KACO implementation (https://community.openems.io/t/setactivepowerlessorequals-verhalten/10697/7) and the GoodWe (FENECON) implementation (already referenced). Sadly I've not found a solution for the KACO device as it looks like it's not about the implementation by OpenEMS (probably related to firmware/external library?). I cannot fiddle out the reason for PV routing for those.

It's related to the prevent discharge function implemented by the opemens template in evcc.

Because I have no hardware to test,  you need to take it as theoretical work and someone else needs to the testing. It's additional unit tests are based on real-world scenarios and proove to solve the issue. Users having issues are FEMS users and because of this a test is not possible easily.

iseeberg79:
@sfeilmeier friendly reminder to discuss?

Sn0w3y:
Thanks for the analysis, but I believe the current behavior is actually correct.

  activePowerSetPoint represents the desired AC output power of the ESS - it does not mean "battery power". When activePowerSetPoint == 0, it means: "The ESS should produce 0W on the AC side." If PV is producing 300W and 0W should go to AC, then charging the battery with the surplus 300W is the correct behavior. The PV energy has to go somewhere.

  The existing logic in handleRemoteMode is consistent:

  pvProduction (300) >= activePowerSetPoint (0)  →  CHARGE_BAT, pvProduction - activePowerSetPoint = 300

  This means: "Feed activePowerSetPoint to AC, charge the battery with the rest of PV." That's exactly what should happen.

  Your proposed fix (DISCHARGE_BAT, 0) would leave the 300W PV production unaccounted for - the behavior then depends entirely on how the GoodWe inverter internally handles that case, which is undefined.

  If evcc wants to keep the battery neutral while still using PV for household consumption, the constraint needs to be expressed differently - this is not a bug in handleRemoteMode, but a semantic mismatch in how the constraint is set by the external controller.

iseeberg79:
Thanks for your feedback. Yes, I realized from the discussion which started afterwards.

Let's not think about evcc here too much. Of course I'd appreciate a solution to successfully avoid discharging the battery on those external loads. It is clear to me that the use case needs to be resolved. That's another discussion ;-)

I would appreciate if somebody, best those who implemented the case, could clarify the reason for and take the proposal into account. I also see a dependency how the device is handling energy of course. Without a hardware and clear understanding about how it is supposed to get controlled this cannot be decided.

iseeberg79:
To be clearer, I was absolutely sure this is a bug. 

This behavior is to be avoided:
image

Sn0w3y:
> To be clearer, I was absolutely sure this is a bug. 
> 
> This behavior is to be avoided:
> image

AFAIK it is WANTED, that the ess charges the battery with PV-Power. If you see the ctrlGridOptimizedCharge for example - rhere you can see how it is done. This beavior is NOT a bug in GoodWe Implementation how to regulate Battery Power.

iseeberg79:
Yes, but to charge the battery is not the right decision here. There might be an other optimizer to control the route, but there is a better route for PV power quite obviously. 

It's not related to my personal situation but others aren't happy about...

To let others decide is the reason for draft status. I just wanted to point out an issue and suggest a solution: PR is to be closed if not wanted or not a proper solution.

iseeberg79:
AbstractGoodWe computes ACTIVE_POWER = PV + battery discharge (see updatePowerAndEnergyChannels()). This means setActivePowerLessOrEquals(0) — as REST API usage uses for its storage lock — constrains the combined AC output (PV + battery) to ≤ 0 W, not battery power alone.

Per that model, the pre-fix behaviour is internally consistent: with PV = 5000 W, activePowerSetPoint = 0 resolves to "AC output must be 0 W", and CHARGE_BAT, pvProduction achieves exactly that by absorbing all PV into the battery. The practical consequence, however, is that household consumption is served entirely from the grid while PV is charged into the battery — the opposite of what a storage lock is intended to achieve.

This fix changes the handling of activePowerSetPoint == 0 from "charge battery to cancel out PV" to "hold battery at 0 W". The AC constraint is technically no longer satisfied in terms of combined output (total AC = PV > 0 W), but the actual power flow — PV serves household loads, grid covers the remainder — matches user intent.

The architecturally correct resolution could be to expose the GoodWe as a HybridEss that reports battery power separately from PV, so that setActivePowerLessOrEquals for the ESS component targets battery power only. This fix intends to be a pragmatic interim solution for the DC-coupled case?

As concluded above, it's just a suggestion - to be decided as there is no API capable alternative, yet. 
I do not claim to have understood the complete picture...

Sn0w3y:
You're correct that the current behavior is internally consistent. `setActivePowerLessOrEquals(0)` constrains `ACTIVE_POWER <= 0`, and because `ACTIVE_POWER = PV + battery` for the GoodWe `HybridEss`, `CHARGE_BAT` with `pvProduction` naturally satisfies that constraint. So there isn’t actually a bug here.

You’re also right about the architectural side - and the solution is already implemented. GoodWe already implements `HybridEss` (`GoodWeEssImpl.java:54`) and exposes `DC_DISCHARGE_POWER` as a separate read channel that represents the real battery power. The `ACTIVE_POWER` channel documentation in `SymmetricEss` already clarifies this:

> “Discharge or charging Power (including DC-PV power, if applicable). For the actual charging or discharging power of the battery, please refer to address `ess0/DcDischargePower`.”

So the distinction between AC-side power and actual battery power already exists in the API.

Because of that, the proposed fix isn’t the right approach. Special-casing `activePowerSetPoint == 0` in `handleRemoteMode` would break the contract between the Power Solver and the ESS. The Solver calculates a setpoint that satisfies all constraints, and `ManagedSymmetricEss.applyPower` explicitly states that implementations should not modify those parameters. If the ESS silently reinterprets the setpoint, it could violate the constraints the Solver just resolved and potentially interfere with other controllers running in the same cycle.

There’s also already an established pattern in OpenEMS for targeting DC power. `Controller.Ess.FixActivePower` supports a `HybridEssMode.TARGET_DC` mode that performs exactly this translation (`ControllerEssFixActivePowerImpl.java:137–150`):

```java
case TARGET_DC ->
    switch (ess) {
    case HybridEss he -> {
        var pv = ess.getActivePower().orElse(0) - he.getDcDischargePower().orElse(0);
        yield pv + power;
    }
    default -> power;
    };
```

This logic reads the current PV production from the `HybridEss` channels and translates the desired DC (battery) power into the correct AC power constraint. In other words, the controller performs the DC-to-AC translation - not the inverter driver. That’s the established OpenEMS design pattern.

For the `evcc` use case (“battery neutral”), the same principle applies. The external controller - or a dedicated OpenEMS controller - would need to read the `HybridEss` PV production and set `setActivePowerEquals(pvProduction)` in order to achieve `dcDischargePower = 0`.

Given that, the GoodWe implementation is already behaving correctly, and this PR should probably be closed.

I also dug into how evcc actually talks to OpenEMS, and the real issue comes from evcc’s OpenEMS template. The problem is with the batterymode switch:

# case 1 (Normal) and case 2 (Hold/Prevent Discharge) are IDENTICAL:
body: '{"value": 0}'   →   SetActivePowerLessOrEquals = 0

For a HybridEss setup where ACTIVE_POWER equals PV plus battery, this means PV + battery ≤ 0. In practice, you end up forcing all PV into the battery. That’s exactly what your screenshot shows.

The real fix needs to happen in evcc’s template, not in the GoodWe driver. Here’s what to do:

When you want “Hold” mode (battery-neutral), evcc should first read the current PV production, then set ACTIVE_POWER equal to PV. This way, the battery’s contribution stays at zero watts. OpenEMS’s own Controller.Ess.FixActivePower does it the same.

In evcc terms, here’s how “Hold” mode (case 2) should work via REST:

Step 1: Read the current ESS values:
```
GET /rest/channel/ess0/ActivePower        → maybe 4800 W
GET /rest/channel/ess0/DcDischargePower   → maybe -200 W (negative means charging)

```
So, PV = ActivePower - DcDischargePower = 4800 - (-200) = 5000 W

Step 2: Set the AC power target to PV production (so battery stays at 0):

```
POST /rest/channel/ess0/SetActivePowerEquals
body: {"value": 5000}
```

This should keep the battery neutral and lets PV handle household loads, which is what you want (and what your screenshot shows as a problem).

For “Normal” mode (case 1 or watchdog reset), evcc shouldn’t write SetActivePowerLessOrEquals = 0. Instead, just release control by writing null:

```
POST /rest/channel/ess0/SetActivePowerLessOrEquals
body: {"value": null}
```

The Modbus template (openems-modbus.yaml) has this same bug with battery_set_register (register 710, which also maps to SetActivePowerLessOrEquals).

Bottom line: This is something that needs to be fixed on the controller side in evcc, not in the OpenEMS inverter driver. The GoodWe already does ACTIVE_POWER exactly as documented.

Sn0w3y:
> AbstractGoodWe computes ACTIVE_POWER = PV + battery discharge (see updatePowerAndEnergyChannels()). This means setActivePowerLessOrEquals(0) — as REST API usage uses for its storage lock — constrains the combined AC output (PV + battery) to ≤ 0 W, not battery power alone.
> 
> Per that model, the pre-fix behaviour is internally consistent: with PV = 5000 W, activePowerSetPoint = 0 resolves to "AC output must be 0 W", and CHARGE_BAT, pvProduction achieves exactly that by absorbing all PV into the battery. The practical consequence, however, is that household consumption is served entirely from the grid while PV is charged into the battery — the opposite of what a storage lock is intended to achieve.
> 
> This fix changes the handling of activePowerSetPoint == 0 from "charge battery to cancel out PV" to "hold battery at 0 W". The AC constraint is technically no longer satisfied in terms of combined output (total AC = PV > 0 W), but the actual power flow — PV serves household loads, grid covers the remainder — matches user intent.
> 
> The architecturally correct resolution could be to expose the GoodWe as a HybridEss that reports battery power separately from PV, so that setActivePowerLessOrEquals for the ESS component targets battery power only. This fix intends to be a pragmatic interim solution for the DC-coupled case?
> 
> As concluded above, it's just a suggestion - to be decided as there is no API capable alternative, yet. I do not claim to have understood the complete picture...

Nice AI Summary ! :) I summarized my whole Picture of the Code above and let AI rewrite it - maybe that approach is better to have this full picture :)

@iseeberg79

iseeberg79:
Yes, I let AI summarize, but analyzed and verified. The conclusion is something to avoid battery discharge being externally controlled is missing. Personally, after evaluation of the responses, I think it requires a controller.


It should be another issue and is getting OT here but to mention: Actually the implementation (template based) in evcc is not able to send dynamic values and take actual PV into account.
To adress openEMS and FEMS it requires the template to be changed.
Currently I have difficulties to describe the limitations and an alternative approach to the ESS power control would be appreciated (adressing FEMS and openEMS systems).

I think to open another issue is due.

Thanks for reviewing,
PR is to be closed.

Ingo

timo-schlegel:
@Sn0w3y 
This is a very good summary of the situation 👍

Since the discussion so far has mainly focused on **hybrid inverters**, I would like to add one small architectural point: since we are talking about a **generic EVCC interface**, it should ideally work for **both hybrid and AC-coupled storage systems**.

The statement itself is actually also correct for **AC-coupled storage systems**. The practical problem is just that **DcDischargePower** is not available there, which means we currently have to distinguish between the two cases.

Conceptually, the same formula still applies:

`ACTIVE_POWER = PV + BATTERY`

For **AC-coupled batteries**, the PV power at that inverter is always **0**, so it effectively becomes:

`ACTIVE_POWER = BATTERY`.

From the EVCC perspective it would therefore make sense to allow choosing between two modes:

* **TARGET_DC**
  → EVCC reports the **PV power as AC power target**

* **TARGET_AC**
  → EVCC actually writes **0 as AC power**

As a control target, the **ESS instance(s)** should be used (not the **ESS Cluster**), because hybrid and AC-coupled batteries can be operated **in parallel**.

Especially when **multiple batteries or inverters** are present, it is important to remember that the **PV value must always be the PV production of the respective ESS**, not a global summary. This is also consistent with the example mentioned above, where the **PV production of the ESS** is used.

---

**Alternative approach**

Another possible approach could be to use the **OpenEMS `FixActivePower` controller** for this purpose (one per ESS, configured accordingly with **TARGET_AC** or **TARGET_DC**) and let EVCC simply control whether the controller is **active or inactive**.

With this approach, **EVCC would not need to perform any calculations** and would also not need to distinguish between **TARGET_DC** and **TARGET_AC**. EVCC would simply send an **active/inactive state** to block or allow battery discharge.

From an architectural perspective this might actually be quite clean, because EVCC would not need to care whether the system is a **Hybrid ESS** or an **AC-coupled inverter** – it would simply control whether battery discharge should currently be allowed or blocked.

This would also allow controlling the **priority** and therefore whether EVCC setpoints may be **overridden by other controllers**, depending on the desired setup. Some users might explicitly want the possibility to override EVCC behavior with another controller.

The downside of this approach is that the corresponding `FixActivePower` controller(s) would need to be configured in OpenEMS (typically one per ESS).

A limitation of this approach is that EVCC could then only control whether battery discharge is blocked or not. More advanced behavior would require additional logic, which in this approach would need to be implemented on the OpenEMS side.

See also the related discussion:
[https://community.openems.io/t/setactivepowerlessorequals-verhalten/10697/13](https://community.openems.io/t/setactivepowerlessorequals-verhalten/10697/13)

iseeberg79:
> From an architectural perspective this might actually be quite clean, because EVCC would not need to care whether the system is a Hybrid ESS or an AC-coupled inverter – it would simply control whether battery discharge should currently be allowed or blocked.

That's was it basically requires, yes.

PR #3516, opened 2026-01-14
Fix: Handle undefined SOC in solver strategy "Optimize by keeping all near equal"

tsicking:
Previously, if the SOC of an ESS was undefined, the solver strategy `OPTIMIZE_BY_KEEPING_ALL_NEAR_EQUAL` couldn't handle it, leading to an error during `handleEvent()` in the power component.
This is fixed by extending the `try`-block, so that `NO_RESULT` is returned, and the solver tries the next strategy `OPTIMIZE_BY_KEEPING_ALL_EQUAL`. A JUnit test has been added.

### DIFF
--- a/io.openems.edge.ess.core/src/io/openems/edge/ess/core/power/optimizers/KeepAllNearEqual.java
+++ b/io.openems.edge.ess.core/src/io/openems/edge/ess/core/power/optimizers/KeepAllNearEqual.java
@@ -46,16 +46,16 @@ public class KeepAllNearEqual {
 			return NO_RESULT;
 		}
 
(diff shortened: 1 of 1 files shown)


DerStoecki (approved):

pooran-c:
@DerStoecki : Thanks for indentifying this, 
We could avoid many computations if just check for null after getting the essList.
"
var essList = getGenericEssList(esss);
if (essList.stream().anyMatch(ess -> ess.getSoc().get() == null)) {
      return NO_RESULT;
}
"
 Instead of the try catch generic block.

PR #3517, opened 2026-01-14
Shelly Firmware: Check if stable version update is present

sthhs:
It's confusing that OpenEMS shows an available firmware update for a Shelly plug if it's a beta version.
This modifications checks if a stable update is available and ignores beta.

### DIFF
--- a/io.openems.edge.io.shelly/test/io/openems/edge/io/shelly/shellyplusplugs/IoShellyPlusPlugSImplTest.java
+++ b/io.openems.edge.io.shelly/test/io/openems/edge/io/shelly/shellyplusplugs/IoShellyPlusPlugSImplTest.java
@@ -117,6 +117,22 @@ public class IoShellyPlusPlugSImplTest {
 
 						.output(IoShellyPlugSBase.ChannelId.HAS_UPDATE, false)) //
 
+				.next(new TestCase("No update available") //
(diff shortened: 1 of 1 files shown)


Sn0w3y:
It appears that this issue is not limited to a single Shelly device but affects multiple ones. If this change is truly required for clarity, it should be applied consistently across all Shelly integrations.

The flag has_update currently indicates whether an update is available at all. However, users will still verify in the Shelly device UI itself whether the update is a beta or a stable release.

To better distinguish both cases and avoid ambiguity, it might make sense to introduce an additional flag such as has_beta_update, so both stable and beta updates are clearly covered.

sfeilmeier:
For me the fix is in principle ok; this has also been reported as confusing by FENECON customers. 

Currently `available_updates` is not covered by a JUnit test (see https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.io.shelly/test/io/openems/edge/io/shelly/shellyplus1pm/IoShellyPlus1PmImplTest.java#L91-L92). Can you provide a real JSON response with `stable` or other values if you have?

sthhs:
> Can you provide a real JSON response with `stable` or other values if you have?

I've added the test to 
[io.openems.edge.io.shelly/test/io/openems/edge/io/shelly/shellyplusplugs/IoShellyPlusPlugSImplTest.java](url)

PR #3547, opened 2026-01-29
Fix Modbus task scheduling and simplify task handling

timo-schlegel:
As discussed and suggested in https://community.openems.io/t/asking-for-feedback-new-modbusworker-implementation/1747/18

The Modbus task handling did not behave as expected because the INITIAL_WAIT task waited for the full planned delay instead of reacting to the EXECUTE_WRITE event. Since it used WaitDelayTask instead of waitMutexTask, it could not be interrupted by the write event. As a result, the planned wait effectively ended after the WRITE event.

This caused write tasks to execute too late, while read tasks were executed immediately after the WRITE event. Consequently, the READ_BEFORE_WRITE and WAIT_FOR_WRITE tasks were never executed.

From a design perspective, the WRITE event should be processed as early as possible after the "execute write" trigger, while READ tasks should finish as late as possible before the "before process image" event. Therefore, the planned waiting time must only occur within the WAIT_BEFORE_READ task, between WRITE and READ. The INITIAL_WAIT task must wait for the EXECUTE_WRITE event instead of a fixed delay.

This commit simplifies the Modbus task handling by removing unused tasks and fixing the INITIAL_WAIT task to use an interruptible wait via waitMutexTask. The associated tests were updated accordingly to ensure consistent and predictable system behavior.


timo-schlegel:
Hi @da-Kai,

thanks for looking into this. The System.out logging has been removed from the tests.

PR #3554, opened 2026-02-02
Intellij + Gradle Setup Guide

alexbelkedev:
Hi! This PR adds an English IntelliJ + Gradle setup guide (AsciiDoc).
Could someone from the docs/IDE maintainers take a look and assign a reviewer if needed? Thanks!
I haven't deleted the old intelliJ Guide yet. I would prefer to have both for a while. After confirm that new setup works and everybody are happy, the old one can be removed.

@sfeilmeier


sfeilmeier (changes requested):
Please replace all "FEMS" with "OpenEMS"

alexbelkedev:
> Please replace all "FEMS" with "OpenEMS"

Good catch. Done.

PR #3562, opened 2026-02-06
Implement OpenDTU meter

sthhs:
This component implements an OpenDTU single phase meter. It sends http requests to an OpenDTU and evaluates power, voltage and current from the response.


io.openems.edge.meter.opendtu/src/io/openems/edge/meter/opendtu/Config.java
```diff
@@ -0,0 +1,36 @@
...
+	@AttributeDefinition(name = "Is enabled?", description = "Is this Component enabled?")
+	boolean enabled() default true;
+
+	@AttributeDefinition(name = "Phase", description = "Which Phase is measured by this Meter?")
+	SinglePhase phase() default SinglePhase.L1;
+
+	@AttributeDefinition(name = "Meter-Type", description = "Grid, Production (=default), Consumption")
+	MeterType type() default MeterType.PRODUCTION;
+
+	@AttributeDefinition(name = "IP-Address", description = "The IP address of the OpenWB.")
```
Sn0w3y: OpenDTU

io.openems.edge.meter.opendtu/src/io/openems/edge/meter/opendtu/Config.java
```diff
@@ -0,0 +1,36 @@
...
+	@AttributeDefinition(name = "Alias", description = "Human-readable name of this Component; defaults to Component-ID")
+	String alias() default "";
+
+	@AttributeDefinition(name = "Is enabled?", description = "Is this Component enabled?")
+	boolean enabled() default true;
+
+	@AttributeDefinition(name = "Phase", description = "Which Phase is measured by this Meter?")
+	SinglePhase phase() default SinglePhase.L1;
+
+	@AttributeDefinition(name = "Meter-Type", description = "Grid, Production (=default), Consumption")
```
Sn0w3y: Should be fixed in this case, as OpenDTU only's Use-Case is Production.

io.openems.edge.meter.opendtu/src/io/openems/edge/meter/opendtu/Config.java
```diff
@@ -0,0 +1,36 @@
...
+	@AttributeDefinition(name = "Component-ID", description = "Unique ID of this Component")
+	String id() default "meterOpenDTU0";
+
+	@AttributeDefinition(name = "Alias", description = "Human-readable name of this Component; defaults to Component-ID")
+	String alias() default "";
+
+	@AttributeDefinition(name = "Is enabled?", description = "Is this Component enabled?")
+	boolean enabled() default true;
+
+	@AttributeDefinition(name = "Phase", description = "Which Phase is measured by this Meter?")
```
Sn0w3y: "... is this Inverter Connected to" would be better in this case i guess

io.openems.edge.meter.opendtu/src/io/openems/edge/meter/opendtu/MeterOpenDtuImpl.java
```diff
@@ -0,0 +1,222 @@
...
+		}
+	}
+
+	private void processHttpResult(HttpResponse<JsonElement> result, Throwable error) {
+		setValue(this, MeterOpenDtu.ChannelId.SLAVE_COMMUNICATION_FAILED, result == null);
+
+		Integer power = null;
+		Integer voltage = null;
+		Integer current = null;
+		//boolean restartRequired = false;
```
Sn0w3y: unrelated

io.openems.edge.meter.opendtu/src/io/openems/edge/meter/opendtu/MeterOpenDtuImpl.java
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	@Override
+	public Timedata getTimedata() {
+		return this.timedata;
+	}
+
+	@Override
+	public MeterType getMeterType() {
+		return this.config.type();
```
Sn0w3y: fixed to production would be better imho

io.openems.edge.meter.opendtu/test/io/openems/edge/meter/opendtu/MeterOpenDtuImplTest.java
```diff
@@ -0,0 +1,25 @@
...
+import io.openems.edge.common.test.AbstractComponentTest.TestCase;
+import io.openems.edge.common.test.ComponentTest;
+import io.openems.edge.common.type.Phase.SinglePhase;
+
+
+public class MeterOpenDtuImplTest {
+
+	@Test
+	public void test() throws Exception {
+		new ComponentTest(new MeterOpenDtuImpl()) //
```
Sn0w3y: please extend the tests with actual json data

Sn0w3y (dismissed):
Thank you! Generally solide! I still had some Comments.

Please see my PR (outdatet) - we already tried to implement it a longer Time ago - @DerWahreKlinki and me. Our implementation also supported Setting Power Limits (required in some cases).

Maybe we should consider to take the "best of two Worlds" and merge our Codes or join Forces and Improve both :)

Sn0w3y:
@sthhs what do you think of my suggestion?

sthhs:
> @sthhs what do you think of my suggestion?

When should the limit be set? Only once at the first connection? Or should it be regularly requested and set if the received value doesn't match the target in OpenEMS?

My point of view with this component is: 

- Keep it as simple as possible. If someone uses an OpenDTU, he should be able to configure the inverter from there. The limit may not be changed very frequently. 
- If the API changes, this doesn't affect the OpenEMS component.
- Without the ability of writing values to an OpenDTU there is no need for storing credentials and therefore no possible security risk, even if this would be very low.

Sn0w3y:
IMHO every time it changes (see other Components) as in Germany (as you know) there is a 800W Limit for Balcony Solar and many of the Customers want to set it. Would be nice if OpenEMS could do so aswell.

The limit can and should be changed frequently (see Dynamic Limit in OpenDTU) - also because of Zero-Feed-in use case where ine wants to use 100% (lets say 1000W) of his Balcony Solar but not Feed into Grid as this would again not be allowed.

Well if we see it from an API perspective - we have many many Components which depend on APIs/Registers and so on - why should we make an Exception on the OpenDTU?

We handle Security pretty well in OpenEMS (Backend Communication, Api-Keys, etc.) since the communication is local i do not see any problems in that tbh

ChristianSill:
As far as I know, a reversible software-side throttling of the WR to 800 W does not fall into the balcony solar category. If you want to operate it in accordance with the rules. On the other hand, a factory throttling by the manufacturer is permissible. But I like to be taught a better one.

Sn0w3y:
> As far as I know, a reversible software-side throttling of the WR to 800 W does not fall into the balcony solar category. If you want to operate it in accordance with the rules. On the other hand, a factory throttling by the manufacturer is permissible. But I like to be taught a better one.

I still believe it absolutely makes sense to include the dynamic limit feature in this component.

OpenEMS is an Energy Management System, not just a passive monitoring layer. If we only read values from OpenDTU, we reduce this integration to a pure meter adapter. One of the core purposes of OpenEMS is to actively control and optimize energy flows — and dynamic power limiting is a fundamental EMS use case (zero-feed-in, self-consumption optimization, grid constraint handling, etc.).

Regarding the 800 W balcony solar discussion: the regulatory requirement is that no more than 800 W AC are fed into the grid. A dynamic EMS-based limit fully respects that physical constraint. In practice, many installations already rely on software-based caps or dynamic limiting (especially for zero-feed-in scenarios). This is not an exotic edge case but a real-world requirement.

Technically, dynamic limiting is superior to static manufacturer throttling because it allows full self-consumption while preventing grid feed-in. A fixed 800 W factory limit wastes potential if local load is higher. With EMS control, we can adapt in real time — which is exactly what OpenEMS is built for.

Security-wise, this does not introduce a new risk class. We already store credentials and write to APIs/Modbus devices in many components. Communication is local, and write access can be made optional via configuration. Read-only mode can remain the default to keep things simple.

So this is not about overengineering — it’s about enabling proper EMS functionality. We can keep the implementation clean and optional, but removing write capability would unnecessarily limit the integration and contradict the overall design philosophy of OpenEMS.

sthhs:
How ist this solved for other PV inverters? I've been searching for the corresponding code in the controllers of Fronius or Kostal inverters and they only seem to request data from them, but don't send anything for setting the power limit. Maybe I haven't found it, or does it not exist?
Sending a Post request to the OpenDTU is quite simple, but there seems to be done much more for giving OpenEMS the possibility to set the limit dynamically dependant on the current feed strategy.

Sn0w3y:
Another thought of mine:

- We should check, if there is a way to Implement it, so multiple Inverters can be fetched at once with one call as the DTU is running on an ESP32 it should not be fetched 5 times per second (1000ms Cycle).
- There is an Endpoint http://<dtu-ip>/api/livedata/status which gives us the Info of all connected Inverters to the DTU so we do not need 5 Components if there are 5 Inverters connected

TBD: We should also check, if there is a possibility to then still see 5 Components in the UI (@sfeilmeier is this possible, if there is only one Component activated?)

JFYI: My Implementation of the DTU is running smoothly aswell with the Feature of the Limit Setting - but i discovered Issues after some Runtime because of the multiple Requests made from OpenEMS'es side (i have 5 Inverters).

sfeilmeier:
> TBD: We should also check, if there is a possibility to then still see 5 Components in the UI (@sfeilmeier is this possible, if there is only one Component activated?)

I did not read through everything - but no, the current UI strictly expects separate Components. This is also required to use standard Channel-IDs like "ActivePower" per component

timo-schlegel:
I believe it should be possible (and beneficial) to introduce a central core component that handles communication/data retrieval, and then have multiple ElectricityMeter components that consume their data from this shared core.

A similar architectural pattern is already used in the Kaco implementation, specifically in
OpenEMS Kaco Blueplanet Hybrid10 ESS,
where a central component manages the communication, while ESS and Charger components derive their data from it.

From my perspective, this approach would bring several advantages:
	•	Avoid duplicated communication connections
	•	Clear separation of concerns (protocol handling vs. domain components like meters)
	•	Better extensibility (e.g. adding additional meter instances becomes straightforward)

On the other hand, a downside would be that for simple use cases (e.g. a single inverter), users would need to configure two components instead of one, which adds some complexity.

Would it make sense to align this implementation with that pattern, or is the additional configuration overhead considered too high?

sthhs:
In my understanding of the Kaco approach, the core is responsible for one device that can consist of several components, but all of different types and only one of each type. The serial number is configured in the core and not in the other components. So there is only one physical device with exactly one serial number handled by one core.

With the OpenDTU there are multiple identical device types with different serial numbers to communicate with.
Since I haven't found a way to request the status of all connected inverters from the same OpenDTU in a single request I don't think that this is a useful approach. The inverter serial number is a mandatory parameter in the request and it's not possible to attach more than one. So the number of requests will not be reduced, but instead the complexity will be higher. 
I don't know if it's possible at all in OpenEMS for multiple components of the same type to use the same core component and tell it dynamically the serial numbers of the inverters that have to be requested.
So I don't see a benefit splitting the component into a core and a meter/inverter part.

timo-schlegel:
That should in principle be possible by implementing something like an addInverter(serialNumber) method in the core component, which is then called by the ElectricityMeter/PVInverter.

However, if the implementation still requires sending dedicated requests per inverter, then the core component approach would not provide any real benefit and would only introduce additional complexity.

My intention was mainly to point out that, in OpenEMS, it is generally possible to have multiple components while retrieving all data via a shared communication layer (i.e. such a core component).

If OpenDTU does not provide the data in a way that allows a single combined query, then this approach would not add value in this specific case.

io.openems.edge.meter.opendtu/src/io/openems/edge/meter/opendtu/MeterOpenDtuImpl.java
```diff
@@ -0,0 +1,221 @@
...
+		Integer current = null;
+
+		if (error != null) {
+			this.logDebug(this.log, error.getMessage());
+
+		} else {
+			try {
+				var jsonResponse = getAsJsonObject(result.data());
+				var inverters = JsonUtils.getAsJsonArray(jsonResponse, "inverters");
+				var inverter = inverters.get(0).getAsJsonObject();
```
sfeilmeier: This uses the internal `getAsJsonObject()` from GSON, which throws an `IllegalArgumentException`. Better replace with OpenEMS JsonUtils or JsonPath.

io.openems.edge.meter.opendtu/readme.adoc
```diff
@@ -0,0 +1,5 @@
+= OpenDTU Meter
+
+This bundle implements requesting power, voltage and current from an OpenDTU.
```
sfeilmeier, line 3: Some info on how to use this bundle with OpenDTU would we very helpful

io.openems.edge.meter.opendtu/test/io/openems/edge/meter/opendtu/MeterOpenDtuImplTest.java
```diff
@@ -0,0 +1,119 @@
...
+						.output(ElectricityMeter.ChannelId.VOLTAGE_L3, null) //
+						.output(ElectricityMeter.ChannelId.FREQUENCY, null) //
+						.output(ElectricityMeter.ChannelId.ACTIVE_POWER_L1, 123) //
+						.output(ElectricityMeter.ChannelId.ACTIVE_POWER_L2, null) //
+						.output(ElectricityMeter.ChannelId.ACTIVE_POWER_L3, null) //
+						.output(ElectricityMeter.ChannelId.REACTIVE_POWER, null) //
+						.output(ElectricityMeter.ChannelId.REACTIVE_POWER_L1, null) //
+						.output(ElectricityMeter.ChannelId.REACTIVE_POWER_L2, null) //
+						.output(ElectricityMeter.ChannelId.REACTIVE_POWER_L3, null) //
+						.output(ElectricityMeter.ChannelId.ACTIVE_PRODUCTION_ENERGY, null) //
```
sfeilmeier, line 107: Many channels are not yet filled. I made this visible by setting `activateStrictMode()`

sfeilmeier (approved):
Very solid work! Thank you!

sthhs:
Thank you, but unfortunately the production meter does not show up in the production widget of the OpenEMS monitoring live view.
The log shows that the communication is working: 
`meterOpenDTU0[L2:4 W]`
And the total production sum contains this amount, too.

If I make the meter type configurable in Config.java and set it to production, it appears.

sfeilmeier:
Yes, this is a known problem with osgi not forwarding these properties to EdgeConfig and then to UI. Providing a config property is a good workaround. Or manually adding to "isProducer()" in UI.

sthhs:
> Providing a config property is a good workaround. Or manually adding to "isProducer()" in UI.

Which one do you prefer? The first one as already finished and I could put it in a new PR.

sfeilmeier:
Option 3: implement    io.openems.edge.pvinverter.api.ManagedSymmetricPvInverter. That's covered in isProducer.

sthhs:
Thank you, here it is: [https://github.com/OpenEMS/openems/pull/3709](https://github.com/OpenEMS/openems/pull/3709)

PR #3569, opened 2026-02-18
fix: Resolve title disappear from the header on first load.

Jasonlee6789:
Problem:
Title( e.g. OpenEMS-UI). disappear from the header on first load. 

Root cause:
Upstream removed GlobalRouteChangeHandler injection from AppComponent, making the route-based title mechanism dead code. Additionally, the live route changed from component to loadChildren, introducing a child route layer without explicit data.
The key issue was that route.data || data  would incorrectly treat empty objects {} as truthy, causing the system to use empty route data instead of falling back  to previous data or the environment title. The fix logic properly identifies empty objects and continues the fallback chain.  

Since I noticed the upstream team intentionally removed the GlobalRouteChangeHandler injection from AppComponent, if upstream prefer to keep it removed, I have an alternative fix with fewer code changes: simply add `setCurrentComponent("", this.route) `in LiveComponent ionViewWillEnter(), which also resolves this issue.

Test:
Before
image
 
After
image


Sn0w3y:
Thank you! I have been looking into this for days aswell :D

lukasrgr:
thanks for your contribution 🚀

PR #3571, opened 2026-02-18
Add support for SolarEdge ESS PV hybrid inverters

timo-schlegel:
Adds compatibility for SolarEdge ESS PV hybrid inverters (single-phase and three-phase models).
E.g. SolarEdge SE4000 with SESTI interface or SolarEdge SE10K-RWS.


timo-schlegel:
@sfeilmeier, as requested at https://community.openems.io/t/asking-for-feedback-new-modbusworker-implementation/1747/6?u=mrt, please find my SolarEdge ESS implementation here.

The Checkstyle test fails due to incorrect naming in `SolarEdgeSunSpecModel`, which uses the same naming as `DefaultSunSpecModel`. The purpose of `SolarEdgeSunSpecModel` is to provide the `DefaultSunSpecModel` without events, as these cause warnings when using the SunSpec model with SolarEdge (either with my implementation or with the included SolarEdge PV inverter implementation). Would it make sense to remove the events from the `DefaultSunSpecModel` instead of defining a modified model?

sfeilmeier:
> The Checkstyle test fails due to incorrect naming in `SolarEdgeSunSpecModel`, which uses the same naming as `DefaultSunSpecModel`. The purpose of `SolarEdgeSunSpecModel` is to provide the `DefaultSunSpecModel` without events, as these cause warnings when using the SunSpec model with SolarEdge (either with my implementation or with the included SolarEdge PV inverter implementation). Would it make sense to remove the events from the `DefaultSunSpecModel` instead of defining a modified model?

Thank you! 

1. We already have a bundle for SolarEdge - please move your code there: https://github.com/OpenEMS/openems/tree/develop/io.openems.edge.solaredge

2. Yes, I see this also with SolarEdge PV inverters, that all Events are active. Can you find out if the Events are just inverted (i.e. set to `0` if Event is active) or are they just not supported by SolarEdge? I would definitely not just copy the file, but instead add a parameter that disables or inverts the Events.

sfeilmeier (changes requested):
See above

timo-schlegel:
@sfeilmeier
According to the SolarEdge SunSpec Technical Note:
[https://knowledge-center.solaredge.com/sites/kc/files/sunspec-implementation-technical-note.pdf](https://knowledge-center.solaredge.com/sites/kc/files/sunspec-implementation-technical-note.pdf)

Events are only supported for the meter models (e.g. SunSpec models 201–204).

For the Multiple MPPT Inverter Extension Model (160), events are explicitly marked as *not supported*.
For the standard inverter models (which we use for SolarEdge inverters — models 101, 102, 103), events are not listed at all in the Modbus register mapping.

Based on this documentation, I assume that events are simply *not supported* for the inverter models (101–103), rather than being inverted.

I agree that copying the file is not the best approach. Adding a parameter to disable event handling makes sense. However, I am currently not sure at which level this parameter should be introduced.

timo-schlegel:
@sfeilmeier
I updated the implementation and pushed a new commit.

The code was moved to the existing SolarEdge bundle as suggested. Instead of copying or modifying `DefaultSunSpecModel`, I introduced a small wrapper (`FilteredSunSpecModel`) that allows excluding specific `SunSpecPoint`s from an existing model. This way the generic `DefaultSunSpecModel` can still be reused unchanged, while the SolarEdge implementation can disable the unsupported event registers.

In the same change, I also fixed the existing `SolarEdgePvInverter` implementation to use the filtered model so that the unsupported event registers no longer cause warnings.

Could you please have another look and let me know if this approach fits better with the intended design?

Thank you!

timo-schlegel:
@sfeilmeier quick ping – I implemented the suggested changes above. Could you take another look?

timo-schlegel:
@sfeilmeier quick follow-up on this PR.

Is there still interest in merging this, or should I close it for now?

timo-schlegel:
Closing this for now due to inactivity.

Feel free to reopen or let me know if there is still interest.

sfeilmeier (approved):

PR #3587, opened 2026-02-22
WIP: EssPower: LowPassFilter alternative to PidFilter

sfeilmeier:
Alternative approach to https://github.com/OpenEMS/openems/pull/3581

### DIFF
--- a/io.openems.edge.common/src/io/openems/edge/common/filter/Filter.java
+++ b/io.openems.edge.common/src/io/openems/edge/common/filter/Filter.java
@@ -1,6 +1,6 @@
 package io.openems.edge.common.filter;
 
-public sealed abstract class Filter permits PidFilter, LowPassFilter {
+public abstract sealed class Filter permits PidFilter, LowPassFilter {
 
 	protected Integer lowLimit = null;
 	protected Integer highLimit = null;


Sn0w3y:
Hey @sfeilmeier, nice work on this! Here's my review:

**Overall:** The sealed class hierarchy is a clean design choice and the implementation looks solid. Few points to consider:

1. **Open-Loop vs Closed-Loop**: The LowPassFilter only uses the target value, not the actual measured ESS power. If the hardware cant follow the setpoint (lag, saturation etc.), the filter wont self-correct - unlike PID which corrects via its error term.

2. **Missing `reset()` on LowPassFilter**: If the filter is idle for a while, `lastOutput` gets stale. First value after resuming will blend with the old state. A `reset()` method (ideally on the `Filter` base class) would be nice to have.

3. **Unified method signature**: `applyPidFilter(current, target)` vs `applyLowPassFilter(target)` forces pattern matching. A common `apply(int currentValue, int targetValue)` on the base class would make them truly interchangeable (LowPass just ignores `currentValue`).

4. **API naming**: `setActivePowerEqualsWithPid` is misleading now → should probably be `setActivePowerEqualsWithFilter` or similar.

**I'd also like to make the filter "switchable" at runtime between PID and LowPass.** The architecture is already like 80% there. With the unified method signature and enum config this would be pretty straightforward.

Happy to contribute — if you could grant me merge/write access to the repo, i could also help out with existing PRs directly! :)

FYI: I wrote the Text and let AI rewrite it :)

sfeilmeier:
@Sn0w3y: I also use AI for this a lot... but some questions have been asked before. (BTW: Copilot allows to take existing comments of PRs as Context)

> Overall: The sealed class hierarchy is a clean design choice and the implementation looks solid.

Thank you so much. 😍😉

> 1. Open-Loop vs Closed-Loop: The LowPassFilter only uses the target value, not the actual measured ESS power. If the hardware cant follow the setpoint (lag, saturation etc.), the filter wont self-correct - unlike PID which corrects via its error term.

Yes, this is by intention. Considering this comment/image https://github.com/OpenEMS/openems/pull/2960#issuecomment-2672118310, we explicitely want to split:
- outside control loop (e.g. in Balancing-Controller and other ESS controllers)
- smoothing (via PID- or LowPassFilter)
- internal ramp/PID of inverter (i.e. we should be able to rely on the inverter to actually apply what we write as a set-point; this is by the way also what we do for EVSE control)

> 2. Missing reset() on LowPassFilter: If the filter is idle for a while, lastOutput gets stale. First value after resuming will blend with the old state. A reset() method (ideally on the Filter base class) would be nice to have.

I deleted `reset()` because it was never actually called.

> 3. Unified method signature: applyPidFilter(current, target) vs applyLowPassFilter(target) forces pattern matching. A common apply(int currentValue, int targetValue) on the base class would make them truly interchangeable (LowPass just ignores currentValue).

I did this by purpose to make more obvious, that LowPassFilter does not use `currentValue`.

> 4. API naming: setActivePowerEqualsWithPid is misleading now → should probably be setActivePowerEqualsWithFilter or similar.

True. Once we settle on the actual path, theis naming should be streamlined. The Write-Channel should be renamed as well.

> 5. I'd also like to make the filter "switchable" at runtime between PID and LowPass. The architecture is already like 80% there. With the unified method signature and enum config this would be pretty straightforward.

To remain backwards compatible I decided to not go for a Enum, but add these new settings. If we eventually deprecate the PidFilter, these config params could be removed entirely in future.

> Happy to contribute — if you could grant me merge/write access to the repo, i could also help out with existing PRs directly! :)

I sent you an invitation.

tsicking:
Hi, I haven't had time to do a detailed review, but here are my comments so far:
I agree with @Sn0w3y on 3. You could simply have one abstract method `apply(int currentValue, int targetValue)` in `Filter`, which all subclasses have to implement. This would avoid the `switch` statement in `ManagedSymmetricEss.setActivePowerEqualsWithPid()` (which should be renamed, as @Sn0w3y pointed out), and the whole lines 443-454 could be replaced with 

```java
ess.setActivePowerEquals(filter.apply(ess.getActivePower().orElse(0), value));
```

In that case we don't need `Filter` to be `sealed`. This would allow other filters, e.g. `RampFilter`, to extend the subclass.

I talked to a customer today, who has a system where the PID filter is not working very well in my opinion. He is open for testing the low pass filter if we don't find another solution. So maybe I have a system for testing soon. I'll keep you updated.

Best regards,
Thomas

sfeilmeier:
@tsicking: We have also been working on this internally and improved the code in this PR quite a bit. It's not release ready just yet - but if you want to test it in that system and provide some experiences back, I can update this PR.

sfeilmeier:
We are still finetuning the Filter and trying different approaches; but this PT1-Filter implementation has been merged to develop in latest backport (https://github.com/OpenEMS/openems/pull/3624 -> https://github.com/OpenEMS/openems/commit/da3f41cce2500ce7b00145a15e0ee7b309f948ed)

Issue #3530, opened 2026-01-19
GridOptimizedCharge: Missing bounds check in calculateAvailEnergy causes IndexOutOfBoundsException


cvabc:
## Summary

`Controller.Ess.GridOptimizedCharge` throws `IndexOutOfBoundsException` when prediction data is insufficient. The code lacks input validation before accessing prediction arrays.

## Error

```
[_cycle] WARN [ms.edge.core.cycle.CycleWorker] Error in Controller [ctrlGridOptimizedCharge0]. IndexOutOfBoundsException: toIndex = 5
```

## Root Cause

In `DelayCharge.calculateAvailEnergy()` (line 562), `subList()` is called without checking array bounds:

```java
if (endIndex > 0) {
    // No check: endIndex could exceed array length
    List<Integer> productionList = Arrays.asList(quarterHourlyProduction).subList(1, endIndex);
```

Additionally, line 579 accesses `[0]` without checking if array is empty:
```java
var currentProduction = quarterHourlyProduction[0] == null ? 0 : quarterHourlyProduction[0];
```

**Comparison**: `calculateTargetMinute()` correctly includes bounds checking at line 509:
```java
for (var i = 0; i < min(96 - predictionStartQuarterHourIndex, quarterHourlyProduction.length); i++)
```

## Suggested Fix

Add bounds checking in `calculateAvailEnergy()`:

```java
if (quarterHourlyProduction.length == 0 || quarterHourlyConsumption.length == 0) {
    return 0; // or handle gracefully
}

int safeEndIndex = Math.min(endIndex, Math.min(quarterHourlyProduction.length, quarterHourlyConsumption.length));
if (safeEndIndex > 1) {
    List<Integer> productionList = Arrays.asList(quarterHourlyProduction).subList(1, safeEndIndex);
    // ...
}
```

## Impact

Controller enters FAULT state every cycle when prediction data is insufficient, even though the core charging calculation (`remainingCapacity / remainingTime`) does not require prediction data.

PR #3588, opened 2026-02-23
Fix index bounds and current-quarter handling

Sn0w3y:
Clamp endIndex to the lengths of production/consumption arrays to avoid out-of-bounds slices and change the subList check to require more than one quarter (>1) so only full future quarters are aggregated. Also add guards when reading index 0 values to handle empty arrays and null entries, defaulting to 0 for current production/consumption.

Should fix: https://github.com/OpenEMS/openems/issues/3530


sfeilmeier:
https://github.com/OpenEMS/openems/pull/3645 is fixing the same issue, but provides JUnit tests. I am closing this one and merging the other. Will reference you as a co-author, though.

PR #3589, opened 2026-02-23
[UI] Sort components alphabetically in flat widgets

### LINKED ISSUE #3386: UI: Sort components (inverters, meters, ...) by Alias
### Component

UI

### Description

UX improvement: Right now it seems, that the components are sorted by ID (or order how they were created). For the end user IMHO it would be easier to sort them by the alias (which is displayed in the UI).
cf screenshots below for current solution:

Image

Sn0w3y:
Import ArrayUtils and apply ArrayUtils.sortedAlphabetically to various component lists in production/consumption flat widgets (live and history). Chargers, meters, EVCSs and heat components are now sorted by alias for consistent ordering. No functional logic changes aside from deterministic ordering; changes affect ui/src/app/edge/*/flat/flat.ts files.

Should fix: https://github.com/OpenEMS/openems/issues/3386

### DIFF
--- a/ui/src/app/edge/history/common/production/flat/flat.ts
+++ b/ui/src/app/edge/history/common/production/flat/flat.ts
@@ -17,16 +17,14 @@ export class FlatComponent extends AbstractFlatWidget {
 
     protected override getChannelAddresses(): ChannelAddress[] {
         //  Get Chargers
-        this.chargerComponents = ArrayUtils.sortedAlphabetically(
-            this.config.getComponentsImplementingNature("io.openems.edge.ess.dccharger.api.EssDcCharger")
-                .filter(component => component.isEnabled),
-            (c) => c.alias);
+        this.chargerComponents = this.config.getComponentsImplementingNature("io.openems.edge.ess.dccharger.api.EssDcCharger")
+            .filter(component => component.isEnabled)
+            .sort(ArrayUtils.alphabetically(c => c.alias));
 
         // Get productionMeters
-        this.productionMeterComponents = ArrayUtils.sortedAlphabetically(
-            this.config.getComponentsImplementingNature("io.openems.edge.meter.api.ElectricityMeter")
-                .filter(component => component.isEnabled && this.config.isProducer(component)),
-            (c) => c.alias);
+        this.productionMeterComponents = this.config.getComponentsImplementingNature("io.openems.edge.meter.api.ElectricityMeter")
+            .filter(component => component.isEnabled && this.config.isProducer(component))
+            .sort(ArrayUtils.alphabetically(c => c.alias));
         return [];
     }
 }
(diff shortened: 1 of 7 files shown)


da-Kai:
@Sn0w3y 

Wrapping everything in `ArrayUtils.sortedAlphabetically` feels a bit weird, especially when we could just call `.sort` directly.

Should we consider rewriting `sortedAlphabetically` so that the array can be sorted in-place, like this:
```TypeScript
    /**
     * Sort arrays alphabetically, according to the string returned by fn.
     * Elements for which fn returns null or undefined are sorted to the end in an undefined order.
     *
     * @param array to sort
     * @param fn to get a string to sort by
     * @returns sorted array
     */
    export function sortedAlphabetically<T>(array: T[], fn: (arg: T) => string): T[] {
        return array.sort(alphabetically(fn));
    }

    /**
     * Creates a comparator that sorts items alphabetically by a string key returned from `fn`.
     *
     * The comparator places falsy keys (null, undefined or empty string) after non-falsy keys;
     * two falsy keys compare as equal.
     * 
     * @param fn to get a string to sort by
     * @returns A comparator `(a, b) => number` suitable for `Array.sort`.
     */
    export function alphabetically<T>(fn: (arg: T) => string): (a: T, b: T) => number {
        return (a: T, b: T) => {
            const aVal = fn(a);
            const bVal = fn(b);
            if (!aVal) {
                return !bVal ? 0 : 1;
            } else if (!bVal) {
                return -1;
            }
            return aVal.localeCompare(bVal, undefined, { sensitivity: "accent" });
        }
    }
```
usage:
```TypeScript
.sort(ArrayUtils.alphabetically(a => a.alias));
```

What do you think, @lukasrgr?

Sn0w3y:
@da-Kai you mean like this? :)

da-Kai:
> @da-Kai you mean like this? :)

Yes, and even with a test. 👌

da-Kai (approved):

sfeilmeier:
I am forwarding this to UI/UX team, because this would change visualization for a lot of users. I think sorting by alias is feasible, but sitll need to check.

lukasrgr (approved):

sfeilmeier (changes requested):
Ok, hard to argue if both @lukasrgr and @da-Kai approve... but don't you think we should sort the Components as early as possible to make sure we stay consistent throughout the UI. I.e. directly in EdgeConfig (https://github.com/OpenEMS/openems/blob/develop/ui/src/app/shared/components/edge/edgeconfig.ts#L27)

Sn0w3y:
> but don't you think we should sort the Components as early as possible to make sure we stay consistent throughout the UI. I.e. directly in EdgeConfig

You are 100% right in thsi !
I made the changes and the tests cover it also perfetcly :)

lukasrgr (commented):
Thx for your contribution 🚀 

Could you please add `numeric: true` to the `localeCompare`

PR #3601, opened 2026-03-02
[EDGE] Handle UNDEFINED_VALUE as null in ModbusRecordChannel

Sn0w3y:
This commit updates the ModbusRecordChannel to interpret UNDEFINED_VALUE as null for numeric types (FLOAT64, FLOAT32, UINT16, UINT32, UINT64). This change ensures that when a Modbus record contains a value that is designated as undefined or not applicable, it is correctly handled as a null value in the system, rather than being processed as a valid numeric value. This is crucial for avoiding incorrect data interpretation and ensuring data integrity in applications that rely on Modbus communication.

### DIFF
--- a/io.openems.edge.common/src/io/openems/edge/common/modbusslave/ModbusRecordChannel.java
+++ b/io.openems.edge.common/src/io/openems/edge/common/modbusslave/ModbusRecordChannel.java
@@ -179,40 +179,15 @@ public class ModbusRecordChannel extends ModbusRecord {
 			}
 		}
 
-		// yes, it is full -> Prepare ByteBuffer
-		var buff = ByteBuffer.allocate(this.writeValueBuffer.length);
-		for (Byte element : this.writeValueBuffer) {
-			buff.put(element);
+		// Copy buffer to byte array and clear buffer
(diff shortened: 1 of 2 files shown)


sfeilmeier (changes requested):
Thanks - did you try this? We'll definitely need proper JUnit tests here - maybe by moving the switch-case to a testable static method. 

Probably it would be also better to compare the byte-array with UNDEFINED before converting the value to an Object. That way we could avoid the type-casting.

Sn0w3y:
> Thanks - did you try this? We'll definitely need proper JUnit tests here - maybe by moving the switch-case to a testable static method.

I in fact did not try this - maybe we could make a call out in community or use the FEMS Customer in community as tester if possible on your side? :)
The UNIT Tests should cover the new Code now pretty well

sfeilmeier:
> I in fact did not try this - maybe we could make a call out in community or use the FEMS Customer in community as tester if possible on your side? :) The UNIT Tests should cover the new Code now pretty well

No sorry, that would be a lot of effort. Most of this code should be quite easily testable with Simulated ESS directly in the IDE.

PR #3585, opened 2026-02-21
ESS-Power: alternative implementation without Linear Equation System

sfeilmeier:
- This implementation does not use a Linear Equation System to validate set-points and distribute them to individual inverters
- Activated by new solver strategy `BALANCE`
- Move logic to `PowerDistributionHandler`s

See discussion in https://community.openems.io/t/neuer-optimizer-logik-fur-mehrere-hybrid-esses/10138/34

PR #3602, opened 2026-03-03
Fix EssPower thread-safety race condition (#3500)

### LINKED ISSUE #3500: [Bug] EssPower circular dependency fix from #3113 needs to be applied to all ESS implementations
# [Bug] EssPower circular dependency causes "Coefficient was not found" error in multiple ESS implementations

## Summary

The fix for the circular dependency issue in PR #3113 was only applied to `Edge2EdgeWebsocketEssImpl`. However, **many other ESS implementations** have the same problematic `@Reference` pattern for the `Power` service, causing the same "Coefficient was not found" error.

## Error Message

```
WARN [ems.edge.ess.core.power.Solver] Power-Solver: Solve failed: Coefficient for [ess0,ALL,ACTIVE] was not found. Ess-Power is not (yet) fully initialized.
```

### 1. Circular Dependency (Fixed in #3113 for Edge2Edge only)

The dependency chain `ess -> _power -> ess` causes issues during OSGi component activation:

1. ESS component tries to activate
2. ESS has `@Reference(cardinality = MANDATORY)` to `Power`
3. `Power` (EssPowerImpl) needs ESS to register via `addEss()` to create coefficients
4. Race condition: ESS may be referenced before its coefficients are initialized

**The fix in #3113:**
```java
// Before (problematic):
@Reference
private Power power;

// After (fixed):
@Reference(cardinality = ReferenceCardinality.OPTIONAL,
           policyOption = ReferencePolicyOption.GREEDY,
           policy = ReferencePolicy.DYNAMIC)
private volatile Power power;
```

### 2. Thread Safety Issue in Data.java (Unfixed)

There's also an underlying race condition in `Data.java`:

```java
// EssPowerImpl.java - addEss() is synchronized
protected synchronized void addEss(ManagedSymmetricEss ess) {
    this.data.addEss(ess);  // adds to esss list, then updates coefficients
}

// Data.java - getConstraintsForAllInverters() is NOT synchronized
public List<Constraint> getConstraintsForAllInverters() throws OpenemsException {
    return this.getConstraintsWithoutDisabledInverters(Collections.emptyList());
}
```

**Race condition:**
```
Thread A (OSGi):                    Thread B (Cycle Scheduler):
─────────────────                   ───────────────────────────
addEss(ess0) - acquires lock
  esss.add(ess0) ✓
  updateInverters() starts...
                                    Solver.solve() - NO LOCK!
                                      getConstraintsForAllInverters()
                                        iterates esss → sees ess0
                                        coefficients.of("ess0"...)
                                        → FAILS! (not initialized yet)
  coefficients.initialize() ✓
releases lock
```

### Affected ESS Implementations

| Module | File |
|--------|------|
| io.openems.edge.ess.generic | EssGenericManagedSymmetricImpl.java |
| io.openems.edge.ess.generic | EssGenericOffGridImpl.java |
| io.openems.edge.ess.cluster | EssClusterImpl.java |
| io.openems.edge.goodwe | GoodWeEssImpl.java |
| io.openems.edge.goodwe | GoodWeBatteryInverterImpl.java |
| io.openems.edge.edge2edge | Edge2EdgeEssImpl.java |
| io.openems.edge.edge2edge | Edge2EdgeGenericReadComponentImpl.java |
| io.openems.edge.fenecon.mini | FeneconMiniEssImpl.java |
| io.openems.edge.fenecon.pro | FeneconProEssImpl.java |
| io.openems.edge.ess.fenecon.commercial40 | EssFeneconCommercial40Impl.java |
| io.openems.edge.sma | EssSmaSunnyIslandImpl.java |
| io.openems.edge.kaco.blueplanet.hybrid10 | KacoBlueplanetHybrid10EssImpl.java |
| io.openems.edge.kostal | KostalManagedEssImpl.java |
| io.openems.edge.simulator | SimulatorEssSinglePhaseReactingImpl.java |
| io.openems.edge.simulator | SimulatorEssSymmetricReactingImpl.java |
| io.openems.edge.simulator | SimulatorEssAsymmetricReactingImpl.java |

*Note: Other ESS implementations in third-party modules may also be affected.*

## Steps to Reproduce

1. Configure multiple ESS components (e.g., ess0, ess1, ess2, ess3)
2. Start OpenEMS Edge
3. Observe logs during startup
4. Error appears: `Coefficient for [essX,ALL,ACTIVE] was not found`

The error may:
- Appear briefly during startup and then stop
- Persist for a long time if component activation order is unfavorable
- Reappear when ESS components reconnect (e.g., Modbus disconnect/reconnect)

### Fix 1: Apply OPTIONAL/DYNAMIC pattern to all ESS implementations

For each affected ESS implementation:

```java
// Change from:
@Reference
private Power power;
// or
@Reference(cardinality = ReferenceCardinality.MANDATORY, policyOption = ReferencePolicyOption.GREEDY)
private Power power;

// To:
@Reference(cardinality = ReferenceCardinality.OPTIONAL,
           policyOption = ReferencePolicyOption.GREEDY,
           policy = ReferencePolicy.DYNAMIC)
private volatile Power power;
```

And update `getPower()`:

```java
@Override
public Power getPower() {
    final var power = this.power;
    if (power == null) {
        throw new OpenemsRuntimeException("Ess.Power is not yet available");
    }
    return power;
}
```

### Fix 2: Add synchronization to Data.java (additional fix)

```java
// Data.java - make constraint retrieval synchronized
public synchronized List<Constraint> getConstraintsWithoutDisabledInverters(
        Collection<Inverter> disabledInverters) throws OpenemsException {
    // existing code
}
```

## Environment

- OpenEMS Version: 2025.6.0+ (post PR #3113)
- Java Version: 21
- OS: Linux (ARM64)

## Related

- PR #3113 - FEMS Backports 2025-05-01 (partial fix for Edge2EdgeWebsocketEss only)
- Community discussion: https://community.openems.io/t/how-can-i-prevent-the-error-coefficient-ess0-all-active-was-not-found-ess-power-is-not-yet-fully-initialized-from-occurring/10352

## Checklist

- [ ] Apply fix to all ESS implementations listed above
- [ ] Consider adding synchronization to `Data.getConstraintsWithoutDisabledInverters()`
- [ ] Add unit tests for race condition scenarios

rishabhvaish:
## Summary
- Race condition between OSGi thread (adding ESS) and Solver thread (reading coefficients) causes "Coefficient was not found" errors
- ESS operates at 0W until manual restart — 40+ hour production downtime reported
- Root cause: `Data.getConstraints*` and `Coefficients.of()` are not synchronized, while `Coefficients.initialize()` clears-then-rebuilds (non-atomic)

## Root cause
1. OSGi thread: `Data.addEss()` → `updateInverters()` → `Coefficients.initialize()` → clears coefficient list
2. Solver thread (concurrent): `getConstraintsWithoutDisabledInverters()` → sees new ESS in CopyOnWriteArrayList → calls `Coefficients.of(essId)` → coefficient not found (list was cleared)

### `Coefficients.java` (`io.openems.edge.ess.api`)
- **`of()` → `synchronized`**: Prevents reading coefficients while `initialize()` is rebuilding them
- **`initialize()` → build-then-swap**: Coefficients are built in a temporary `ArrayList` first, then `clear()` + `addAll()` happen at the end while holding the monitor lock. No reader (via `of()`) can observe the empty intermediate state.

### `Data.java` (`io.openems.edge.ess.core`)
- **`getConstraintsForAllInverters()` → `synchronized`**
- **`getConstraintsForInverters()` → `synchronized`**
- **`getConstraintsWithoutDisabledInverters()` → `synchronized`**

These methods read `esss`, `inverters`, `coefficients`, and `symmetricMode` — all of which are mutated by `addEss()`/`removeEss()`/`updateInverters()` (which are already `synchronized` on the same `Data` instance). Without synchronization, the Solver thread can observe partially-updated state.

## Test plan
- [ ] Start OpenEMS with multiple ESS components — verify no "Coefficient not found" errors in log
- [ ] Add/remove ESS component during active operation — verify no race condition errors
- [ ] Verify ESS power limits are correctly applied after component restart
- [ ] Stress test: rapidly add/remove ESS components while solver is running

Fixes #3500

### DIFF
--- a/io.openems.edge.ess.api/src/io/openems/edge/ess/power/api/Coefficients.java
+++ b/io.openems.edge.ess.api/src/io/openems/edge/ess/power/api/Coefficients.java
@@ -26,32 +26,25 @@ public class Coefficients {
 	 * @param essIds        Set of ESS-Ids
 	 */
 	public synchronized void initialize(boolean symmetricMode, Set<String> essIds) {
-		// Build the new coefficients in a temporary list first, then swap atomically.
-		// This avoids a window where the list is cleared but not yet rebuilt, which
-		// would cause concurrent readers in of() to throw "Coefficient was not found".
-		var newCoefficients = new java.util.ArrayList<Coefficient>();
+		this.coefficients.clear();
+		this.symmetricMode = symmetricMode;
 		var index = 0;
 		for (String essId : essIds) {
 			if (symmetricMode) {
 				// Symmetric Mode
 				for (Pwr pwr : Pwr.values()) {
-					newCoefficients.add(new Coefficient(index++, essId, SingleOrAllPhase.ALL, pwr));
+					this.coefficients.add(new Coefficient(index++, essId, SingleOrAllPhase.ALL, pwr));
 				}
 			} else {
 				// Asymmetric Mode
 				for (var phase : SingleOrAllPhase.values()) {
 					for (Pwr pwr : Pwr.values()) {
-						newCoefficients.add(new Coefficient(index++, essId, phase, pwr));
+						this.coefficients.add(new Coefficient(index++, essId, phase, pwr));
 					}
 				}
 			}
 		}
-		this.symmetricMode = symmetricMode;
 		this.noOfCoefficients = index;
-		// Atomic swap: clear and addAll in immediate succession on CopyOnWriteArrayList.
-		// Because of() is also synchronized on 'this', no reader can see the empty state.
-		this.coefficients.clear();
-		this.coefficients.addAll(newCoefficients);
 	}
 
 	/**
--- a/io.openems.edge.ess.api/src/io/openems/edge/ess/power/api/Coefficients.java
+++ b/io.openems.edge.ess.api/src/io/openems/edge/ess/power/api/Coefficients.java
@@ -26,25 +26,32 @@ public class Coefficients {
 	 * @param essIds        Set of ESS-Ids
 	 */
 	public synchronized void initialize(boolean symmetricMode, Set<String> essIds) {
-		this.coefficients.clear();
-		this.symmetricMode = symmetricMode;
+		// Build the new coefficients in a temporary list first, then swap atomically.
+		// This avoids a window where the list is cleared but not yet rebuilt, which
+		// would cause concurrent readers in of() to throw "Coefficient was not found".
+		var newCoefficients = new java.util.ArrayList<Coefficient>();
 		var index = 0;
 		for (String essId : essIds) {
 			if (symmetricMode) {
 				// Symmetric Mode
 				for (Pwr pwr : Pwr.values()) {
-					this.coefficients.add(new Coefficient(index++, essId, SingleOrAllPhase.ALL, pwr));
+					newCoefficients.add(new Coefficient(index++, essId, SingleOrAllPhase.ALL, pwr));
 				}
 			} else {
 				// Asymmetric Mode
 				for (var phase : SingleOrAllPhase.values()) {
 					for (Pwr pwr : Pwr.values()) {
-						this.coefficients.add(new Coefficient(index++, essId, phase, pwr));
+						newCoefficients.add(new Coefficient(index++, essId, phase, pwr));
 					}
 				}
 			}
 		}
+		this.symmetricMode = symmetricMode;
 		this.noOfCoefficients = index;
+		// Atomic swap: clear and addAll in immediate succession on CopyOnWriteArrayList.
+		// Because of() is also synchronized on 'this', no reader can see the empty state.
+		this.coefficients.clear();
+		this.coefficients.addAll(newCoefficients);
 	}
 
 	/**
@@ -57,7 +64,7 @@ public class Coefficients {
 	 * @return the {@link Coefficient}
 	 * @throws OpenemsException on error
 	 */
-	public Coefficient of(String essId, SingleOrAllPhase phase, Pwr pwr) throws OpenemsException {
+	public synchronized Coefficient of(String essId, SingleOrAllPhase phase, Pwr pwr) throws OpenemsException {
 		if (this.symmetricMode && phase != SingleOrAllPhase.ALL) {
 			throw new OpenemsException("Symmetric-Mode is activated. Coefficients for [" + essId + "," + phase + ","
 					+ pwr + "] is not available!");
--- a/io.openems.edge.ess.core/src/io/openems/edge/ess/core/power/v1/Data.java
+++ b/io.openems.edge.ess.core/src/io/openems/edge/ess/core/power/v1/Data.java
@@ -166,7 +166,7 @@ public class Data {
 	 * @return List of Constraints
 	 * @throws OpenemsException on error
 	 */
-	public List<Constraint> getConstraintsForAllInverters() throws OpenemsException {
+	public synchronized List<Constraint> getConstraintsForAllInverters() throws OpenemsException {
 		return this.getConstraintsWithoutDisabledInverters(Collections.emptyList());
 	}
 
@@ -177,7 +177,7 @@ public class Data {
 	 * @return List of {@link Constraint}s
 	 * @throws OpenemsException on error
 	 */
-	public List<Constraint> getConstraintsForInverters(Collection<Inverter> enabledInverters) throws OpenemsException {
+	public synchronized List<Constraint> getConstraintsForInverters(Collection<Inverter> enabledInverters) throws OpenemsException {
 		List<Inverter> disabledInverters = new ArrayList<>(this.inverters);
 		disabledInverters.removeAll(enabledInverters);
 		return this.getConstraintsWithoutDisabledInverters(disabledInverters);
@@ -190,7 +190,7 @@ public class Data {
 	 * @return List of Constraints
 	 * @throws OpenemsException on error
 	 */
-	public List<Constraint> getConstraintsWithoutDisabledInverters(Collection<Inverter> disabledInverters)
+	public synchronized List<Constraint> getConstraintsWithoutDisabledInverters(Collection<Inverter> disabledInverters)
 			throws OpenemsException {
 		final var esss = this.esssSupplier.get();
 


Sn0w3y:
**The synchronization is unnecessary:**
  - `coefficients`, `esss`, `inverters`, and `constraints` are all `CopyOnWriteArrayList`, which is inherently thread-safe for concurrent read/write. Iteration always works on a snapshot
  of the internal array.
  - `initialize()` is already `synchronized`.
  - The brief window between `clear()` and the `add()` calls can produce at most a single WARN log entry during startup, which self-corrects on the next cycle.

  **The "atomic swap" is not actually atomic:**
  ```java
  this.coefficients.clear();           // list is empty here
  this.coefficients.addAll(newCoefficients);  // list is filled here
  The same window exists between clear() and addAll(). It only "works" because of() is now also synchronized - making the temporary list entirely redundant.
 ```
  The real issue from #3500 is something else entirely.
  The evidence posted there shows:
  org.osgi.framework.ServiceException: ServiceFactory.getService() resulted in a cycle.
  This is an OSGi SCR service cycle error - meaning Felix detected a circular dependency during service resolution and addEss() was never called for that ESS. That's why the "Coefficient
  not found" error is permanent and never self-corrects. No amount of synchronized keywords will fix a method that was never invoked.

  Adding synchronized to the Data getter methods introduces unnecessary lock contention on a hot path - getConstraintsWithoutDisabledInverters() runs every cycle (~1x/second) and performs
  multiple ConstraintUtil calls. Holding the Data lock during that time blocks addEss()/removeEss() for no benefit.

rishabhvaish:
@Sn0w3y — Thanks for the thorough review. I've re-read the code carefully and want to address each point.

### You are right: the specific #3500 production incident is an OSGi cycle problem

The logs posted by @cvabc show `ServiceFactory.getService() resulted in a cycle`, meaning Felix never completed activation of ess1, so `addEss()` was never called. My PR does not fix that. That needs an OSGi-level fix (e.g., the OPTIONAL/DYNAMIC pattern from #3113, or restructuring the service dependency chain). I should have been clearer in the PR description that this addresses the *race condition pathway* described in #3500, not the *OSGi cycle pathway* that caused the specific 40-hour incident.

### However, the cross-object race condition IS real — CopyOnWriteArrayList does not protect against it

You're correct that each individual `CopyOnWriteArrayList` is thread-safe for concurrent read/write. But the race is not about a single list — it's about **consistency across multiple data structures**. Here's the exact sequence:

```
Thread A (OSGi):                         Thread B (Solver, every ~1s):
────────────────                         ──────────────────────────────
Data.addEss(ess0) [holds Data lock]
  esss.add(ess0)  ← VISIBLE IMMEDIATELY
  updateInverters()
    inverters.clear()
    ... rebuilding inverters ...
                                         getConstraintsWithoutDisabledInverters() [NO Data lock in original code]
                                           ConstraintUtil.createGenericEssConstraints(coefficients, esss, ...)
                                             iterates esss snapshot → sees ess0
                                             coefficients.of("ess0", ALL, ACTIVE)
                                             → FAILS: coefficient not found (initialize() hasn't run yet)
    coefficients.initialize()  ← too late
```

`CopyOnWriteArrayList` guarantees safe iteration over one list. It cannot guarantee that `esss` and `coefficients` are consistent with each other at the point of observation. This is a classic compound-operation problem — you need a lock that spans the check-then-act across both data structures.

The `synchronized` on `Data.getConstraintsWithoutDisabledInverters()` is what fixes this: since `addEss()` already holds the `Data` lock, making the getter also lock on `Data` ensures a reader never observes an intermediate state where `esss` has a new entry but `coefficients` hasn't been rebuilt yet.

### You are right about the "atomic swap" redundancy

Your observation is sharp: the temporary `ArrayList` + `clear()`/`addAll()` inside `initialize()` only avoids the empty-list window because `of()` is also `synchronized` on the same `Coefficients` instance. If `of()` is `synchronized`, the temporary list is unnecessary — the original `clear()` + rebuild-in-place is equally safe under the monitor lock. I'll simplify `initialize()` to remove the redundant temp list.

### On lock contention

The `Data`-level lock contention concern is fair to raise but not material in practice:
- `addEss()`/`removeEss()` are OSGi lifecycle operations — they happen at startup or on component reconnect, not on the hot path
- The solver runs ~1x/second; holding the `Data` lock for the duration of `getConstraintsWithoutDisabledInverters()` blocks `addEss()` for at most one solver cycle (~milliseconds)
- This is vastly preferable to 40+ hours of silent 0W operation

### Summary of what I'll change

1. **Keep** `synchronized` on `Data.getConstraintsForAllInverters()`, `getConstraintsForInverters()`, `getConstraintsWithoutDisabledInverters()` — these are the essential fix for cross-object consistency
2. **Keep** `synchronized` on `Coefficients.of()` — needed for callers that go through `Coefficients` directly outside the `Data` lock (e.g., `EssPowerImpl.getCoefficient()`, `Solver.solve()` calling `InverterPrecision.apply()` and `AddConstraintsForNotStrictlyDefinedCoefficients`)
3. **Remove** the temporary `ArrayList` build-then-swap in `Coefficients.initialize()` — as you pointed out, it's redundant when `of()` is synchronized. I'll restore the original in-place rebuild
4. **Update** the PR description to clearly separate the two root causes: OSGi cycle (not fixed here) vs. thread-safety race (fixed here)

Sn0w3y:
> @Sn0w3y — Thanks for the thorough review. I've re-read the code carefully and want to address each point.
> 
> ### You are right: the specific #3500 production incident is an OSGi cycle problem
> The logs posted by @cvabc show `ServiceFactory.getService() resulted in a cycle`, meaning Felix never completed activation of ess1, so `addEss()` was never called. My PR does not fix that. That needs an OSGi-level fix (e.g., the OPTIONAL/DYNAMIC pattern from #3113, or restructuring the service dependency chain). I should have been clearer in the PR description that this addresses the _race condition pathway_ described in #3500, not the _OSGi cycle pathway_ that caused the specific 40-hour incident.
> 
> ### However, the cross-object race condition IS real — CopyOnWriteArrayList does not protect against it
> You're correct that each individual `CopyOnWriteArrayList` is thread-safe for concurrent read/write. But the race is not about a single list — it's about **consistency across multiple data structures**. Here's the exact sequence:
> 
> ```
> Thread A (OSGi):                         Thread B (Solver, every ~1s):
> ────────────────                         ──────────────────────────────
> Data.addEss(ess0) [holds Data lock]
>   esss.add(ess0)  ← VISIBLE IMMEDIATELY
>   updateInverters()
>     inverters.clear()
>     ... rebuilding inverters ...
>                                          getConstraintsWithoutDisabledInverters() [NO Data lock in original code]
>                                            ConstraintUtil.createGenericEssConstraints(coefficients, esss, ...)
>                                              iterates esss snapshot → sees ess0
>                                              coefficients.of("ess0", ALL, ACTIVE)
>                                              → FAILS: coefficient not found (initialize() hasn't run yet)
>     coefficients.initialize()  ← too late
> ```
> 
> `CopyOnWriteArrayList` guarantees safe iteration over one list. It cannot guarantee that `esss` and `coefficients` are consistent with each other at the point of observation. This is a classic compound-operation problem — you need a lock that spans the check-then-act across both data structures.
> 
> The `synchronized` on `Data.getConstraintsWithoutDisabledInverters()` is what fixes this: since `addEss()` already holds the `Data` lock, making the getter also lock on `Data` ensures a reader never observes an intermediate state where `esss` has a new entry but `coefficients` hasn't been rebuilt yet.
> 
> ### You are right about the "atomic swap" redundancy
> Your observation is sharp: the temporary `ArrayList` + `clear()`/`addAll()` inside `initialize()` only avoids the empty-list window because `of()` is also `synchronized` on the same `Coefficients` instance. If `of()` is `synchronized`, the temporary list is unnecessary — the original `clear()` + rebuild-in-place is equally safe under the monitor lock. I'll simplify `initialize()` to remove the redundant temp list.
> 
> ### On lock contention
> The `Data`-level lock contention concern is fair to raise but not material in practice:
> 
> * `addEss()`/`removeEss()` are OSGi lifecycle operations — they happen at startup or on component reconnect, not on the hot path
> * The solver runs ~1x/second; holding the `Data` lock for the duration of `getConstraintsWithoutDisabledInverters()` blocks `addEss()` for at most one solver cycle (~milliseconds)
> * This is vastly preferable to 40+ hours of silent 0W operation
> 
> ### Summary of what I'll change
> 1. **Keep** `synchronized` on `Data.getConstraintsForAllInverters()`, `getConstraintsForInverters()`, `getConstraintsWithoutDisabledInverters()` — these are the essential fix for cross-object consistency
> 2. **Keep** `synchronized` on `Coefficients.of()` — needed for callers that go through `Coefficients` directly outside the `Data` lock (e.g., `EssPowerImpl.getCoefficient()`, `Solver.solve()` calling `InverterPrecision.apply()` and `AddConstraintsForNotStrictlyDefinedCoefficients`)
> 3. **Remove** the temporary `ArrayList` build-then-swap in `Coefficients.initialize()` — as you pointed out, it's redundant when `of()` is synchronized. I'll restore the original in-place rebuild
> 4. **Update** the PR description to clearly separate the two root causes: OSGi cycle (not fixed here) vs. thread-safety race (fixed here)

Did you even look at the Code yourself or is this all AI I am speaking with?

rishabhvaish:
@Sn0w3y Yeah, I used AI to help draft the response, but I did read the code. Apologies, if it's against policies.

The technical point stands — the race isn't about individual list safety, it's about esss and coefficients being out of sync between add() and initialize(). CopyOnWriteArrayList doesn't help there. 

Already removed the temp list since you were right about that being redundant. Happy to walk through the race window if you want to dig in further.

sfeilmeier:
@rishabhvaish Please validate if this still exists with latest develop. We have done some refactoring of the code, as announced in https://github.com/OpenEMS/openems/pull/3585 as part of latest backports: https://github.com/OpenEMS/openems/pull/3624 -> https://github.com/OpenEMS/openems/commit/7cb0d28424fadc3bd5868dedc8ddd20aba6ddb6c

rishabhvaish:
@sfeilmeier — Validated against latest `develop` (including the v1 refactoring from #3624 / `7cb0d28`).

The refactoring moved `Data.java` to `power/v1/Data.java` and replaced the internal `esss` list with a `Supplier<List<ManagedSymmetricEss>>` from `EssPowerImpl`. However, the cross-object race condition **still exists**:

- `Coefficients.of()` is still not `synchronized` — a reader can observe the coefficient list mid-rebuild (after `clear()` but before all coefficients are re-added in `initialize()`)
- `Data.getConstraintsWithoutDisabledInverters()` is still not `synchronized` — the solver thread can call `esssSupplier.get()`, see a new ESS, and then call `coefficients.of(essId)` before `updateInverters()` has rebuilt the coefficients for that ESS
- `updateInverters()` holds the `Data` monitor lock, but the getter methods do not — so they run concurrently

The supplier pattern actually makes the window slightly wider: `esssSupplier.get()` returns the live `CopyOnWriteArrayList` from `EssPowerImpl`, which can include a newly added ESS even before `updateInverters()` has been called.

I've rebased onto latest `develop` — the changes now target the new `power/v1/Data.java` path. The diff is unchanged: `synchronized` on `Coefficients.of()` and the three `Data.getConstraints*()` methods.

PR #3629, opened 2026-03-13
script to run the already included j2mod cli tool modpoll easily without java hussle

janklostermann:
This is a script to make the CLI modpoll tool, which is already included into the OpenEMS codebase (within the j2mod library), easily available. 
It serves to debug modbus connections.

co-authored by: claude@anthropic.com


janklostermann:
or would `io.openems.edge.bridge.modbus/tools/modpoll_j2mod.sh` be a better location for it?

da-Kai:
Hi @janklostermann,

thanks for putting this together, it is an interesting approach. That said, I have a few concerns regarding its long-term viability:

1. **Maintenance**: This is a Frankensteined script that attempts to extract a tool from a library within a prebuilt JAR. This is highly dependent on the internal structure of the EdgeApp.jar and the local Gradle and Maven cache. Using find against ~/.gradle or ~/.m2 is non-deterministic and increases the risk of "works on my machine" issues.

2. **Redundancy**: There are well-established, standalone Modbus debugging tools, like the original modpoll binary, or qModMaster, that are purpose-built for this task. These tools don't require a Java runtime or an existing OpenEMS build. For quick debugging, these are typically more efficient and reliable.

3. **Scope**: We aim to keep this repository focused on the OpenEMS framework itself. Adding wrapper scripts around internal utilities can lead to unnecessary tooling bloat.

TL;DR: I see a maintenance burden without a clear benefit. If there’s a concrete use case where external tools are insufficient, please share it. That would help justify this addition.

Thanks!

janklostermann:
Hi @da-Kai ,
I can fully see and and whole-heartedly agree to your concerns.

The problems I have with the external tools:
* there are many "modpoll" around, everyone finds a different one if searching for it - causes confusion - no single point of reference.
* all cover different parts of the modbus universe: some do not support Modbus/ASCII, for some you need complete register tables, ...

Using a modpoll based on exactly the same library that openems uses (even regarding version) seems perfect for debugging modbus in an openems environment - if it works/doesn't work with this modpoll, you are likely to experience the same issues in openems.
This makes debugging much more straightforward. It helped me.

Becoming aware of the existence and successfully using this modpoll by hand, without such script is highly unlikely - eventhough it is already integrated in the codebase. Saving others my efforts and providing a common reference point for modbus debugging around openems, was my intention. 

If it breaks, with today's help of AI it is still easier to fix than to discover...

I hope, I could explain myself properly.
I am open to further discussion on how to go on with it. If you have questions please ask.

PR #3632, opened 2026-03-13
io.openems.edge.bridge.modbus.ascii - adding support for Modbus/ASCII over serial

janklostermann:
This PR adds a Modbus/ASCII bridge to openems to support devices that communicate through this protocol via a serial port (one example would be the ABL eMH1). 

It is implemented as a separate component to not interfere with the existing modbus bridge, which implements the Modbus/RTU protocol over serial under the name BridgeModbusSerialImpl.java.

It might be a good idea to integrate this code into the `io.openems.edge.bridge.modbus` component. There could be even an abstraction be made to separate-out common code for the serial RTU and ASCII implementations (both use `j2mod` for the heavy-lifting) and rename the current `Serial` to `SerialRtu`.
If you wish I can adjust this PR accordingly. 
(I chose the path of separating them to simplify evaluation and merging, and avoid the risk of mixing things up in the field-proven modbus bridge.)

Unfortunatetely I was not able to bring this component into the section `device interfaces` alongside the other bridges. I would be happy to fix this if I knew how.


janklostermann:
Sorry, [cbd5948](https://github.com/OpenEMS/openems/pull/3632/commits/cbd5948563edd4f0af29f66b390cd33972dc9d6c) was definitely premature. I am going to fix the issues and integrate some further improvements of the PR. I will come back soon.

PR #3672, opened 2026-04-09
Taking EVSE chargepoints off of Meters category by refactoring getFactoriesByNature to support multiple exclusions

janklostermann:
To avoid EVSE chargepoints ending up in the Meters category, this fix allows more then one exclusion per category, now allowing a list. It is used to enhance the CATEGORY.TITLE.METER with not only excluding evcs.api.Evcs, but also evse.api.chargepoint.EvseChargePoint.

### DIFF
--- a/ui/src/app/shared/components/edge/edgeconfig.ts
+++ b/ui/src/app/shared/components/edge/edgeconfig.ts
@@ -127,7 +127,10 @@ export class EdgeConfig {
                 category: { title: translate.instant("SETTINGS.CATEGORY.TITLE.METER"), icon: "speedometer-outline" },
                 factories: [
                     EdgeConfig.getFactoriesByNature(factories, "io.openems.edge.meter.api.SymmetricMeter"), // TODO replaced by ElectricityMeter
-                    EdgeConfig.getFactoriesByNature(factories, "io.openems.edge.meter.api.ElectricityMeter", "io.openems.edge.evcs.api.Evcs"),
+                    EdgeConfig.getFactoriesByNature(factories, "io.openems.edge.meter.api.ElectricityMeter", [
+                        "io.openems.edge.evcs.api.Evcs",
+                        "io.openems.edge.evse.api.chargepoint.EvseChargePoint",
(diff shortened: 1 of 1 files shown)


janklostermann:
Can someone please help me with this linter issue in the CI? 

If I run the sequence locally: "All files pass linting."

Beyond that: It complains that `const` is not allowed to be used to declare variables. A practice that was used in this function before my additions. I just sticked with how it was done...

Therefore: How can I make this PR acceptable and let it pass the linter?

sfeilmeier:
@lukasrgr Could you please have a look?

ui/src/app/shared/components/edge/edgeconfig.ts
```diff
@@ -281,13 +284,14 @@ export class EdgeConfig {
      *
      * @param factories the given EdgeConfig.Factory
      * @param includeNature the name of the Nature to be included
-     * @param excludeNature an optional name of a Nature to be excluded
+     * @param  excludeNature an optional name of a Nature (or list of Natures) to be excluded
      */
-    public static getFactoriesByNature(factories: { [id: string]: EdgeConfig.Factory }, includeNature: string, excludeNature?: string): EdgeConfig.Factory[] {
+    public static getFactoriesByNature(factories: { [id: string]: EdgeConfig.Factory }, includeNature: string, excludeNature?: string | string[]): EdgeConfig.Factory[] {
```
lukasrgr, line 289: includeNature and excludeNature could be two different types then, i would probably only expect a string if includeNature is a string, 

if you just use string[] as param type, you could also save yourself from checking if it is an array or not

lukasrgr (changes requested):

ui/src/app/shared/components/edge/edgeconfig.ts
janklostermann: I did not want to break anything. Therefore I kept string and just added another option.
But if you prefer just string[], this is fine for me..
Shall I do it, or do you want to take over?

Issue #3484, opened 2025-12-18
Add IPv6 support


MAngel666:
### Component

UI

### Description

It would be nice to have ipv6 settings/support in the WebUI.

Sn0w3y:
While IPv6 support is generally a good idea for future-proofing, in the context of OpenEMS and local energy management networks,
  the practical benefit is quite limited:

  1. Energy devices use IPv4: Almost all inverters, batteries, wallboxes, and meters communicate via IPv4 (Modbus TCP, SunSpec, REST
  APIs). These protocols and device implementations are IPv4-based.
  2. Local network scope: OpenEMS Edge typically runs in private networks (192.168.x.x, 10.x.x.x) where IPv4 address space is more
  than sufficient for typical installations with 10-50 devices.
  3. Added configuration complexity: IPv6 would add complexity for end users without providing tangible benefits in most deployments.
  4. No address exhaustion locally: Unlike public internet scenarios, local energy management networks don't face IPv4 address
  scarcity.

  Potential edge cases where IPv6 might be relevant:
  - Large industrial sites with hundreds of devices
  - ISPs providing IPv6-only connectivity (with CGNAT for IPv4)

  For now, this seems like a low-priority enhancement. The development effort would likely be better spent on features that address
  more common use cases.

  That said, if there's a specific deployment scenario where IPv6 is required, please share more details - that context would help
  reassess the priority.

MAngel666:
IPv6 is becoming increasingly widespread and will eventually completely replace IPv4. Some people (like me) are already trying to switch to IPv6 in their LANs to avoid dual-stack configurations.
It's not a high priority, that's clear. But offering communication via IPv6 is simply part of the package these days.

PR #3689, opened 2026-04-21
UI: Add IPv6 support

da-Kai:
Add IPv6 capabilities for UI
Move Network related regex pattern and validation functions into a separate `InetUtils`-namespace.
Network settings: Keep IPv4 validation, until edge supports IPv6

Address #3484 partially


sfeilmeier:
@lukasrgr, @DerMuaaa, @venu-sagar Could anyone of you have a look please?

ui/src/app/shared/utils/inet/inet.utils.ts
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    /**
+     * Checks whether a string is a valid network address in CIDR notation.
+     *
+     * @param value the input string in the format "address/prefix"
+     * @returns the detected network address type
+     */
+    export function isNetworkAddress(value: string): IpType {
+        if (value === null || value.length == 0) { return IpType.None; }
```
venu-sagar: possible null-check inconsistency.
Every other function in the file guards with `value != null` (catches both `null` and `undefined`). This one uses strict `=== null`, so if `undefined` is passed (e.g. an untouched formly field), `value.length` throws a `TypeError`. Should be `value == null` for consistency and safety. Not caught by the spec since it only tests `null`, not `undefined`

ui/src/app/shared/utils/inet/inet.utils.ts
```diff
@@ -0,0 +1,172 @@
...
+
+        const parts: string[] = value.split("/");
+        if (parts.length != 2) { return IpType.None; }
+
+        const cidrNum: number = Number.parseInt(parts[1], 10);
+        if (Number.isNaN(cidrNum)) { return IpType.None; }
```
venu-sagar: if i understand correctly, `"1.1.1.1/24abc"` or `"1.1.1.1/ 24"` would parse as CIDR `24` and validate as a valid network address. not strict but permissive. do you think it makes sense to validate with `parts[1]` against `/^\d{1,3}$/` before parsing? just want it to be tightened with stricter regex check.

ui/src/app/shared/utils/inet/inet.utils.ts
```diff
@@ -0,0 +1,172 @@
...
+
+        const ipType = isIP(parts[0]);
```
venu-sagar: isIp returns an enum, not an boolean. "is" convention is generally used for boolean purposes. isIPv4 / isValidIP in the same file are well used, and isIp() just breaks that predicate convention.

ui/src/app/shared/utils/inet/inet.utils.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+        expect(InetUtils.isHostnameOrIp("openems::io")).toBeFalse();
+        expect(InetUtils.isHostnameOrIp(".openems.io")).toBeFalse();
+        expect(InetUtils.isHostnameOrIp("1:1:1:1")).toBeFalse();
+        expect(InetUtils.isHostnameOrIp(null)).toBeFalse();
+    });
+
+    it("#isNetworkAddress", () => {
+        expect(InetUtils.isNetworkAddress("0.0.0.0/0")).toBe(InetUtils.IpType.IPv4);
+        expect(InetUtils.isNetworkAddress("1.1.1.1/24")).toBe(InetUtils.IpType.IPv4);
+        expect(InetUtils.isNetworkAddress("1::1/64")).toBe(InetUtils.IpType.IPv6);;
```
venu-sagar: double semicolon

venu-sagar (changes requested):
Solid refactor. regexes are well-known and tested.

venu-sagar (approved):

PR #3699, opened 2026-04-24
ENTSO-E: add mathematical expression

sfeilmeier:
- Adds Javaluator library and wrapper
- Adds configurable calculateExpression to calculate gross price per period from EPEX spot price and ancillary cost
- Fallback to x + y as before if no expression is given

Relates to https://community.openems.io/t/info-about-json-format-for-ancillary-costs-in-timeofusetarrif-entsoe/10261/15


io.openems.edge.timeofusetariff.entsoe/src/io/openems/edge/timeofusetariff/entsoe/TouEntsoeImpl.java
```diff
@@ -102,6 +103,8 @@ private synchronized void activate(ComponentContext context, Config config) {
 		this.priceProvider.getMarketPrices().subscribe(this.onNewPrices);
 		this.priceProvider.getUpdateState().subscribe(this.onUpdateEvent);
 
+		this.priceCalculator = new PriceCalculator(config.calculateExpression());
```
Howaner: We should make sure to throw an exception here if the expression is invalid.
WIthout that, the exception is thrown a lot of times at every price upgrade.

sfeilmeier: Ok. There is _no_ native way to test the expression, but I introduced a simple calculation on activate to validate it.

Howaner (approved):

PR #3717, opened 2026-05-07
Update modbus device template to new reference target system

simon-lgb:
This PR updates the Eclipse code template for adding Modbus device bundles to use the new GenerateTargetsFromReferences annotation. It also cleans up some imports.


sfeilmeier (approved):
Thank you!

PR #3728, opened 2026-05-13
Add DTSU666 support to Chint meter bundle

tushabe:
## Summary

Add support for the Chint DTSU666 three-phase, four-wire meter in the existing `io.openems.edge.meter.chint` bundle.

This implementation follows the bundle structure and maintainer feedback from the merged DDSU666 work:
- uses `FC4` / input registers
- supports `phaseRotation`
- maps the physical total import-energy register directly to standard OpenEMS energy channels
- keeps bridge serial settings configurable while documenting `9600-N-8-2` as the known-good DTSU666 setup
- keeps `DSSU666` out of scope for this PR

## Implementation

Added a new `Meter.Chint.DTSU666` component with these mappings:
- `0x2006` -> `VOLTAGE_L1/L2/L3`
- `0x200C` -> `CURRENT_L1/L2/L3`
- `0x2012` -> `ACTIVE_POWER` and `ACTIVE_POWER_L1/L2/L3`
- `0x202A` -> DTSU666-specific `TOTAL_POWER_FACTOR`
- `0x2044` -> `FREQUENCY`
- `0x4000` -> standard cumulative import energy channel

Scaling:
- voltage `/10`
- current values are already exposed by the meter in the scaled form needed for OpenEMS `mA`
- active power `/10`
- total power factor is exposed as an integer-style OpenEMS value where `1000 = 1.000`
- frequency `/100`
- energy `kWh -> Wh` through the standard OpenEMS conversion

`invert` is handled in the standard OpenEMS way:
- `invert=false` maps the physical import counter to `ACTIVE_CONSUMPTION_ENERGY`
- `invert=true` maps the same counter to `ACTIVE_PRODUCTION_ENERGY`

## Tests

Added strict tests that cover:
- FC4 input-register decoding
- scaling of voltage/current/power/frequency/energy
- `phaseRotation`
- `invert=false` and `invert=true`
- DTSU666-specific total power factor channel

## Docs

Updated the Chint bundle README to:
- include DTSU666 as a supported device
- document the DTSU666 config surface
- clarify that DTSU666 uses `FC4` / input registers, unlike the DDSU666 single-phase support in this bundle
- clarify that the OSGi attribute is `modbus_id`, while FileInstall/Felix configs typically use `modbus.id`
- document the verified DTSU666 assumptions:
  - `FC4` input registers
  - recommended bridge settings `9600-N-8-2`
  - total import energy at `0x4000`
  - total power factor at `0x202A`

## Scope

This PR adds `DTSU666` only.

The separate `DSSU666` three-phase, three-wire variant is intentionally left for a later change.


tushabe:
@sfeilmeier @da-Kai I am hoping this one will also make it into this month's release.

Sn0w3y:
Thank you! @tushabe did you test it in real life Scenario ot only Vibe-Coded it ?

Also afaik the DTSU Meter hat CT/PT-Ratio Configurable - would be a nice to have if we could set this aswell od even mandatory if CT or PTs are used..

tushabe:
Thank you for the feedback! @Sn0w3y I did both :) 

Tested on real DTSU666 hardware, yes and it looks beautiful. (see screenshot attached)
Screenshot 2026-05-12 at 23 05 48

This PR was validated against a live DTSU666 unit that is wired directly to the line, and that is the variant we are currently using in our pilot microgrid and plan to use in our near-term deployments.

On CT/PT support: I agree that would be useful. At the moment I do not have access to DTSU666 variants deployed with CT/PT instrumentation, so I was not able to verify the relevant behavior or implement that confidently in this PR.

For that reason, this PR is intentionally scoped to the directly wired base variant that we could test on real hardware. I would be happy to handle CT/PT support in a follow-up once we have access to the appropriate hardware or a verified register/config reference for those deployments.

Sn0w3y (commented):
Looks good from my side. Only thing i am not sure about is the change in the Docs. Apart from that thank you again! :)

sfeilmeier (approved):

PR #3743, opened 2026-05-22
Controller Api Modbus: Fix Modify Method

tsicking:
In the previous version, the `modified` method always returned in line 117, as `this.config = config` was set too early. Therefore, if you e.g. modified the timeout, it didn't have any effect (only after restarting the system).


clehne:
> In the previous version, the `modified` method always returned in line 117, as `this.config = config` was set too early. Therefore, if you e.g. modified the timeout, it didn't have any effect (only after restarting the system).

I took me some time to understand the above comment. A look at the code finally helped: 
* L112 sets the config
* L116-117 compares and aborts the code modification
* L120 is the correct location for setting the config. 

see here: https://github.com/opernikus-common/openems/blob/938524631af3150ee471f7e627adba76f09b8a2c/io.openems.edge.controller.api.modbus/src/io/openems/edge/controller/api/modbus/AbstractModbusApi.java#L111-L120

clehne (approved):
Ok.

sfeilmeier (approved):

PR #3753, opened 2026-06-03
Use default case for Unit methods

simon-lgb:
These long lists of all units are difficult to read and frequently cause merge conflicts with some custom units we have in our fork.

While using a default case can generally lead to unhandled enum constants when refactoring, it seems appropriate in these instances where this clearly should be the default behavior for new units.


sfeilmeier (changes requested):
I have a different opinion here. I always prefer explicit ENUM handling, because it guarantees that the IDE or Gradle points the developer to think about this.

To avoid merge conflicts, we could group the ENUMS better, i.e. add some newlines/empty lines.

simon-lgb:
That's a fair point, I will close this for now.

Issue #3765, opened 2026-06-07
Keba P40 Energy Limit Scaling Bug


Frischid:
### Description

Thanks for the many integrations. Unfortunately, I encountered an issue with the Keba P40 integration.

The energy limit is off by a factor of 10. Looking into the code, the issue seems to be already known during code writing.
``` java
new FC6WriteRegisterTask(5010, // TODO Scalefactor for Unit: 10 Wh
		m(EvseKeba.ChannelId.SET_ENERGY_LIMIT, new UnsignedWordElement(5010))),
```
Line 204 in https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.evse.chargepoint.keba/src/io/openems/edge/evse/chargepoint/keba/modbus/EvseKebaModbusImpl.java

I would propose adding a constant like SCALE_FACTOR_0_10 in order to fix this bug.

### How to reproduce the Error?

Using the Keba P40 implementation and setting the energy limit with the openems charging station controller

PR #3766, opened 2026-06-09
Add scale factor for keba p40

wedwincode:
fixes issue #3765 by applying the correct scale factor for the energy limit register (5010) of the keba p40.
the register uses a unit of 10 Wh, so without a scale factor the written value would be interpreted off by a factor of 10.

### DIFF
--- a/io.openems.edge.evse.chargepoint.keba/src/io/openems/edge/evse/chargepoint/keba/modbus/EvseKebaModbusImpl.java
+++ b/io.openems.edge.evse.chargepoint.keba/src/io/openems/edge/evse/chargepoint/keba/modbus/EvseKebaModbusImpl.java
@@ -1,5 +1,6 @@
 package io.openems.edge.evse.chargepoint.keba.modbus;
 
+import static io.openems.edge.bridge.modbus.api.ElementToChannelConverter.SCALE_FACTOR_1;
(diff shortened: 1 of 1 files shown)


Frischid:
If I'm not mistaken, `SCALE_FACTOR_1` scales by a factor of **1** and not by **1/10=0.1**, maybe using mult() instead might work, unfortunately, I currently don't have the time to fix it myself.

wedwincode:
Thanks for checking this. I rechecked the conversion direction, and I think SCALE_FACTOR_1 should be correct here, unless I am misunderstanding the mapper logic.

Since this is a write register, it looks like the converter is used in the reverse direction via channelToElement(). So for SCALE_FACTOR_1, the read direction would be element -> channel = * 10, but the write direction should be channel -> element = / 10.

That seems to match register 5010, because its unit is 10 Wh. An OpenEMS value of 50000 Wh should be written as 5000.

Please correct me if I am missing something here, since I haven't worked much with this converter logic before.

sfeilmeier:
@johannk24 Could you validate this please?

sfeilmeier:
This has been solved in https://github.com/OpenEMS/openems/pull/3863. Sorry, I didn't realize, that there had been two PRs for the same issue.

PR #3768, opened 2026-06-11
Update systemd.adoc: updated jar filename

sjjh:
changed `openems.jar` to `openems-edge.jar` as this is the filename of the (suggested) github build and it is easier to differentiate it from a potential backend jar file.


sjjh:
maybe it's worth to update the filename in https://openems.github.io/openems.io/openems/latest/edge/build.html as well.

sfeilmeier (approved):

PR #3795, opened 2026-06-29
[Edge] Controller.Api.Backend: support Backend 2026.x edge.manager

arindahills:
Backend 2026.x replaced the edgewebsocket module with `edge.manager`, which identifies Edges by an `id` HTTP header and only accepts notifications wrapped in `EdgeRpcNotification`. An Edge connecting directly to a 2026.x Backend currently stays offline, because it sends only the apikey and raw, unwrapped notifications.

This adds an `edgeId` config property to `Controller.Api.Backend`. When it is set:
- the `id` header is sent on the WebSocket handshake;
- every outgoing notification is wrapped in `EdgeRpcNotification`, including the initial `OnOpen` `EdgeConfigNotification`.

When `edgeId` is empty the behaviour is unchanged, so connections to 2025.x Backends keep working.

Tested against Backend 2026.6.0: the Edge connects and streams live data.

Pairs with a separate PR that lets `edge.manager` register a directly-connected Edge.

### DIFF
--- a/io.openems.edge.controller.api.backend/src/io/openems/edge/controller/api/backend/Config.java
+++ b/io.openems.edge.controller.api.backend/src/io/openems/edge/controller/api/backend/Config.java
@@ -25,6 +25,12 @@ import io.openems.common.channel.PersistencePriority;
 	@AttributeDefinition(name = "Apikey", description = "Apikey for authentication at OpenEMS Backend.", type = AttributeType.PASSWORD)
 	String apikey();
 
+	@AttributeDefinition(name = "Edge-ID", description = "Edge ID for Edge.Manager protocol (Backend 2026.2+). "
+			+ "Set to the Edge ID registered in Backend Metadata.File (e.g. 'my-site-pi-1'). "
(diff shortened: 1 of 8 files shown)


tushabe:
@sfeilmeier Please have a review at this PR and https://github.com/OpenEMS/openems/pull/3796 when you have a moment. It's enabling Core direct Edge-to-Backend 2026.x support. We realised this was missing when we were trying to connect edge and backend using 2026.6 release images.

sfeilmeier:
Hi @tushabe and @arindahills,

I think there is a misunderstanding on how Backend Edge Manager works. Unfortunately we did not yet document this, so I try to explain it here. Before, we used to have a EdgeWebsocket-Component inside a central Backend server. This did not scale well, so we split this up to be able to run the Edge-facing service on one or more separate physical servers. This is the general setup:

grafik

There is no need to update the Edge `Controller.Api.Backend`. Of course it's also possible to run `Backend.Edge.App` on the Central Server and connect to localhost.

Does this clarify the structure? 
If you try this and it works for you it would be very helpful if you could help in documenting this feature.

I'll close this specific PR here, but we can continue discussions here or on https://community.openems.io/.

PR #3796, opened 2026-06-29
[Backend] Edge.Manager: register directly-connected Edge so its data persists

arindahills:
`edge.manager` persists data only for Edges present in `WsData.edges`, which is populated exclusively by `ConnectedEdges.Notification` (the aggregator path for multi-Edge deployments).

An Edge connecting directly authenticates via the `id` header in `OnOpen` (`setId(id)`) but is never added to that map. `getEdgeCache(edgeId)` then returns null and `handleDataNotification` silently drops every `TimestampedData` / `AggregatedData` notification (`if (edgeCache == null) return;`).

This registers the directly-connected Edge in `OnOpen` via `updateEdgeStatus(id, true)`, so its `EdgeCache` exists and its data is persisted.

Pairs with a separate Controller.Api.Backend PR that lets an Edge connect directly to a 2026.x Backend.

### DIFF
--- a/io.openems.backend.edge.manager/src/io/openems/backend/edge/manager/OnOpen.java
+++ b/io.openems.backend.edge.manager/src/io/openems/backend/edge/manager/OnOpen.java
@@ -55,6 +55,14 @@ public class OnOpen implements io.openems.common.websocket.OnOpen {
 
 		wsData.setId(id);
 
+		// Register this directly-connected Edge so its data notifications are accepted
+		// and persisted. In the multi-edge topology Edges are announced via a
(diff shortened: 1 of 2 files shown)


tushabe:
Thanks for the clarification on #3795, @sfeilmeier. Since the intended Backend 2026.x setup is Edge -> Backend.Edge.App -> Edge.Manager, I think this direct Edge registration change in Edge.Manager is no longer needed either. Please confirm, and we can close this PR as superseded by the aggregator-based architecture.

sfeilmeier:
Yes, exactly 👍

PR #3797, opened 2026-06-29
[Backend] Metadata.File: implement generateUpdateMetadataCacheNotification (Backend.Edge.App support)

arindahills:
When connecting Edges through the `Backend.Edge.App` aggregator (the supported architecture, see #3795 and #3818), the app needs the API-key to Edge-ID mapping from the central Backend's Metadata service. `MetadataOdoo` and `MetadataDummy` implement `generateUpdateMetadataCacheNotification()`, but `MetadataFile` inherited the empty default, so a File-metadata Backend cannot serve the aggregator (the cache never initializes and Edges are rejected).

This implements it for `MetadataFile`, mirroring `MetadataOdoo`/`MetadataDummy`, so deployments that use File-based Metadata (i.e. without running Odoo) can use the Backend.Edge.App aggregator.


tushabe:
@sfeilmeier Could you please help review this PR when you have a moment? It adds Metadata.File support for generating the UpdateMetadataCache notification, so Backend.Edge.App can get the apikey -> edgeId mapping when running with file-based metadata instead of Odoo.

PR #3798, opened 2026-06-29
[Backend/Edge] Make resend (backfill) work with a plain Timedata.InfluxDB

arindahills:
Backfill of historic data after a reconnect did not work end-to-end with a plain `Timedata.InfluxDB` backend (only `AggregatedInflux` handled it). Two independent gaps, fixed here:

1. **Backend** `Timedata.InfluxDB` did not implement the `ResendDataNotification` write, so resent data was silently ignored. Implemented it to write resent points at their original timestamps, mirroring the timestamped write.

2. **Edge** `Timedata.Rrd4j.queryResendData` silently dropped AVERAGE-consolidated channels (Voltage, Current, Frequency, ActivePower): the `adjustSeconds = arcStep - 300` shift is 0 for the 300s AVERAGE archive, so the narrow (<=300s) resend window left slots NaN or just outside the window and only cumulated (3600s) channels survived. Now reads each channel at the archive matching its consolidation and native step (mirroring `queryHistoricData`), and coerces the rrd4j double back to the channel's `OpenemsType` so integer/boolean channels are written with the same field type as live data (otherwise a float-vs-integer field conflict makes InfluxDB drop the whole resent point).

Verified on Backend 2026.6.0 + a real Edge over a controlled reconnect: Voltage, Current, Frequency and Energy backfill into the gap at 5-minute resolution with correct types and no field-type conflicts.

### DIFF
--- a/io.openems.backend.timedata.influx/bnd.bnd
+++ b/io.openems.backend.timedata.influx/bnd.bnd
@@ -10,6 +10,7 @@ Bundle-Version: 1.0.0.${tstamp}
 	io.openems.shared.influxdb,\
 	io.openems.wrapper.influxdb-client-core,\
 	io.openems.wrapper.influxdb-client-java,\
+	io.openems.wrapper.influxdb-client-utils,\
 
 -testpath: \
 	${testpath}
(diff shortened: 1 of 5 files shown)


tushabe:
@sfeilmeier Could you please help review this PR when you have a moment? It makes resend/backfill work with a plain Timedata.InfluxDB setup: Backend writes ResendDataNotification data, and Edge RRD4J resend reads AVERAGE channels with the correct serialized field types.

PR #3807, opened 2026-07-02
[Edge] io.openems.edge.victron: drop all write tasks in Read-Only-Mode

jankaifer:
I was seeing this issues: https://github.com/OpenEMS/openems/issues/3815

This PR should resolve it. Not sure if I fixed it in the right place.

The read-only flag was used in `applyPower` but to late there were already some writes happening before we checked the flag so I moved it to the start of that function to make sure there are no writes happening.

I also added some tests to verify that we are properly not writing anything in read-only mode.


io.openems.edge.victron/src/io/openems/edge/victron/ess/VictronEssImpl.java
```diff
@@ -956,7 +958,15 @@ protected ModbusProtocol defineModbusProtocol() { // Unit-ID 227
 								this.m(VictronEss.ChannelId.ESS_DISABLE_CHARGE_FLAG, new SignedWordElement(38)),
 								this.m(VictronEss.ChannelId.ESS_DISABLE_FEEDBACK_FLAG, new SignedWordElement(39)))
 
-		);
+		) {
+			@Override
+			public synchronized void addTask(Task task) {
+				if (VictronEssImpl.this.config.readOnlyMode() && task instanceof WriteTask) {
```
sfeilmeier: I don't think we should do it like this. There is already similar handling in `applyPower(int activePowerTarget, int reactivePower)`. I think we can copy this approach to `applyPower(int activePowerTargetL1, int reactivePowerTargetL1, int activePowerTargetL2, int reactivePowerTargetL2, int activePowerTargetL3, int reactivePowerTargetL3)` - while I am actually not sure, why both methods are implemented.

```
if (this.config.readOnlyMode()) {
	this.logDebug(this.log, "Read Only Mode is active. Power is not applied");
	return;
}
```

sfeilmeier (changes requested):

io.openems.edge.victron/src/io/openems/edge/victron/ess/VictronEssImpl.java
jankaifer: Thanks for taking a look. I originally tried to change it over-there but it required more investigation because I'd need to change the order in which `applyPower` does things (or add multiple `if`s for this).

I can try to investigate some more and apply it nicely into `applyPower`.

But either way I think it would be nice to have it here at least as an assertion so that when similar regression slips where the shadow mode starts messing up with the modbus it would just fail instead of actually writing through. I spent two days debugging my Victron system until I figured out this bug in Open EMS where it would mess-up my Victron setup even though it was in read-only mode.

Is that reasonable from your pov?

sfeilmeier: I understand your point, but I still don't really like it, because it would just hide bugs. In fact this should be secured by JUnit tests. 

Should I interfere here and try to improve the code myself? I did not look into details and have no Victron myself, though.

jankaifer: No worries I'll dig around some more. Will let you know if I'm not able to figure it out.

> I understand your point, but I still don't really like it, because it would just hide bugs. In fact this should be secured by JUnit tests.

Good point I'll add a test covering this issue.

And I'd still like to have an assertion that would just fail and produce an error instead of writing in a read-only mode. In extreme even crashing the whole thing would be better than writing in a read-only mode IMO. Either way I'm new around here so not my call. I'll code it up so that you see what I mean and you can decide whether having such assertion make sense to you.

Specifically I want to run multile OpenEMS instances on my system in shadow mode for benchmarking them against production instance and I'll run my fork with this assertion included.

jankaifer: You were right to push back, thanks for that. I re-read the code again and the place where I did it before was running during setup and not during run-time so it would not properly error on write.

sfeilmeier (approved):

PR #3808, opened 2026-07-02
[Edge] io.openems.edge.victron: keep configured capacity when BMS reports 0

jankaifer:
## Summary

`VictronEssImpl` re-sets the ESS `Capacity` channel from `battery.getCapacity()` on every cycle:

```java
if (this.battery != null) {
    this._setCapacity(this.battery.getCapacity().get());
}
```

Some BMS report the installed-capacity register as `0`/absent (observed on a Pylontech stack). That collapses the ESS `Capacity` channel to `0`, which in turn zeroes the usable energy the Time-of-Use optimizer plans against, producing a degenerate schedule.

## Fix

Fall back to the configured `Config.capacity()` when the battery reports `0` or `null`; otherwise use the battery-reported value as before.


sfeilmeier (approved):
Thank you. The Victron binding was coming from the Community and the code quality is not too good unfortunately. Thanks for your contributions!

PR #3818, opened 2026-07-09
[Doc] Backend: document the Edge connection architecture (Backend.Edge.App)

arindahills:
The backend *Architecture* page was empty. This documents how an OpenEMS Edge connects to the Backend:

- the Edge-facing endpoint was split out of the central Backend into its own scalable application, `Backend.Edge.App`;
- an Edge always connects (via its unchanged `Controller.Api.Backend`) to a `Backend.Edge.App`, which relays to the central `Edge.Manager`;
- deployment options: one or more `Backend.Edge.App` on separate servers, or a single one on the central server via `localhost`.

This follows up on the discussion in #3795, where this setup was noted as not yet documented. Happy to adjust the wording or add a diagram image if preferred.

### DIFF
--- a/doc/modules/ROOT/pages/backend/architecture.adoc
+++ b/doc/modules/ROOT/pages/backend/architecture.adoc
@@ -1,2 +1,59 @@
 = Backend Architecture
+:sectnums:
+:sectnumlevels: 4
+:toclevels: 4
+:experimental:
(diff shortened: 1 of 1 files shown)


sfeilmeier (approved):
Thank you, this is very helpful! 👍

PR #3824, opened 2026-07-13
Fix null body for http error responses

simon-lgb:
This PR fixes the body of http error responses always being null in the http bridge.

The behavior is not documented clearly, but getInputStream() always throws an IOException when the server returned an error response. If a body was still sent, getErrorStream() has to be called to read it.

The error stream can be null, so an additional check was added for this. Any remaining IOExceptions thrown when attempting to read the body will now cause an UnknownError instead of a ResponseError.

### DIFF
--- a/io.openems.common.bridge.http/src/io/openems/common/bridge/http/NetworkEndpointFetcher.java
+++ b/io.openems.common.bridge.http/src/io/openems/common/bridge/http/NetworkEndpointFetcher.java
@@ -88,12 +88,16 @@ public class NetworkEndpointFetcher implements EndpointFetcher {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


simon-lgb:
Rebased onto the changes in #3825

michaelgrill (commented):
I also tested it with a keycloak endpoint. 

Tried with response codes:
200 => success 
and 400, 500 => error

Is there anything left from your side @simon-lgb, or is it ready for merge?

simon-lgb:
Thanks for the review @michaelgrill , this is ready to merge from my side.

michaelgrill (approved):

sfeilmeier (approved):

PR #3866, opened 2026-08-04
Backend AppCenterHandler: return empty results when no AppCenter metadata service is present

schlabsch:
## Problem

Deployments without an AppCenter metadata service — e.g. `Metadata.File` based setups, which have no Odoo/AppCenter backend — receive a hard JSON-RPC error for every `appCenter/*` request:

```
Objects.requireNonNull(metadata, "No AppCenter Metadata provided.");
```

Edges send `getInstalledApps` periodically and the UI issues appCenter requests on edge pages, so backend logs fill with `JSON-RPC Error "No AppCenter Metadata provided." of type NullPointerException`, and the 2026.x UI can crash when navigating away from pages that triggered these requests (observed on 2026.8.0: UI breaks when returning from the edge app view on a file-metadata deployment).

## Fix

When the injected `AppCenterMetadata` is `null`, `AppCenterHandler` now returns benign empty results instead of throwing:

- `getInstalledApps` → `{ "installedApps": [] }`
- `getPossibleApps` → `{ "bundles": [] }`
- `getRegisteredKeys` → `{ "keys": [] }`
- `isKeyApplicable` → `{ "isKeyApplicable": false, "additionalInfo": {} }`
- `isAppFree` → `{ "isAppFree": false }`
- history/write requests → generic empty success (no-op)

Clients degrade gracefully: the UI shows an empty App Center, Edges get valid empty responses, logs stay clean. Verified on a live 2026.8.0 deployment (file metadata, 6 edges): the UI navigation crash disappears and the appCenter requests return the empty results above.

Related: #3865 (same deployment class — file-based metadata with the split backend / backend-edge topology).

### DIFF
--- a/io.openems.backend.common/src/io/openems/backend/common/metadata/AppCenterHandler.java
+++ b/io.openems.backend.common/src/io/openems/backend/common/metadata/AppCenterHandler.java
@@ -1,6 +1,7 @@
 package io.openems.backend.common.metadata;
 
-import java.util.Objects;
+import com.google.gson.JsonArray;
+import com.google.gson.JsonObject;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
@@ -31,6 +32,7 @@ import io.openems.common.jsonrpc.response.AppCenterGetPossibleAppsResponse;
 import io.openems.common.jsonrpc.response.AppCenterGetRegisteredKeysResponse;
 import io.openems.common.jsonrpc.response.AppCenterIsAppFreeResponse;
 import io.openems.common.jsonrpc.response.AppCenterIsKeyApplicableResponse;
+import io.openems.common.utils.JsonUtils;
 
 public final class AppCenterHandler {
 
@@ -53,7 +55,12 @@ public final class AppCenterHandler {
 			final User user, //
 			final String edgeId //
 	) throws OpenemsNamedException {
-		Objects.requireNonNull(metadata, "No AppCenter Metadata provided.");
+		if (metadata == null) {
+			// OrbitEOS patch: no AppCenter metadata service in this deployment.
+			// Answer with benign empty results instead of an error so UIs and
+			// Edges degrade gracefully.
+			return handleWithoutMetadata(request);
+		}
 
 		return switch (request.getPayload().getMethod()) {
 		case AppCenterAddRegisterKeyHistoryRequest.METHOD -> handleAsync(metadata, request, //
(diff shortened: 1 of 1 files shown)


io.openems.backend.common/src/io/openems/backend/common/metadata/AppCenterHandler.java
```diff
@@ -53,7 +55,12 @@ public static CompletableFuture<? extends JsonrpcResponseSuccess> handleUserRequ
 			final User user, //
 			final String edgeId //
 	) throws OpenemsNamedException {
-		Objects.requireNonNull(metadata, "No AppCenter Metadata provided.");
+		if (metadata == null) {
+			// OrbitEOS patch: no AppCenter metadata service in this deployment.
```
Sn0w3y: OrbitEOS Reference should be removed.

io.openems.backend.common/src/io/openems/backend/common/metadata/AppCenterHandler.java
```diff
@@ -130,7 +137,10 @@ public static CompletableFuture<? extends JsonrpcResponseSuccess> handleEdgeRequ
 			final AppCenterRequest request, //
 			final String edgeId //
 	) throws OpenemsNamedException {
-		Objects.requireNonNull(metadata, "No AppCenter Metadata provided.");
+		if (metadata == null) {
+			// OrbitEOS patch: see handleUserRequest.
```
Sn0w3y: OrbitEOS should be removed.

Sn0w3y (changes requested):

sfeilmeier:
@michaelgrill: I'd like to have your opinion on this patch. Seems legit to me.
@schlabsch: Is this you? https://orbiteos.com/. It would be great to welcome you as a member of the OpenEMS Association (https://openems.io/association/). It's not expensive and helps the OpenEMS project a lot. We are also working on similar AI integrations, so maybe we could even collaborate on that.

io.openems.backend.common/src/io/openems/backend/common/metadata/AppCenterHandler.java
schlabsch: Done — removed.

io.openems.backend.common/src/io/openems/backend/common/metadata/AppCenterHandler.java
schlabsch: Done — removed here as well.

PR #3873, opened 2026-08-09
Ess.SaxPower: add single-phase ESS and integrated grid meter

Hans-4:
Implements the SAX Power Home Plus (5.8 / 7.7 kWh) as two components:
- Battery read and write functions
- Built-in grid meter

Modbus register documentation and manual:
https://sax-power.net/wp-content/uploads/2026/08/Installationsanleitung_58kWh_77kWh_1.4.pdf


Sn0w3y:
In my Tests the Sax Battery (AC) has been very unreliable in terms of controlling it. Did you encounter the same Issues? It has been slow in Charge and Discharge Response.

Hans-4:
At the moment I can't test the setpoint power P because I don't have my OpenEMS
completely set up. As I know I have to contact the support every time I want to
change the mode, or have to ask for an installateur account. Or is there an easy
way to switch the mode?

I only tested setpoint power p in the simulator and there it worked very well.

Hans-4:
@Sn0w3y  I’ve now unlocked my battery and, in my tests, it responded quite quickly. Not immediately, but quickly.

Sn0w3y:
> @Sn0w3y  I’ve now unlocked my battery and, in my tests, it responded quite quickly. Not immediately, but quickly.

see my discussion here:

https://github.com/evcc-io/evcc/discussions/22155

to be hones i don't think they fixed this issue already?

Hans-4:
No you are right @Sn0w3y . I completely overlooked that in my first test. 

 Its exactly the same as in the discussion. After 2–3 minutes, it stops writing and remains at the current discharge/charge level.

`026-08-11T20:43:32,427 [modbus0 ] ERROR [e.modbus.api.task.AbstractTask]   Execute failed FC16WriteRegisters [ess0;unitid=64;ref=41/0x29;length=1;request=3a88] ModbusIOException: Executing transaction 01 96 00 00 00 09 40 10 00 29 00 01 02 3A 88 failed (tried 2 times) Executing transaction failed (tried 2 times)`

What sould i do now? A read mode only switch as long as they don’t release an update? I think it still works when you got it in mode 1 you just get an error saying it couldnt write, but you can still read all the read addresses.

image

Sn0w3y:
> No you are right @Sn0w3y . I completely overlooked that in my first test. 
> 
>  Its exactly the same as in the discussion. After 2–3 minutes, it stops writing and remains at the current discharge/charge level.
> 
> `026-08-11T20:43:32,427 [modbus0 ] ERROR [e.modbus.api.task.AbstractTask]   Execute failed FC16WriteRegisters [ess0;unitid=64;ref=41/0x29;length=1;request=3a88] ModbusIOException: Executing transaction 01 96 00 00 00 09 40 10 00 29 00 01 02 3A 88 failed (tried 2 times) Executing transaction failed (tried 2 times)`
> 
> What sould i do now? A read mode only switch as long as they don’t release an update? I think it still works when you got it in mode 1 you just get an error saying it couldnt write, but you can still read all the read addresses.
> 
> image
> 

Hi,

tbh o can not answer that question - defacto a "read only" implementation is not really helpful for users without Grid Meter and "only" OpenEMS - i tried to reach out to the Support more than once and they do not really care about it so I gave this Project a stop.

Greetings!

Hans-4:
Yeah, you’re probably right, and we might have to wait ages for the update or it might never come. They are even selling their own power manager, so why would they?

Sn0w3y:
> Yeah, you’re probably right, and we might have to wait ages for the update or it might never come. They are even selling their own power manager, so why would they?

Well.. you see thats the Point. Fenecon also sells their FEMS and their Storage and still provides a free OpenSource alternative to FEMS and their Storages. Others just try to sell and not Improve 😉

sfeilmeier:
So what does this mean? Should we proceed with this PR or wait for SAX Power?

Hans-4:
I dont know @sfeilmeier . We can proceed, and if you like, I can add a warning to the README saying that address 41 doesnt work, or we can wait for Sax. Whichever suits you best.

sfeilmeier:
From my side we can proceed 👍

Hans-4:
Could someone please restart the workflows? Had to fix one checkstyle error.

Hans-4:
Ive added several tests they should now run if someone restarts the workflows

Sn0w3y:
> I dont know @sfeilmeier . We can proceed, and if you like, I can add a warning to the README saying that address 41 doesnt work, or we can wait for Sax. Whichever suits you best.

It is not only Register 41.. the whole Controlling Mechnanism from SAX side has this Issue.

io.openems.edge.ess.saxpower/src/io/openems/edge/ess/saxpower/ess/Config.java
```diff
@@ -0,0 +1,55 @@
...
+            description = "Maximum discharge power in W, as a positive value. " //
+                    + "Must match the value configured in the SAX dashboard. " //
+                    + "See the readme for a guide.")
+    int maxDischargePower() default 4600;
+
+    @AttributeDefinition(name = "Min SoC [%]", description = "Discharging is blocked while State of Charge is below Min-SoC.")
+    int minSoc() default 15;
+
+    @AttributeDefinition(name = "Modbus target filter", description = "This is auto-generated by 'Modbus-ID'.")
+    String Modbus_target() default "(enabled=true)";
```
Howaner: The Modbus_target here can be removed because that's automaticly resolved by ```@GenerateTargetsFromReferences``` annotation on SaxPowerImpl.  
Take a look here: https://github.com/OpenEMS/openems/pull/3889

io.openems.edge.ess.saxpower/src/io/openems/edge/ess/saxpower/ess/SaxPowerImpl.java
```diff
@@ -0,0 +1,214 @@
...
+                "Modbus", config.modbus_id())) {
+            return;
+        }
+
+        SinglePhaseEss.initializeCopyPhaseChannel(this, this.phase);
+
+        this._setCapacity(this.config.capacity());
+        this._setMaxApparentPower(MAX_APPARENT_POWER);
+
+        this.getGridModeChannel().setNextValue(GridMode.ON_GRID);
```
Howaner: Is there no way to read the power state from the sax power device?

io.openems.edge.ess.saxpower/src/io/openems/edge/ess/saxpower/ess/SaxPowerImpl.java
```diff
@@ -0,0 +1,214 @@
...
+@Component(//
+    name = "Ess.SaxPower", //
+    immediate = true, //
+    configurationPolicy = ConfigurationPolicy.REQUIRE //
+)
+@GenerateTargetsFromReferences("Modbus")
+public class SaxPowerImpl extends AbstractOpenemsModbusComponent
+        implements SaxPower, ManagedSinglePhaseEss, SinglePhaseEss, ManagedAsymmetricEss, AsymmetricEss, ManagedSymmetricEss, SymmetricEss, OpenemsComponent, ModbusComponent, ModbusSlave {
+
+    private static final int ACTIVE_POWER_OFFSET = 16384;
```
Howaner, line 54: Is this really a offset or is it some kind of bit flag? If it's a bit flag, applying a bitmask would be better instead of subtracting

io.openems.edge.ess.saxpower/src/io/openems/edge/ess/saxpower/ess/SaxPowerImpl.java
```diff
@@ -0,0 +1,214 @@
...
+        this._setMaxApparentPower(MAX_APPARENT_POWER);
+
+        this.getGridModeChannel().setNextValue(GridMode.ON_GRID);
+
+
+        this.getSocChannel().onSetNextValue(soc -> {
+            final Integer allowedCharge;
+            final Integer allowedDischarge;
+            if (soc.isDefined()) {
+                allowedCharge = soc.get() >= 100 ? 0 : -this.config.maxChargePower();
```
Howaner, line 117: Have you tested the SoC reporting from SAX power? Do they stop consuming energy at 100% SoC or are they still charging, but at low current?

Howaner:
Nice to see a implementation for SAX power - it's really interesting to produce AC 230V Sinus Waves with Battery Cells and Mosfets without the need for an inverter.  

@juergen-weber mentioned in the EVCC discussion (https://github.com/evcc-io/evcc/discussions/22155#discussioncomment-17474467) that he got a new firmware installed on his battery and it seems like they have a new modbus protocol (https://github.com/user-attachments/files/29573813/Modbus._Sunspec_Dokumentation.pdf). Maybe that is fixing the modbus problem?

Hans-4:
I'm currently on vacation so I will take a look on it when I'm back on Friday.

io.openems.edge.ess.saxpower/src/io/openems/edge/ess/saxpower/ess/SaxPowerImpl.java
Hans-4: Yes, Im pretty sure. Thats the number you get when the battery has 0 watt power. I also found that number in this forum.
https://forum.iobroker.net/topic/60831/gel%C3%B6st-modbus-tcp-verbindung-mit-sax-batterie

io.openems.edge.ess.saxpower/src/io/openems/edge/ess/saxpower/ess/SaxPowerImpl.java
Hans-4: In normal mode, it stops charging completely when it reaches 100 per cent
image

Hans-4:
As @Howaner  pointed out, I’ve removed Modbus_target. I also added another test that should bring the Grid Meter file  up to 100% code cov. Its actually my first time working with code cov and Im wondering what percentage of coverage is actually required? Ive noticed that many PRs are accepted even if the code cov test is failing, so if the current coverage is enough, I think the PR can be reviewed.

Howaner:
As I understood, the current implementation stops working after a few minutes because of the transaction_id limitation, right?  
How do we want proceed?
A) Add a way to modify the transaction_id of our modbus frames and cap them to 8-bit. I can help you with that.
B) Implement the newly updated SAX-Power protocol specification and hope everyone gets this update

Hans-4:
> B) Implement the newly updated SAX-Power protocol specification and hope everyone gets this update

You mean the protocol @LaRaven97 an the firmware update @juergen-weber mentioned in the [evcc-io discussion](https://github.com/evcc-io/evcc/discussions/22155)?

I will get in touch with the customer support to get the firmware updated and then test the new Sunspec protocol.
> Die Schnittstelle wie hier beschrieben, steht ab der Firmwarekombination Master V61 und Gateway
V54 zur Verfügung!

If this works i would prefer option B. 

image

Hans-4:
I go my battery updated and tested it with a little Python script, and the write error is  gone. I will experiment a bit more with the new register and, if everything works out, integrate it into OpenEMS.

Hans-4:
Ok, @Howaner   Im almost done. Ive left it running for a while and have’t come across any more  write errors. You can have a look at it when you get a chance. What Ive noticed, though, is that the grid meter always displays at least two-digit numbers – so 0, 10, 20 and so on. But i think you cant fix that. But if youve got any ideas about that, do let me know.

Hans-4:
This is from a short test run.
image
The core functions all work.
What I noticed is that with the Controller Ess Balancing, the battery reacts a bit slow. It might improve when configuring a better controling setup, but I can't test it long-term right now.

The green line is the grid power:
image

Normal Mode:
image

Hans-4:
But when charging with fix active power the battery power is stable.
image

But for example. To change the power from 200 W to 800 W the battery needs about 30 seconds.
Was that that issue you also noticed @Sn0w3y?

What do you think about this @sfeilmeier?

Howaner:
Can you test if some ramp is applied and the active power is increased by a few watts every few seconds after sending the power or does it just wait 30 seconds and then it jumps directly to 800 W? It's a bit hard to see it in the graph.

If it's applied after 30 seconds, can you test if it's always exactly 30 seconds?  
Would be interesting if they have some fixed delay (Received fix power request -> wait 30 seconds -> apply) or if they have some cycle internally (every 30 seconds -> apply fix power)

io.openems.edge.ess.saxpower/src/io/openems/edge/ess/saxpower/gridmeter/SaxPowerEssGridMeterImpl.java
```diff
@@ -0,0 +1,122 @@
...
+)
+@GenerateTargetsFromReferences("Modbus")
+public class SaxPowerEssGridMeterImpl extends AbstractOpenemsModbusComponent
+        implements SaxPowerEssGridMeter, ElectricityMeter, OpenemsComponent, ModbusComponent, ModbusSlave {
+
+    private final int gridPowerAddress = AddressList.GRID_POWER.getAddress();
+    private final int gridPowerL1Address = AddressList.GRID_POWER_L1.getAddress();
+    private final int gridPowerL2Address = AddressList.GRID_POWER_L2.getAddress();
+    private final int gridPowerL3Address = AddressList.GRID_POWER_L3.getAddress();
+    private final int gridPowerScaleFactorAddress = AddressList.GRID_POWER_SCALE_FACTOR.getAddress();
```
Howaner, line 47: Are the local variables required? You can directly use ```AddressList.GRID_POWER.getAddress()``` instead of this local variable.

But do you need the ```AddressList``` class at all? In other entities, we are directly using the register number without constant fields.
I know that it's not good to use "magic numbers" in code, but these constants are hardly helpful.

io.openems.edge.ess.saxpower/src/io/openems/edge/ess/saxpower/ApplyScaleFactor.java
```diff
@@ -0,0 +1,44 @@
+package io.openems.edge.ess.saxpower;
+
+import io.openems.edge.bridge.modbus.api.ElementToChannelConverter;
+
+public class ApplyScaleFactor {
```
Howaner, line 5: Do I understand it correctly that we need to do the following, because the scale factor is dynamic?
1) Read scale factor register (e.g. GRID_POWER_SCALE_FACTOR) for a specific register
2) Read the register (ACTIVE_POWER) and apply the read scale factor to it?

io.openems.edge.ess.saxpower/src/io/openems/edge/ess/saxpower/ess/ControlMode.java
```diff
@@ -0,0 +1,36 @@
+package io.openems.edge.ess.saxpower.ess;
+
+import io.openems.common.exceptions.OpenemsError;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ControlMode {
```
Howaner, line 7: What is this doing?

io.openems.edge.ess.saxpower/src/io/openems/edge/ess/saxpower/ApplyScaleFactor.java
Hans-4: Yes, the function applys the different scale factors, but also converts the sign from the read number if necessary. For example Grid power is converted because when the battery says it's feeding into the grid openems thinks it's getting power from the grid.

io.openems.edge.ess.saxpower/src/io/openems/edge/ess/saxpower/gridmeter/SaxPowerEssGridMeterImpl.java
Hans-4: I thought that would make the code easier to read, but if you suggest to just use the raw Number I can change that.

io.openems.edge.ess.saxpower/src/io/openems/edge/ess/saxpower/ess/ControlMode.java
Hans-4: The battery neads to set the control mode to 1 to switch it. The timeout is the time after the battery switches to normal mode when it dosent get write signal. The function makes sure that if there is a connection error or something like that the battery goes back into control mode without restarting openems.

Hans-4:
> Can you test if some ramp is applied and the active power is increased by a few watts every few seconds after sending the power or does it just wait 30 seconds and then it jumps directly to 800 W? It's a bit hard to see it in the graph.
> 
> If it's applied after 30 seconds, can you test if it's always exactly 30 seconds? Would be interesting if they have some fixed delay (Received fix power request -> wait 30 seconds -> apply) or if they have some cycle internally (every 30 seconds -> apply fix power)

No its not a fixt 30 seconds delay. When i apply the new power, in this case from 300 to 200 W it  jumps after 5 seconds to 236 W and after 5 more seconds to 200 W.

When i want tho change the power from 0 W to 500 W the battery neads 30 Seconds with six 5 second blocks: 6 W, 69 W, 188 W, 288 W, 407 W, 491 W, 503 W

For 500 W to 200 W its 20 seconds with four 5 second blocks: 499 W, 452 W, 344 W, 236 W, 198 W.

So we have always this 5 second blocks where the power slowly increases or drops.

Howaner:
What happens when you send a large power setpoint? For example from 100W to 2000W?
Doee it increase a lot more per 5s block or does it take forever?

io.openems.edge.ess.saxpower/src/io/openems/edge/ess/saxpower/gridmeter/SaxPowerEssGridMeterImpl.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    private final ApplyScaleFactor applyScaleFactor = new ApplyScaleFactor();
+    private final CheckScaleFactorMap checkScaleFactorMap = new CheckScaleFactorMap();
+
+    @Override
+    protected ModbusProtocol defineModbusProtocol() {
+        return new ModbusProtocol(this,
+                new FC3ReadRegistersTask(this.gridPowerAddress, Priority.HIGH, //
+                        m(ElectricityMeter.ChannelId.ACTIVE_POWER, new SignedWordElement(this.gridPowerAddress), this.applyScaleFactor.createScalingConverter(this.gridPowerAddress, -1)), //
```
sfeilmeier: Is there a reason why you are not following the OpenEMS standard here? Could we remove `AddressList` and `Scales` and directly define everything here inline - following the approach described in our Getting Started Guide? (https://openems.github.io/openems.io/openems/latest/edge/implement.html#_implement_the_openems_component)

io.openems.edge.ess.saxpower/test/io/openems/edge/ess/saxpower/gridmeter/SaxPowerEssGridMeterImplTest.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class SaxPowerEssGridMeterImplTest {
+
+    @Test
+    public void test() throws Exception {
+        new ComponentTest(new SaxPowerEssGridMeterImpl())
+                .addReference("cm", new DummyConfigurationAdmin())
```
sfeilmeier: Please add one test that activates strict mode (`activateStrictMode()`). This way we can make sure Modbus registers are correctly mapped and scaled to Channels

sfeilmeier (changes requested):
Thank you for the contribution. I would like to merge this ASAP, because I consider it a very valuable integration. The code currently appears "largely vibe-coded" to me, as it is following some non-standard approaches. Do you have time to work on that?

Hans-4:
> Thank you for the contribution. I would like to merge this ASAP, because I consider it a very valuable integration. The code currently appears "largely vibe-coded" to me, as it is following some non-standard approaches. Do you have time to work on that?

Sure, I can continue to work on that.
The first approach with the broken protocol was mostly vibe-coded because I had no idea about Open EMS and very little Java experience. In the SunSpec protocol, I tried to do as much by myself as I could, but I understand why you think that.

Hans-4:
I switch to the more standard approach and add the test as requested. I hope its better now but let me know if theres anything else I should improve.

sfeilmeier (approved):
Ready to go 🎉🚀

Hans-4:
> What happens when you send a large power setpoint? For example from 100W to 2000W? Doee it increase a lot more per 5s block or does it take forever?

No its just very slow. I tested your example and it was about 1 minute and 40 seconds with about 100 W jumps. (Its not 100 W fix more in a range between 90 W - 130 W mostly more in the upper part of the range)
I think it dosent matter which power you chose, its always increasing with this 100 W jumps per 5 seconds block.

PR #3903, opened 2026-08-21
Feature/controller ess cycle surplusfeedtogrid selltogridlimit reference target

PunugupatiSaikumar:
Migrates three ESS controllers to the Reference Target Plugin (`@GenerateTargetsFromReferences`).
Bundles:
- Controller.Ess.Cycle
- Controller.Ess.Hybrid.SurplusFeedToGrid
- Controller.Ess.SellToGridLimit
For each bundle:
- Replace `OpenemsComponent.updateReferenceFilter()` with a static target filter on `@Reference`
- Remove `ConfigurationAdmin` where it was only used for that
- Remove `ess_target()` / `meter_target()` from Config and MyConfig where they existed
- Update tests to JUnit Jupiter and drop `DummyConfigurationAdmin`
SellToGridLimit uses two references (`ess` and `meter`). Cycle and SurplusFeedToGrid each use a single `ess` reference.


io.openems.edge.controller.ess.cycle/src/io/openems/edge/controller/ess/cycle/ControllerEssCycleImpl.java
```diff
@@ -68,23 +70,18 @@ public ControllerEssCycleImpl() {
 	@Activate
 	private void activate(ComponentContext context, Config config) {
```
sfeilmeier, line 71: Please remove `ComponentContext context` as well, it's not required anymore

sfeilmeier (changes requested):

io.openems.edge.controller.ess.cycle/src/io/openems/edge/controller/ess/cycle/ControllerEssCycleImpl.java
PunugupatiSaikumar: Hi Stefan, I looked into this but context is still consumed on the next line by super.activate(context, config.id(), config.alias(), config.enabled()). AbstractOpenemsComponent only provides that one overload (the activate(String id) variant throws "Use the other activate() method."), and the stored context is what drives servicePid() and the _PROPERTY_* channels via MetaTypeService. All migrated components in develop,  e.g. SimulatorPvInverterImpl and the bosch.bpts5hybrid ones, still use activate(ComponentContext context, Config config). Did you mean removing it somewhere else, or should AbstractOpenemsComponent get a context-free activate(...) overload first?

sfeilmeier: Oh, you are right. I missed that - sorry!

sfeilmeier (approved):

sfeilmeier (approved):

PR #3915, opened 2026-08-25
[Edge] Add Alfen EVSE chargepoint support

Sn0w3y:
Introduced support for Alfen NG9xx charging station series in the OpenEMS Edge. This includes new configuration files, implementation classes, and tests for the Alfen EVSE chargepoint. Updated the EdgeApp.bndrun to include the new Alfen chargepoint component. This addition enables communication with Alfen devices via Modbus/TCP, supporting various charging and monitoring functionalities.

Implementation is based on this Modbus mapping:
[modbus-register-mapping.pdf](https://github.com/user-attachments/files/31466889/modbus-register-mapping.pdf)


sfeilmeier:
@Sn0w3y Thanks! Did you actually test this with hardware? At FENECON we also have this on our roadmap and have one in our lab.

Sn0w3y:
> @Sn0w3y Thanks! Did you actually test this with hardware? At FENECON we also have this on our roadmap and have one in our lab.

My pleasure - yes, has been tested (even not all functions but basic functions) @sjjh

sjjh:
@sfeilmeier we have two around here. I tested setting values via API, as well as setting different charge modes (force and no) with a Renault Zoe attached.

Would be very cool, if we could get it upstream for 2026.9 :tada:

io.openems.edge.evse.chargepoint.alfen/test/io/openems/edge/evse/chargepoint/alfen/EvseAlfenImplTest.java
```diff
@@ -0,0 +1,424 @@
...
+						.withRegisters(308, voltageL2) // VOLTAGE_L2_RAW
+						.withRegisters(310, voltageL3) // VOLTAGE_L3_RAW
+						.withRegisters(312, floatToRegisters(399.5f)) // VOLTAGE_L1_L2
+						.withRegisters(314, floatToRegisters(400.1f)) // VOLTAGE_L2_L3
+						.withRegisters(316, floatToRegisters(398.8f)) // VOLTAGE_L3_L1
+						.withRegisters(318, currentN) // CURRENT_N
+						.withRegisters(320, currentL1) // CURRENT_L1_RAW
+						.withRegisters(322, currentL2) // CURRENT_L2_RAW
+						.withRegisters(324, currentL3) // CURRENT_L3_RAW
+						.withRegisters(326, floatToRegisters(31.5f)) // CURRENT_SUM
```
sfeilmeier: There would have been `withRegistersFloat32()` which your AI probably didn't see... ;-)

sfeilmeier (commented):
Yes, I try to have this PR ready before we internally start developing so that there are not two versions eventually.

io.openems.edge.evse.chargepoint.alfen/src/io/openems/edge/evse/chargepoint/alfen/EvseAlfenImpl.java
```diff
@@ -0,0 +1,529 @@
...
+			var power = asFloatOrNull(newValue);
+			this._setActivePowerL2(power != null ? Math.round(power) : null);
+		});
+		this.getChannel(EvseAlfen.ChannelId.CHARGE_POWER_L3).onUpdate(newValue -> {
+			var power = asFloatOrNull(newValue);
+			this._setActivePowerL3(power != null ? Math.round(power) : null);
+		});
+	}
+
+	private Channel<?> getChannel(EvseAlfen.ChannelId channelId) {
```
sfeilmeier: Why? There is already `this.channel()`

io.openems.edge.evse.chargepoint.alfen/src/io/openems/edge/evse/chargepoint/alfen/EvseAlfenImpl.java
```diff
@@ -0,0 +1,529 @@
...
+						m(EvseAlfen.ChannelId.POWER_FACTOR_L3, new FloatDoublewordElement(332)), //
+						m(EvseAlfen.ChannelId.POWER_FACTOR_SUM, new FloatDoublewordElement(334)), //
+						m(new FloatDoublewordElement(336)) //
+								.build().onUpdateCallback(value -> {
+									// Convert Hz (float) to mHz (integer)
+									this._setFrequency(value != null ? Math.round(value * 1000) : null);
+								}), //
+						m(EvseAlfen.ChannelId.CHARGE_POWER_L1, new FloatDoublewordElement(338)), //
+						m(EvseAlfen.ChannelId.CHARGE_POWER_L2, new FloatDoublewordElement(340)), //
+						m(EvseAlfen.ChannelId.CHARGE_POWER_L3, new FloatDoublewordElement(342)), //
```
sfeilmeier: Why do you not directly map to `ElectricityMeter` Channels?

io.openems.edge.evse.chargepoint.alfen/src/io/openems/edge/evse/chargepoint/alfen/EvseAlfenImpl.java
```diff
@@ -0,0 +1,521 @@
...
+		final Channel<Integer> volL1 = this.channel(phaseRotation.channelVoltageL1());
+		final Channel<Integer> volL2 = this.channel(phaseRotation.channelVoltageL2());
+		final Channel<Integer> volL3 = this.channel(phaseRotation.channelVoltageL3());
+
+		this.getChannel(EvseAlfen.ChannelId.CURRENT_L1_RAW).onUpdate(newValue -> {
+			int current = 0;
+			var rawCurrent = asFloatOrNull(newValue);
+			if (rawCurrent != null) {
+				current = TypeUtils.getAsType(OpenemsType.INTEGER, rawCurrent);
+				if (current > DETECT_PHASE_ACTIVITY) {
```
sfeilmeier: Did you experience this in reality? If yes, you could solve it with a custom `ElementToChannelConverter` (See FeneconMiniEssImpl). Same logic should not appear three times here

io.openems.edge.evse.chargepoint.alfen/src/io/openems/edge/evse/chargepoint/alfen/EvseAlfenImpl.java
```diff
@@ -0,0 +1,521 @@
...
+			} else {
+				this.phasePattern &= ~0x04;
+			}
+			curL3.setNextValue(current);
+		});
+
+		this.getChannel(EvseAlfen.ChannelId.VOLTAGE_L1_RAW).onUpdate(newValue -> {
+			var voltage = asFloatOrNull(newValue);
+			volL1.setNextValue(voltage != null //
+					? Math.round(voltage * 1000) // Convert V to mV
```
sfeilmeier: `SCALE_FACTOR_3`

io.openems.edge.evse.chargepoint.alfen/src/io/openems/edge/evse/chargepoint/alfen/EvseAlfenImpl.java
```diff
@@ -0,0 +1,521 @@
...
+			var voltage = asFloatOrNull(newValue);
+			volL3.setNextValue(voltage != null //
+					? Math.round(voltage * 1000) // Convert V to mV
+					: null);
+		});
+
+		// Map charge power to active power
+		this.getChannel(EvseAlfen.ChannelId.CHARGE_POWER).onUpdate(newValue -> {
+			var power = asFloatOrNull(newValue);
+			this._setActivePower(power != null ? Math.round(power) : null);
```
sfeilmeier: Simply map directly

io.openems.edge.evse.chargepoint.alfen/src/io/openems/edge/evse/chargepoint/alfen/EvseAlfenImpl.java
```diff
@@ -0,0 +1,521 @@
...
+		});
+
+		// Map energy delivered to production + consumption energy
+		// "Energy Delivered" in Alfen context = energy delivered to EV = consumption
+		// from grid perspective
+		this.getChannel(EvseAlfen.ChannelId.ENERGY_DELIVERED_SUM).onUpdate(newValue -> {
+			var rawEnergy = asFloatOrNull(newValue);
+			var energy = rawEnergy != null ? (Long) (long) Math.round(rawEnergy) : null;
+			this._setActiveProductionEnergy(energy);
+			this._setActiveConsumptionEnergy(energy);
```
sfeilmeier: Wrong. Don't just write the same value in both Channels.

io.openems.edge.evse.chargepoint.alfen/src/io/openems/edge/evse/chargepoint/alfen/EvseAlfenImpl.java
```diff
@@ -0,0 +1,521 @@
...
+			// 0xFFFF (see Alfen "Modbus Slave TCP/IP" 1.2), so anything else
+			// falls back to the current activity heuristic below.
+			if (configuredPhases == 1) {
+				return SingleOrThreePhase.SINGLE_PHASE;
+			}
+			if (configuredPhases == 3) {
+				return THREE_PHASE;
+			}
+		}
+		var bitCount = Integer.bitCount(this.phasePattern);
```
sfeilmeier: I don't think we should use bitwise masking here in Java if there is no good reason to do so.

io.openems.edge.evse.chargepoint.alfen/src/io/openems/edge/evse/chargepoint/alfen/Config.java
```diff
@@ -0,0 +1,52 @@
...
+	@AttributeDefinition(name = "Phase Rotation", description = "Apply standard or rotated wiring")
+	PhaseRotation phaseRotation() default PhaseRotation.L1_L2_L3;
+
+	// Physical wiring of the charge point. SINGLE_PHASE wiring can never do phase
+	// switching; it is also used as fallback when Modbus register 1215 ("Charge
+	// using 1 or 3 phases") is not available.
+	@AttributeDefinition(name = "Wiring", description = "Is the charge point wired with one or three phases?")
+	SingleOrThreePhase wiring() default SingleOrThreePhase.THREE_PHASE;
+
+	@AttributeDefinition(name = "Minimum current", description = "Minimum charging current in mA.", required = true)
```
sfeilmeier: Why do you have those currents configurable here? In EVSE at least minimum should not be required anymore.

io.openems.edge.evse.chargepoint.alfen/src/io/openems/edge/evse/chargepoint/alfen/EvseAlfenImpl.java
```diff
@@ -0,0 +1,521 @@
...
+			return ChargePointAbilities.create().build();
+		}
+
+		final var phases = this.getPhases();
+		final int minCurrent = this.config.minCurrent();
+		final int maxCurrent = this.config.maxCurrent();
+
+		final StringReadChannel mode3StateChannel = this.channel(EvseAlfen.ChannelId.MODE_3_STATE);
+		final var mode3State = mode3StateChannel.value().orElse("A");
+		final var isEvConnected = mode3State.length() == 2; // e.g., "B1", "C2"
```
sfeilmeier: No.

io.openems.edge.evse.chargepoint.alfen/src/io/openems/edge/evse/chargepoint/alfen/EvseAlfenImpl.java
```diff
@@ -0,0 +1,521 @@
...
+		// Mode 3 states:
+		// A = Not connected
+		// B = Connected, not ready
+		// C1 = Connected, ready, not charging
+		// C2 = Charging
+		// D1, D2 = With ventilation
+		// E = Error (short circuit)
+		// F = Error
+
+		boolean isReady = mode3State.startsWith("B") || mode3State.startsWith("C") || mode3State.startsWith("D");
```
sfeilmeier: Ideally `MODE_3_STATE` should be an Enum directly. Even if it is not, a `switch-case` would be better here.

sfeilmeier (changes requested):
Sorry, code quality is not yet sufficient to merge.

io.openems.edge.evse.chargepoint.alfen/test/io/openems/edge/evse/chargepoint/alfen/EvseAlfenImplTest.java
Sn0w3y: Tests are 100% written in AI as i never understood them correctly 😃

io.openems.edge.evse.chargepoint.alfen/src/io/openems/edge/evse/chargepoint/alfen/EvseAlfenImpl.java
Sn0w3y: Well actually no - but the Documentation says so - so we should be prepared i guess?

sfeilmeier: I did not read the docs, but if it explicitely mentions this behaviour, then yes 👍

io.openems.edge.evse.chargepoint.alfen/test/io/openems/edge/evse/chargepoint/alfen/EvseAlfenImplTest.java
sfeilmeier: You could have mentioned it at least. Reading and fixing AI generated code eats a lot of my time recently.

Sn0w3y: Sorry, will do in the Future and completely understand this Issue aswell - wilco

io.openems.edge.evse.chargepoint.alfen/src/io/openems/edge/evse/chargepoint/alfen/EvseAlfenImpl.java
Sn0w3y: https://eu-assets.contentstack.com/v3/assets/blt08d332658a89f766/blt64e1bba919b76c5e/IP%20for%20Alfen%20NG9xx%20platform

image

sfeilmeier: Ok, this PDF is different from the one attached in `doc`. How is that comment related to your `DETECT_PHASE_ACTIVITY` logic? It just says, that newer Alfen chargers will show NAN here - nothing about a minimum threshold.

Sn0w3y: The NaN note only backs the null-handling, not the threshold that was my mistake in wording.

regarding the PDF: the comment was written against a different Alfen document than the one committed in doc/, which is why the reference didn't match. the behaviour is in the attached one too - "Modbus Slave TCP/IP" 2.3, section 1.2: "Whenever a register is reserved or not available, the register reply is filled with Not a Number (NaN), which is set to 0xFFFF for a 16 bit register." That also covers register 1215 returning 0xFFFF, which I rely on for the phase mode. Shoukd I re-point the comment there?

Should we handle the case otherwise? Phase Switching Logic should be handled carfully as i learned the hard way at Home with KEBA haha

io.openems.edge.evse.chargepoint.alfen/src/io/openems/edge/evse/chargepoint/alfen/EvseAlfenImpl.java
```diff
@@ -0,0 +1,383 @@
...
+@GenerateTargetsFromReferences("Modbus")
+public class EvseAlfenImpl extends AbstractOpenemsModbusComponent implements EvseAlfen, EvseChargePoint,
+		ElectricityMeter, OpenemsComponent, TimedataProvider, EventHandler, ModbusComponent {
+
+	public static final int MAX_CURRENT = 32_000;
+
+	/**
+	 * Chargers with a Reallin power meter (produced after 2021) report NaN for
+	 * unavailable registers (see Alfen "Modbus for ACE" Appendix E). Without this
+	 * converter the NaN would silently end up as '0' in the Channel.
```
sfeilmeier, line 83: Did you track this down? `NaN` should never get converted to `0`. That should be handled e.g. here: https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.common/src/io/openems/edge/common/type/TypeUtils.java#L61

io.openems.edge.evse.chargepoint.alfen/src/io/openems/edge/evse/chargepoint/alfen/EvseAlfenImpl.java
```diff
@@ -0,0 +1,383 @@
...
+			value -> value);
+
+	/**
+	 * Ignores NaN like {@link #IGNORE_NAN}, then converts [A] to [mA], [V] to [mV]
+	 * or [Hz] to [mHz].
+	 */
+	private static final ElementToChannelConverter IGNORE_NAN_AND_SCALE_FACTOR_3 = chain(IGNORE_NAN, SCALE_FACTOR_3);
+
+	/** Parses the raw String of register 1201 to a {@link Mode3State}. */
+	private static final ElementToChannelConverter TO_MODE_3_STATE = new ElementToChannelConverter(
```
sfeilmeier, line 100: This is only used once, so declare `ElementToChannelConverter` directly inline below

sfeilmeier (changes requested):

sfeilmeier (approved):
Implementation is good enough now - no blockers. There is still room for improvement (see review comments) and our team at FENECON will also test it and work on it in the next weeks. Thank you! 🚀

PR #3919, opened 2026-08-26
Feature/controller evcs analog heatingelement reference target

PunugupatiSaikumar:
### Summary
Migrates three controllers from `OpenemsComponent.updateReferenceFilter()` to the Reference Target Plugin (`@GenerateTargetsFromReferences`).
This continues the same migration as the ESS controller PRs. The plugin resolves `${config.*_id}` into the OSGi reference target **before** `@Activate`, so the runtime filter rewrite is no longer needed.
#### 1. `Controller.Evcs`
- `@GenerateTargetsFromReferences("evcs")`
- Reference: `STATIC` / `GREEDY` / `MANDATORY` with
  `target = "(&(id=${config.evcs_id})(enabled=true))"`
- Removed `evcs_target()` from `Config` / `MyConfig`
- **Kept `ConfigurationAdmin` (`cm`)** — still used to rewrite force-/default-charge power when hardware limits are exceeded
- Tests: JUnit 4 → Jupiter; `DummyConfigurationAdmin` kept because `cm` is still used
#### 2. `Controller.IO.Analog`
- `@GenerateTargetsFromReferences("analogOutput")`
- Reference: `STATIC` / `GREEDY` / `MANDATORY` with
  `target = "(&(id=${config.analogOutput_id})(enabled=true))"`
- Removed `analogOutput_target()` from `Config` / `MyConfig`
- **Removed `cm`** — it was only used for `updateReferenceFilter`
- Optional `timedata` reference unchanged
- Tests: JUnit 4 → Jupiter; `DummyConfigurationAdmin` removed
#### 3. `Controller.IO.HeatingElement`
- `@GenerateTargetsFromReferences("meter")`
- Optional meter (same pattern as `Controller.IO.HeatPump.SgReady`):
  `cardinality = OPTIONAL`, `policyOption = GREEDY`,
  `target = "(&(id=${config.meter_id})(enabled=true))"`
  (not `STATIC`, because `meter_id` defaults to `""`)
- Removed the old empty-id `(false=true)` / `updateReferenceFilterRaw` handling
- Removed `meter_target()` from `Config` / `MyConfig`
- **Removed `cm`** — it was only used for the meter filter
- Tests: removed unused `DummyConfigurationAdmin` wiring only (left on JUnit 4 to keep the diff small)


sfeilmeier (approved):
Thank you for proceeding. I am awaiting the moment when we can remove `updateReferenceFilter()` 🚀 (not sure if this will actually be possible, though)

PR #3931, opened 2026-08-28
[Edge] Controller.IO.Heating.Room: migrate reference target

PunugupatiSaikumar:
## Summary
Migrates `Controller.IO.Heating.Room` from `OpenemsComponent.updateReferenceFilter()` to the Reference Target Plugin (`@GenerateTargetsFromReferences`), following the pattern used in e.g. `Controller.Api.Modbus` for multi-component references.
- Add `@GenerateTargetsFromReferences` for all four references: `floorThermometer`, `ambientThermometer`, `floorRelayComponents`, `infraredRelayComponents`
- Add static `target` filters on the `@Reference` annotations
- Remove all four `updateReferenceFilter()` calls from `applyConfig()`
- Remove obsolete `*_target()` properties from `Config` and `MyConfig`
- Add `floorRelayComponent_ids` / `infraredRelayComponent_ids` config fields for relay `List<DigitalOutput>` references (the plugin needs component IDs; `floorRelays` / `infraredRelays` remain channel addresses for runtime relay switching)
- Keep `ConfigurationAdmin` — still required for JSCalendar schedule updates via JSON API


io.openems.edge.controller.io.heating.room/src/io/openems/edge/controller/io/heating/room/Config.java
```diff
@@ -44,9 +44,15 @@
 	@AttributeDefinition(name = "Floor Heating Relay(s)", description = "Channel addresses of the Relays for Floor Heating")
 	String[] floorRelays();
 
+	@AttributeDefinition(name = "Floor Relay Component IDs", description = "Component-IDs of the IO components used for Floor Heating Relays")
+	String[] floorRelayComponent_ids() default {};
```
sfeilmeier, line 48: This would be a breaking change, because it would require reconfiguration of productive systems.
Also the information about Component-ID would be redundant.

We either need to find another way to be able to use `GenerateTargetsFromReferences` (I don't know if we can parse ChannelAddresses there) or revert back to old approach using `updateReferenceFilter`.

@michaelgrill: Opinions? I believe we have similar topics - i.e. referencing to a ChannelAddress of a Relay - in more places, so a generic solution might be good.

sfeilmeier (changes requested):

io.openems.edge.controller.io.heating.room/src/io/openems/edge/controller/io/heating/room/Config.java
PunugupatiSaikumar: Thanks for catching this, you're right that the plugin can't derive component IDs from ChannelAddresses, which is why I added the extra config fields. I'm happy to either (a) revert the two relay-list references back to updateReferenceFilter and keep the plugin only for the two thermometer references (no config change, non-breaking), or (b) wait for a generic ChannelAddress solution in the Reference Target Plugin. Let me know which you prefer.

michaelgrill: The string can be modified to whatever we want during the configuration change of the plugin, so this can be added. I just dont want the syntax of the string to get too ugly (especially for eclipse with no highlighting (yet)), so the syntax needs to be understandable for everyone.

Depending on how much other tasks you have @PunugupatiSaikumar you can have a look at it if you want, otherwise i will probably look into it next slack day. Feel free to ask questions about it anytime.

PunugupatiSaikumar: Thanks @michaelgrill,  I'd like to take this on.

Looking at the plugin, I think it can be done without touching the parser.
`StringWithParams.getParametersFrom` already treats the part before the dot as a topic,
and `modifyConfiguration` resolves topics through a map that currently only registers
`config`. The `ValueProvider` interface is already there for exactly this. So a second
topic would be enough:

```java
@Reference(policyOption = GREEDY, //
        target = "(&(id=${componentId.floorRelays})(enabled=true))")
private List<DigitalOutput> floorRelayComponents;

Comment (michaelgrill) [reply io.openems.edge.controller.io.heating.room/src/io/openems/edge/controller/io/heating/room/Config.java:48]: I wouldnt add another topic. The topic was intended to be the source of the data/variable another one could be (not implemented yet) `env` for environment variables.

I would add something like a parse/value transformer which then can be used for every topic, and maybe also chained like:

```java
@Reference(target = "(id=${componentIdFromChannelAddress(config.floorRelays)})")
private List<DigitalOutput> floorRelayComponents;
```
(name is a little long)
or 

```java
@Reference(target = "(id=${componentId(channelAddress(config.floorRelays))})")
private List<DigitalOutput> floorRelayComponents;
```
or
```java
@Reference(target = "(id=${config.floorRelays.toChannelAddress().componentId())})")
private List<DigitalOutput> floorRelayComponents;
```

maybe even something next to the variable
```java
@Reference(target = "(id=${config.floorRelays;componentIdFromChannel)})")
private List<DigitalOutput> floorRelayComponents;
```

but i m not fully satisfied with the string yet. It could get quite long and confusing. And also the string parsing stuff could get complex.

But in general i believe it would be more extensable with something separate to the topic.

PunugupatiSaikumar: Thanks @michaelgrill, I understand the distinction now: the topic identifies the data source, while extracting the component ID should be a separate transformation.
I’d like to try your semicolon-style suggestion: ${config.floorRelays;componentIdFromChannel}. This would keep existing expressions working and avoid additional configuration fields.
Would this syntax be acceptable for an initial implementation? I’ll include tests for existing expressions, relay arrays, and invalid inputs.

michaelgrill: Sounds good, if you need any help feel free to reach out to me

PunugupatiSaikumar: @michaelgrill Sure

io.openems.core.referencetarget/src/io/openems/core/referencetarget/ComponentIdFromChannelTransformer.java
```diff
@@ -0,0 +1,51 @@
...
+
+		if (value instanceof String[] channelAddresses) {
+			return Arrays.stream(channelAddresses) //
+					.filter(address -> !"".equals(address)) //
+					.map(ComponentIdFromChannelTransformer::toComponentId) //
+					.distinct() //
+					.toArray(String[]::new);
+		}
+
+		if (value instanceof Collection<?> channelAddresses) {
```
michaelgrill, line 26: have you encountered that a value was a collection?

In a java annotation there can only be primitive types like

```
@interface Config {

	// possible
	String[] ids();

	// not possible
	List<String> ids();

}
```

so i am wondering if this type is even possible here

io.openems.core.referencetarget/src/io/openems/core/referencetarget/ValueTransformers.java
```diff
@@ -0,0 +1,28 @@
+package io.openems.core.referencetarget;
+
+import java.util.Map;
+
+public final class ValueTransformers {
+
+	private static final Map<String, ValueTransformer> BY_NAME = Map.of(//
+			"componentIdFromChannel", new ComponentIdFromChannelTransformer() //
```
michaelgrill, line 8: this solution is totally fine, just to let you know how else it could have been made with osgi and dependency injection:

You can annotate the implementations with `@Component` then they are available to inject with 
`@Reference List<ValueTransformer> transformers` or if you want a custom collection you can use methods like bind, update and unbind e. g. https://github.com/OpenEMS/openems/blob/189ac916fd653d3496797ac20921d18a2e6237f1/io.openems.edge.core/src/io/openems/edge/core/componentmanager/tracker/ComponentTracker.java#L42-L71

maybe the most similar implementation currently would be `ComponentJsonApi`, it has a method for id which in your case could be the name and every implementation gets binded in the [ComponentRequestHandler](https://github.com/OpenEMS/openems/blob/develop/io.openems.edge.controller.api.common/src/io/openems/edge/controller/api/common/handler/ComponentRequestHandler.java) to call methods on that instance mapped to their ids.

PR #3933, opened 2026-08-28
Fix typo in nginx.adoc

Hans-4:
This PR fixes the typo `OpemEMS UI` in the nginx documentation page.

I also thought about if we could update this part to include the following command for users less familiar with terminal commands, but i wanted to hear your opinion about this first. 

```
. Start nginx:
+
[source,bash]
----
sudo systemctl start nginx
----
+
See `man nginx` and the documentation on https://nginx.org for more information.
```

### DIFF
--- a/doc/modules/ROOT/pages/ui/deploy/nginx.adoc
+++ b/doc/modules/ROOT/pages/ui/deploy/nginx.adoc
@@ -15,7 +15,7 @@ NOTE: These instructions are intended for a quick setup, e.g. to get started wit
 
 === Setup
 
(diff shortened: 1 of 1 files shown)


sfeilmeier:
> I also thought about if we could update this part to include the following command for users less familiar with terminal commands, but i wanted to hear your opinion about this first.

Yes, absolutely. But in that situation NGINX should be already started, as it starts automatically after install. Maybe `systemctl restart nginx` would be better - or `nginx -s reload`.

sfeilmeier (approved):

PR #3934, opened 2026-08-29
[Common] Add SECURITY.md file

sfeilmeier:
Adds SECURITY.md with a formal OpenEMS vulnerability reporting and CVE disclosure process.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,162 @@
+# Security Policy
+
+## Supported Versions
+
+OpenEMS accepts security reports for:
+
+- The latest released version
+- The develop branch
+
+Only these two targets are supported for security handling. Older versions are not covered.
+
+## Reporting a Vulnerability
+
+Please report vulnerabilities through one of the following channels:
+
+- Email: security@openems.io
+- OpenEMS Community member-only security category (only for members of OpenEMS Association e.V.): https://community.openems.io/c/openems-association/security-advisories-members-only/23
+
+For members of OpenEMS Association e.V., reporting only in the member-only forum category is sufficient; no additional email is required.
+
+Please do **not** create public GitHub issues for suspected vulnerabilities.
+
+When reporting, include as much detail as possible:
+
+- Description of the vulnerability
+- Affected OpenEMS versions
+- Steps to reproduce (if possible)
+- CVSS v3.1 score or later (mandatory; reporter suggests, maintainer decides final score)
+- Workarounds (if any)
+- Contact information (optional; anonymous reports are possible)
+
+## Roles and Responsibilities
+
+| Role | Responsibility | Contact |
+| --- | --- | --- |
+| Maintainer | Analysis, patch creation, severity decision, internal and public communication | security@openems.io |
+| OpenEMS Association Office | Coordination support | office@openems.io |
+| Members of OpenEMS Association e.V. | Access to internal CVE information in the forum and optional member-only reporting | N/A |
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,162 @@
...
+- The latest released version
+- The develop branch
+
+Only these two targets are supported for security handling. Older versions are not covered.
+
+## Reporting a Vulnerability
+
+Please report vulnerabilities through one of the following channels:
+
+- Email: security@openems.io
```
parapluplu, line 16: out of curiosity - who receives these mails? Is it a list of contacts?

SECURITY.md
```diff
@@ -0,0 +1,162 @@
...
+
+### 2. Acknowledgement
+
+- Receipt is acknowledged manually.
+- Target timeline: within 72 hours.
+
+### 3. Analysis and Patch
+
+- Initial assessment target: within 72 hours.
+- Patch target branch: develop.
```
parapluplu: To ensure, that nobody creates a merge request based on develop for the patch and publishes it,let's add below:

- The fix is developed **privately** (e.g. in the temporary private fork of a GitHub Security Advisory draft), **not** on public `develop`, so the patch does not reveal the vulnerability before coordinated disclosure.
- The patch is merged to `develop` **together with** the public advisory (see step 5), not before.

SECURITY.md
```diff
@@ -0,0 +1,162 @@
...
+
+### 4. Internal Disclosure
+
+After initial triage, CVE handling and coordinated disclosure are discussed internally in the OpenEMS Community member-only category:
+
+- https://community.openems.io/c/openems-association/security-advisories-members-only/23
+
+- Timing: immediately after patch integration to develop.
```
parapluplu: Internal disclosure should happen during the embargo (before the fix is public), otherwise "internal" disclosure lands after the fix is already visible on `develop`.

I would rephrase this to:
Timing: as soon as the report is triaged and confirmed - during the embargo, before the fix is public.

SECURITY.md
```diff
@@ -0,0 +1,162 @@
...
+- Content: status, affected versions, mitigation, patch/release plan.
+
+### 5. Public Disclosure
+
+- Preferred channel: GitHub Security Advisories
+- Link: https://github.com/OpenEMS/openems/security/advisories
+- Public disclosure happens after a fix is available (fix-first rule).
```
parapluplu: let's add a sentence here, that clarifies what means "available":

"Available" means the fix is **ready**, not that it is already public on `develop`.

SECURITY.md
```diff
@@ -0,0 +1,162 @@
...
+## Scope
+
+This policy applies to vulnerabilities in OpenEMS source code and official distributions in this repository.
+
+Out of scope:
+
+- User-specific deployment hardening and operations (users are responsible for their deployments)
+- Local environment hardening and network perimeter security
+- Vulnerabilities only in third-party dependencies not maintained by OpenEMS
+- Forks and custom downstream variants
```
parapluplu: Let's make clear, that it's only out of scope, if these issue **only** exists in forks or downstream:

Vulnerabilities that exist **only** in forks or downstream variants and are **not reproducible in this repository**

parapluplu (commented):
Nice to see that introduced :)

SECURITY.md
sfeilmeier: It's configured to forward to my personal email and the OpenEMS Association office (office@openems.io)

parapluplu (approved):

PR #2528, opened 2024-02-07
Sungrow String and ESS Implementation

Sn0w3y:
Based on the Implementation of @clehne in opernikus-common fork.
I adapted it to use ElectricityMeter

Thanks to them

PR #3935, opened 2026-08-31
Sungrow PV Inverter and ESS Implementation

tsicking:
Updated version of Sungrow implementation. Replaces  #2528
This version is used on a handful of productive systems run by opernikus.


io.openems.edge.sungrow/src/io/openems/edge/sungrow/dccharger/Config.java
```diff
@@ -0,0 +1,28 @@
...
+	String alias() default "";
+
+	@AttributeDefinition(name = "Is enabled?", description = "Is this Component enabled?")
+	boolean enabled() default true;
+
+	@AttributeDefinition(name = "ESS-ID", description = "ID of the Sungrow Ess Component.")
+	String ess_id() default "ess0";
+	
+	@AttributeDefinition(name = "ESS target filter", description = "This is auto-generated by 'ESS-ID'.")
+	String ess_target() default "(enabled=true)";
```
sfeilmeier: Please replace with ReferenceConfigurationPlugin

io.openems.edge.sungrow/src/io/openems/edge/sungrow/dccharger/SungrowDcChargerImpl.java
```diff
@@ -0,0 +1,119 @@
...
+		this.mapChannelValues();
+	}
+
+	@Deactivate
+	protected void deactivate() {
+		super.deactivate();
+	}
+
+	private void mapChannelValues() {
+		this.ess.getTotalDcPowerChannel().onSetNextValue(value -> {
```
sfeilmeier, line 82: Why do you not map those Registers directly?

io.openems.edge.sungrow/src/io/openems/edge/sungrow/ess/EssSungrow.java
```diff
@@ -0,0 +1,493 @@
...
+import io.openems.edge.sungrow.ess.enums.ChargeDischargeCommand;
+import io.openems.edge.sungrow.ess.enums.EmsMode;
+import io.openems.edge.sungrow.ess.enums.EnableDisable;
+import io.openems.edge.sungrow.ess.enums.SystemState;
+
+public interface EssSungrow extends OpenemsComponent {
+
+	public enum ChannelId implements io.openems.edge.common.channel.ChannelId {
+		SERIAL_NUMBER(Doc.of(OpenemsType.STRING) //
+				.accessMode(AccessMode.READ_ONLY) //
```
sfeilmeier: Please cleanup this file. READ_ONLY and Priority.LOW are defaults

io.openems.edge.ess.api/src/io/openems/edge/ess/api/SymmetricEss.java
```diff
@@ -320,6 +320,15 @@ public default Channel<GridMode> getGridModeChannel() {
 	public default GridMode getGridMode() {
 		return this.getGridModeChannel().value().asEnum();
 	}
+	
+	/**
+	 * Internal method to set the 'nextValue' on {@link ChannelId#GRID_MODE} Channel.
+	 *
+	 * @param value the next value
+	 */
+	public default void _setGridMode(GridMode value) {
```
sfeilmeier, line 329: No, it was replaced with `ChannelUtils.setValue()` by purpose. Please reset to develop.

sfeilmeier (changes requested):
See my remarks. Same applies to other similar files.

Who can work on these topics?

io.openems.edge.ess.api/src/io/openems/edge/ess/api/SymmetricEss.java
tsicking: Okay, I see. I was wondering why this method disappeared. For what reason do you use `ChannelUtils.setValue()`?
The same then for `ManagedSymmetricEss._setAllowedChargePower()` I guess? But why is `ManagedSymmetricEss._setAllowedDischargePower()` still there?

io.openems.edge.sungrow/src/io/openems/edge/sungrow/dccharger/SungrowDcChargerImpl.java
tsicking: So the alternative would be to make this a `ModbusComponent`, read the registers here and leave them out in the ESS implementation? This would mean that you send more modbus tasks per second but you can remove some channels from the ESS, right?
I have no strong opinion which is better, but this was the first thing that came into my mind. I just had a look at the GoodWe implementations and saw that the single string chargers have their own modbus protocol, while the two string chargers don't. 
Now, as I write, one more options comes into my mind. If the charger is referenced in the ESS and not the other way round, one could do something like the following in the modbus protocol:

```java
m(new UnsignedDoubleWordElement(5016)).build() //
								.onUpdateCallback(value -> this.charger._setActualPower(value)), //
```
Is that what you have in mind?

io.openems.edge.sungrow/src/io/openems/edge/sungrow/dccharger/SungrowDcChargerImpl.java
```diff
@@ -0,0 +1,121 @@
...
+@GenerateTargetsFromReferences("ess")
+public class SungrowDcChargerImpl extends AbstractOpenemsComponent
+		implements SungrowDcCharger, EssDcCharger, TimedataProvider, OpenemsComponent, EventHandler {
+
+	protected Config config = null;
+
+	@Reference
+	private ConfigurationAdmin cm;
+
+	@Reference(policy = ReferencePolicy.STATIC, policyOption = ReferencePolicyOption.GREEDY, cardinality = ReferenceCardinality.MANDATORY)
```
sfeilmeier: If you use `GenerateTargetsFromReferences`, you have to set the target property here - see documentation: https://github.com/OpenEMS/openems/blob/develop/io.openems.core.referencetarget/readme.adoc

io.openems.edge.sungrow/src/io/openems/edge/sungrow/dccharger/SungrowDcChargerImpl.java
```diff
@@ -0,0 +1,121 @@
...
+				EssDcCharger.ChannelId.values() //
+		);
+	}
+
+	@Activate
+	private void activate(ComponentContext context, Config config) throws OpenemsException {
+		this.config = config;
+		super.activate(context, config.id(), config.alias(), config.enabled());
+
+		if (OpenemsComponent.updateReferenceFilter(this.cm, this.servicePid(), "ess", config.ess_id())) {
```
sfeilmeier: Also this is obsolete then

io.openems.edge.sungrow/src/io/openems/edge/sungrow/dccharger/SungrowDcChargerImpl.java
sfeilmeier: Ok, I did not realize this is actually copying from `ess` component. Then it's ok. Only flaw is, that you would have to remove the listener on `deactivate`, otherwise the garbage collector will not be able to cleanup.

io.openems.edge.sungrow/src/io/openems/edge/sungrow/ess/Config.java
```diff
@@ -0,0 +1,33 @@
+package io.openems.edge.sungrow.ess;
+
+import org.osgi.service.metatype.annotations.AttributeDefinition;
+import org.osgi.service.metatype.annotations.ObjectClassDefinition;
+
+import io.openems.edge.sungrow.ess.enums.ControlMode;
+
+@ObjectClassDefinition(//
+		name = "ESS Sungrow", //
+		description = "Implements the Sungrow SH10RT Hybrid Ess")
```
sfeilmeier, line 10: Do you know how different the protocols of Sungrow devices are? Should we probably rename the Factory-ID to something like `Ess.Sungrow.Hybrid.*` to have a namespace for future implementations?

io.openems.edge.sungrow/src/io/openems/edge/sungrow/pvinverter/PvInverterSungrowImpl.java
```diff
@@ -0,0 +1,219 @@
...
+	private boolean readOnly = true; // oEMS add
+
+	@Activate
+	void activate(ComponentContext context, Config config) throws OpenemsException {
+		this.readOnly = config.readOnly(); // oEMS add
+		if (super.activate(context, config.id(), config.alias(), config.enabled(), config.modbusUnitId(), this.cm,
+				"Modbus", config.modbus_id())) {
+			return;
+		}
+		this.getActivePowerChannel().onSetNextValue(value -> { //
```
sfeilmeier: `ElectricityMeter .calculatePhasesFromActivePower()` does exactly this

io.openems.edge.sungrow/test/io/openems/edge/sungrow/dccharger/SungrowDcChargerTest.java
```diff
@@ -0,0 +1,25 @@
+package io.openems.edge.sungrow.dccharger;
+
+import org.junit.Test;
```
sfeilmeier: JUnit4 is deprecated. Please use Juniper

io.openems.edge.sungrow/test/io/openems/edge/sungrow/dccharger/SungrowDcChargerTest.java
```diff
@@ -0,0 +1,25 @@
...
+
+	private static final String CHARGER_ID = "charger0";
+	private static final String ESS_ID = "ess0";
+
+	@Test
+	public void test() throws Exception {
+		new ComponentTest(new SungrowDcChargerImpl()) //
+				.addReference("cm", new DummyConfigurationAdmin()) //
+				.addReference("ess", new EssSungrowImpl()) //
+				.activate(MyConfig.create() //
```
sfeilmeier, line 15: I can only recommend having one test with `strict-mode` for Modbus devices.

sfeilmeier (approved):

PR #3938, opened 2026-09-01
[Edge]Fixed EssPower circular dependency causes "Coefficient was not found" error in multiple ESS implementations

hvgollar:
-[Edge] Ess.Cluster: use bound ESS ids in getEssIds()
The power solver builds cluster constraints from getEssIds(). Using
config ess_ids included children that were no longer registered, which
caused "Coefficient was not found" when an ESS was disabled or late.

-getEssIds() now returns the ids of currently bound child ESS
(this.esss), not the configured ess_ids list. Felix ESS-IDs are
unchanged and still used only for OSGi targeting.


pooran-c (changes requested):
Trace through the junit test.

PR #3939, opened 2026-09-02
[Edge] Meter.Mqtt: add generic config-driven MQTT meter

ramaguruprakash:
Adds io.openems.edge.meter.mqtt, an ElectricityMeter that subscribes to an MQTT topic via Bridge.Mqtt, parses a flat JSON payload, and maps configured fields to ElectricityMeter channels.

New meters/vendors are onboarded by configuration alone (a topic plus a "jsonField:CHANNEL:scale" mapping list) without code changes. A negative scale inverts a value, and the meter type sets PRODUCTION vs CONSUMPTION. The reference to the MQTT bridge is selected via @GenerateTargetsFromReferences.

Registered in EdgeApp.bndrun.

### DIFF
--- a/io.openems.edge.application/EdgeApp.bndrun
+++ b/io.openems.edge.application/EdgeApp.bndrun
@@ -182,6 +182,7 @@
 	bnd.identity;id='io.openems.edge.meter.hager',\
 	bnd.identity;id='io.openems.edge.meter.janitza',\
 	bnd.identity;id='io.openems.edge.meter.kdk',\
+	bnd.identity;id='io.openems.edge.meter.mqtt',\
 	bnd.identity;id='io.openems.edge.meter.opendtu',\
 	bnd.identity;id='io.openems.edge.meter.phoenixcontact',\
 	bnd.identity;id='io.openems.edge.meter.plexlog',\
@@ -397,6 +398,7 @@
 	io.openems.edge.meter.hager;version=snapshot,\
 	io.openems.edge.meter.janitza;version=snapshot,\
 	io.openems.edge.meter.kdk;version=snapshot,\
+	io.openems.edge.meter.mqtt;version=snapshot,\
 	io.openems.edge.meter.opendtu;version=snapshot,\
 	io.openems.edge.meter.phoenixcontact;version=snapshot,\
 	io.openems.edge.meter.plexlog;version=snapshot,\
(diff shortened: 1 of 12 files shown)


sfeilmeier:
I am generally not a fan of such generic implementations - because from my experience they never work reliably. We had the same discussions often for generic Modbus implementation. I would much more prefer to have discrete device implementations. Which device do you test with?

ramaguruprakash:
Thanks, that's fair, To your question: I'm testing against the MeshEMS OpenAMI street-pole EMS board, which publishes SunSpec model 11 (single-phase) / 213 (three-phase) as JSON over MQTT. 

On "generic": to be clear, this isn't trying to be a universal Modbus-style catch-all; it targets one payload shape (SunSpec-as-JSON over MQTT) and only ever writes native ElectricityMeter channels. The field→channel mapping is config-driven so additional OpenAMI/SunSpec-over-MQTT meters don't each need a new bundle.

That said, our actual requirement is this one device. So I'm happy to make it a discrete io.openems.edge.meter.meshems implementation with the SunSpec-11/213 mapping baked in and unit-tested against the device's real payload, reusing the same MQTT/JSON plumbing. Would that fit OpenEMS better? Happy to go that way.

sfeilmeier:
> Thanks, that's fair, To your question: I'm testing against the MeshEMS OpenAMI street-pole EMS board, which publishes SunSpec model 11 (single-phase) / 213 (three-phase) as JSON over MQTT.

Ok, thank you. I cannot find this product. Can you please add a link to the product page and protocol description in the readme?

> On "generic": to be clear, this isn't trying to be a universal Modbus-style catch-all; it targets one payload shape (SunSpec-as-JSON over MQTT) and only ever writes native ElectricityMeter channels. The field→channel mapping is config-driven so additional OpenAMI/SunSpec-over-MQTT meters don't each need a new bundle.

Also "SunSpec-as-JSON over MQTT" does not seem to be a well established standard.

> That said, our actual requirement is this one device. So I'm happy to make it a discrete io.openems.edge.meter.meshems implementation with the SunSpec-11/213 mapping baked in and unit-tested against the device's real payload, reusing the same MQTT/JSON plumbing. Would that fit OpenEMS better? Happy to go that way.

I would be in favor of that, yes. Let's not overengineer for now - and if a later product implementation can reuse parts, we can still refactor the code - e.g. extracting this logic to MQTT bridge.

PR #3945, opened 2026-09-03
feat(ui/user): expose theme switcher for OpenEMS theme

### LINKED ISSUE #3944: Add Dark Mode with automatic system theme detection
### Component

Edge

## Description

It would be great if the OpenEMS UI supported a **Dark Mode**.

Ideally, the UI should automatically follow the user's operating system
or browser theme preference using `prefers-color-scheme`, while also
allowing users to manually select their preferred theme.

### Proposed behavior

-   **System / Auto** -- automatically use Light or Dark Mode depending
    on the system setting
-   **Light** -- always use the light theme
-   **Dark** -- always use the dark theme
-   Persist a manually selected preference, e.g. in local storage

When **System / Auto** is selected, the UI should also react to theme
changes while the application is open.

### Motivation

OpenEMS dashboards are often kept open for longer periods of time. A
dark theme would make the UI more comfortable to use in low-light
environments and would integrate better with devices that are already
configured to use Dark Mode.

Supporting the system preference would also provide a modern default
behavior without requiring users to configure anything manually.

### Possible implementation

The automatic theme could be detected using the standard CSS media
query:

``` css
@media (prefers-color-scheme: dark) {
  /* dark theme */
}
```

or via JavaScript:

``` javascript
window.matchMedia('(prefers-color-scheme: dark)')
```

### Expected result

OpenEMS should provide a theme setting with:

**System (default) / Light / Dark**

with **System** automatically following the operating system/browser
theme.

Christoph-87:
Introduces `isThemeSelectionAvailable` getter in UserComponent that returns true for OpenEMS, FENECON, and FENECONBeta themes, replacing the hard-coded FENECON/FENECONBeta check in the template.

Closes #3944


ui/src/app/user/user.component.ts
```diff
@@ -123,6 +123,10 @@ export class UserComponent implements OnInit {
         return UserTheme.LIGHT;
     } // Theme as of "Light","Dark" or "System" Themes.
 
+    protected get isThemeSelectionAvailable(): boolean {
```
sfeilmeier: It's problematic to call functions from templates. We should calculate this value statically instead - or simply add `environment.theme === OpenEMS` in the template

sfeilmeier (changes requested):

ui/src/app/user/user.component.ts
Christoph-87: Done — converted `isThemeSelectionAvailable` from a getter to a `protected readonly` field. Since `environment.theme` is fixed at build time the value is now computed once at component instantiation instead of on every change-detection cycle.

sfeilmeier (approved):

PR #3946, opened 2026-09-03
Fix/ess cluster power distribution v1

### LINKED ISSUE #3752: Controller.Symmetric.Balancing accepts Ess.Cluster but fails with missing power coefficients
### Description

Setup:
- OpenEMS Edge with two real ManagedSymmetricEss components: ess0, ess1
- Ess.Cluster component: essCluster0
- Grid meter: meter0
- Controller.Symmetric.Balancing configured with ess.id=essCluster0, meter.id=meter0

Observed:
- ctrlBalancing0/State reports Ok
- Scheduler cycle logs repeated errors:
  Coefficient for [essCluster0,ALL,ACTIVE] was not found
  Coefficient for [essCluster0,ALL,REACTIVE] was not found
- No useful setpoints are applied to the real ESS components.

Expected:
Either:
1. Controller.Symmetric.Balancing should support Ess.Cluster / MetaEss targets, or
2. configuration validation should reject Ess.Cluster as ess.id, and the component state should report Warning/Fault instead of Ok.

Workaround:
Disabled Controller.Symmetric.Balancing and implemented external control that writes setpoints directly to ess0/SetActivePowerEquals and ess1/SetActivePowerEquals.

### Screenshots

_No response_

### Operating System

_No response_

1. Configure an OpenEMS Edge with at least two real ManagedSymmetricEss components, for example:
   - ess0
   - ess1

2. Configure an Ess.Cluster component that includes those ESS components, for example:
   - essCluster0 with ess0 and ess1 as members

3. Configure a grid meter, for example:
   - meter0

4. Configure Controller.Symmetric.Balancing with:
   - ess.id = essCluster0
   - meter.id = meter0
   - targetGridSetpoint = -5

5. Add the balancing controller to Scheduler.FixedOrder and start/restart OpenEMS Edge.

6. Observe the controller state and logs.

Actual result:
- ctrlBalancing0/State reports Ok
- The scheduler cycle logs repeated errors like:
  Coefficient for [essCluster0,ALL,ACTIVE] was not found
  Coefficient for [essCluster0,ALL,REACTIVE] was not found
- The controller does not apply useful setpoints to the underlying ESS components.

Expected result:
Either Controller.Symmetric.Balancing should support Ess.Cluster as ess.id, or the configuration should be rejected / reported as Warning or Fault instead of showing Ok.

Additional observation:
Direct setpoint writes to the real ESS components work, for example ess0/SetActivePowerEquals and ess1/SetActivePowerEquals. The failure appears when the balancing controller targets the Ess.Cluster / MetaEss.

Christoph-87:
## Summary

`Controller.Symmetric.Balancing` (and other controllers using the V1 power
distribution) silently fails when `Ess.Cluster` is used as `ess.id`. The
scheduler cycle logs repeated errors like:

Coefficient for [essCluster0,ALL,ACTIVE] was not found
Coefficient for [essCluster0,ALL,REACTIVE] was not found

Meanwhile `ctrlBalancing0/State` reports `Ok` — no setpoints are applied to
the underlying physical ESS components.

## Root Cause

The V1 power distribution (`Data.updateInverters()`) had two related gaps
when a `MetaEss` wrapper (e.g. `EssCluster`) was present in the ESS list:

1. **Phantom Inverter**: An `Inverter` entry was created for the `MetaEss`
   itself, even though it has no physical inverter and its `applyPower()`
   intentionally throws `"should never be called"`.

2. **Missing member coefficients**: Only the cluster's own ID was added to
   the `Coefficients` set. The member IDs (`ess0`, `ess1`, …) were never
   registered, so `ConstraintUtil.createMetaEssConstraints()` could not
   build the `cluster = ess0 + ess1` constraint — causing the
   "Coefficient not found" error at runtime.

Note: The V2 power distribution (`BALANCE` strategy) already handles
`MetaEss` correctly via its `Virtual`/`Actual` entry pattern. This fix
brings V1 to the same level.

## Fix

Two small changes in
`io.openems.edge.ess.core/…/power/v1/Data.java` (`updateInverters()`):

- **Skip `MetaEss` during `Inverter` creation** — only physical ESS
  instances get an `Inverter` entry.
- **Expand `MetaEss` member IDs into the `Coefficients` set** — member IDs
  are resolved via `MetaEss.getEssIds()` and added alongside the cluster ID.

No other changes are needed: `ConstraintUtil.createMetaEssConstraints()` and
`Solver.applySolution()` already handle `MetaEss` correctly once the
coefficients are in place.

## Tests

Two new test cases added to `DataTest`:

- `testNoInverterForMetaEss` — asserts that a `MetaEss` does not produce an
  `Inverter` entry.
- `testMemberCoefficientsRegisteredWhenOnlyClusterInEsss` — asserts that
  member ESS IDs are registered in `Coefficients` even when only the cluster
  itself appears in the ESS list (the exact failure scenario of #3752).

## Checklist

- [x] Existing tests pass (`./gradlew :io.openems.edge.ess.core:test`)
- [x] New regression tests added
- [x] No changes outside `io.openems.edge.ess.core`
- [x] V2 / `BALANCE` strategy behaviour unchanged

Closes #3752

### DIFF
--- a/io.openems.edge.ess.core/src/io/openems/edge/ess/core/power/v1/Data.java
+++ b/io.openems.edge.ess.core/src/io/openems/edge/ess/core/power/v1/Data.java
@@ -16,6 +16,7 @@ import com.google.common.collect.Streams;
 import io.openems.common.exceptions.OpenemsException;
 import io.openems.edge.common.type.Phase.SingleOrAllPhase;
 import io.openems.edge.ess.api.ManagedSymmetricEss;
+import io.openems.edge.ess.api.MetaEss;
 import io.openems.edge.ess.core.power.EssPower;
 import io.openems.edge.ess.core.power.v1.data.ConstraintUtil;
 import io.openems.edge.ess.core.power.v1.data.WeightsUtil;
@@ -77,16 +78,24 @@ public class Data {
 
 		this.inverters.clear();
 
-		// Create inverters and add them to list
+		// Create inverters and add them to list; skip MetaEss wrappers (e.g.
+		// EssCluster) as they have no physical inverter of their own
 		for (ManagedSymmetricEss ess : esss) {
+			if (ess instanceof MetaEss) {
+				continue;
+			}
 			var essType = EssType.getEssType(ess);
 			Collections.addAll(this.inverters, Inverter.of(this.symmetricMode, ess, essType));
 		}
 
-		// Re-Initialize Coefficients
+		// Re-Initialize Coefficients; also register member IDs of MetaEss wrappers so
+		// that MetaEss constraints (e.g. cluster = ess0 + ess1) can be created
 		Set<String> essIds = new HashSet<>();
 		for (ManagedSymmetricEss ess : esss) {
 			essIds.add(ess.id());
+			if (ess instanceof MetaEss me) {
+				Collections.addAll(essIds, me.getEssIds());
+			}
 		}
 		this.coefficients.initialize(this.symmetricMode, essIds);
 
--- a/io.openems.edge.ess.core/test/io/openems/edge/ess/core/power/v1/DataTest.java
+++ b/io.openems.edge.ess.core/test/io/openems/edge/ess/core/power/v1/DataTest.java
@@ -53,4 +53,39 @@ public class DataTest {
 		data.setSymmetricMode(false);
 		assertEquals(esss.size() * 4 /* phases + all */ * 2 /* pwr */, data.getCoefficients().getNoOfCoefficients());
 	}
+
+	/**
+	 * Verifies that a MetaEss (e.g. EssCluster) does not get its own Inverter entry.
+	 * Only physical ESS members should have Inverters so that the solver never tries
+	 * to call applyPower() on the wrapper.
+	 */
+	@Test
+	public void testNoInverterForMetaEss() {
+		data.setSymmetricMode(true);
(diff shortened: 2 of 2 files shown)


Christoph-87:
**Relation to issue #3500**

Issue #3500 describes the `Coefficient for [essCluster0,...] was not found` error and attributes it to two root causes:

1. A race condition during OSGi startup caused by the `@Reference(cardinality = MANDATORY)` pattern on `Power` in ESS implementations — proposed fix: change to `OPTIONAL`/`DYNAMIC`.
2. Missing synchronization in `Data.getConstraintsWithoutDisabledInverters()`.

This PR addresses a **third, related root cause** that is not covered by #3500: `Data.updateInverters()` was creating a phantom `Inverter` entry for `MetaEss` wrappers (e.g. `EssCluster`) and never registering the member IDs (`ess0`, `ess1`, …) in the coefficient set. As a result, `createMetaEssConstraints()` could not build the `cluster = ess0 + ess1` constraint, which is the direct cause of the `essCluster0` coefficient error in cluster setups.

This PR and #3500 are complementary and do not conflict — they touch different methods in `Data.java`. Both can be merged independently.

pooran-c (approved):
Nice, MetaEss is no longer an Inverter entry.
"Coefficient not found issue" will be fixed in  https://github.com/OpenEMS/openems/pull/3938

sfeilmeier (approved):

PR #3947, opened 2026-09-03
fix(PowerSolverV1): DurationCalculation

DerStoecki:
This PR fixes a wrong Duration calculation in PowerSolver V1.

If for whatever reason, the powersolver might take longer than Integer.MAX nano seconds (~2.14s) -> it will show the wrong time in the duration channel (possibly even negative values simply bc the int casting takes place before applying the milliseconds conversion)

This PR offers a solution, using Duration.

Alternative solution:

```java
(int) ((System.nanoTime() - startTime) / 1_000_000);
```

See: [javaplayground](https://dev.java/p?id=9ef4f6d142d27048829e237f) showing the differences between both castings


sfeilmeier (approved):
No blocking issues found. The duration calculation now correctly converts nanoseconds to milliseconds before casting to int, avoiding overflow for solver runtimes above ~2.14 seconds.

PR #3956, opened 2026-09-04
HTTP Bridge: Replace thread pool

simon-lgb:
This change fixes an issue that prevented http bridge requests from running concurrently, as discussed in https://community.openems.io/t/broken-thread-pools-in-openems/10947.

With this appraoch, I tried to mirror the original intended behavior as closely as possible. However, moving the initialization of the maximum pool size from a method to the constructor would enable the custom subclass of Semaphore to be dropped entirely and the code could be simplified quite a bit. Also, the debug logging could be reduced if desired to remove the need for separate counters.

Originally, I tested this using the MetadataOdoo provider. I also expanded the debug logging there and fixed a deadlock with one of the ConcurrentHashMaps that I discovered once several requests were running concurrently. However, the use of the HTTP bridge in that class was removed in https://github.com/OpenEMS/openems/commit/aa79e27a2f333a7a132e593c1e333ca9678cd4c3, so those changes are not included in this PR. If desired, I can also revert that commit and include those additional changes.

The flawed thread pool pattern exists in other places in the code as well, however this initial PR only touches the HTTP bridge.

I'm looking forward to your feedback :)

### DIFF
--- a/io.openems.common.bridge.http/src/io/openems/common/bridge/http/AsyncBridgeHttpExecutor.java
+++ b/io.openems.common.bridge.http/src/io/openems/common/bridge/http/AsyncBridgeHttpExecutor.java
@@ -91,7 +91,7 @@ public class AsyncBridgeHttpExecutor implements BridgeHttpExecutor {
 	}
 
 	@Deactivate
-	private void deactivate() {
+	protected void deactivate() {
 		ThreadPoolUtils.shutdownAndAwaitTermination(this.scheduler, 0);
 		ThreadPoolUtils.shutdownAndAwaitTermination(this.executor, 0);
 	}
--- a/io.openems.common.bridge.http/src/io/openems/common/bridge/http/AsyncBridgeHttpExecutor.java
+++ b/io.openems.common.bridge.http/src/io/openems/common/bridge/http/AsyncBridgeHttpExecutor.java
@@ -1,9 +1,13 @@
 package io.openems.common.bridge.http;
 
 import java.util.Map;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
-import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
 
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Deactivate;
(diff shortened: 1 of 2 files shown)


io.openems.common.bridge.http/src/io/openems/common/bridge/http/AsyncBridgeHttpExecutor.java
```diff
@@ -16,41 +20,94 @@
...
+	private final ExecutorService executor = Executors.newVirtualThreadPerTaskExecutor();
+	private final MySemaphore semaphore = new MySemaphore(10);
+
+	private final AtomicLong activeCounter = new AtomicLong();
+	private final AtomicLong completedCounter = new AtomicLong();
+
 	@Override
 	public ScheduledFuture<?> schedule(Runnable task, DelayTimeProvider.Delay.DurationDelay durationDelay) {
-		return this.pool.schedule(task, durationDelay.getDuration().toMillis(), TimeUnit.MILLISECONDS);
+		return this.scheduler.schedule(() -> this.execute(task), durationDelay.getDuration().toMillis(),
```
Sn0w3y, line 63: I think this changes the semantics of the returned `ScheduledFuture`.

After the delay this future is already completed, because it only calls `execute(task)`. The real task can still be waiting on the semaphore at this point.

`HttpBridgeTimeServiceImpl` uses `future.cancel(false)` to stop a scheduled endpoint. With this change, canceling the future would not stop a task which was already submitted to the virtual thread executor and is waiting for a permit. It could still execute the HTTP request later.

I think the returned future should still be able to cancel the actual task, or we need another cancellation handling here?

Sn0w3y (commented):
I would add a test for this case: max concurrency = 1, first task blocks, second task gets scheduled and waits, cancel the returned future, release the first task -> second task must not execute.

Also please have a look at changing the semaphore size while tasks are active. Otherwise the idea looks good to me.

io.openems.common.bridge.http/src/io/openems/common/bridge/http/AsyncBridgeHttpExecutor.java
```diff
@@ -16,41 +20,94 @@
...
+				"Permits", (long) this.semaphore.availablePermits(), //
+				"Active", this.activeCounter.get(), //
+				"Pending", (long) this.semaphore.getQueueLength(), //
+				"Completed", this.completedCounter.get());
 	}
 
 	@Override
 	public void setMaximumPoolSize(int maximumPoolSize) {
-		this.pool.setMaximumPoolSize(maximumPoolSize);
+		this.semaphore.setTotalPermits(maximumPoolSize);
```
Sn0w3y, line 110: This also changes the validation behavior compared to `ThreadPoolExecutor#setMaximumPoolSize()`.

The old implementation rejected values `<= 0`, while the semaphore implementation seems to accept them. Should we keep the previous behavior here and reject invalid values explicitly?

io.openems.common.bridge.http/src/io/openems/common/bridge/http/AsyncBridgeHttpExecutor.java
michaelgrill: Someone from [stackoverflow](https://stackoverflow.com/a/79748430) addressed that issue as well and made this implementation: https://github.com/eager-zh/virtual-thread-scheduler/blob/main/src/main/java/com/github/scheduling/ThreadPerTaskScheduledExecutorService.java 
(i havent verified his solution, but it may help here)

michaelgrill:
In general i would prefer a separate/reusable class which can be used for the other places as well where a ScheduledThreadPool is used. Would make it easier to switch once the implemenation works

PR #3961, opened 2026-09-08
[Edge] Add EVSE Charge-Point support for the Voltie AC wallbox

voltie-eu:
# EVSE Charge-Point: Voltie AC wallbox

Adds an EVSE Charge-Point implementation for the Voltie AC wallbox over Modbus/TCP.

One new bundle `io.openems.edge.evse.chargepoint.voltie` plus the two standard lines in
`io.openems.edge.application/EdgeApp.bndrun`. No shared or core code is touched.

We are the manufacturer of the device and will maintain the driver.

## What it adds

Component `Evse.ChargePoint.Voltie`, implementing `EvseChargePoint`, `ElectricityMeter`,
`ModbusComponent` and `TimedataProvider`:

- Per-phase voltage, current and active power; total active power; session and lifetime energy.
- Charge-current set-point (6..32 A) and start/stop.
- 1↔3 phase switching as `PhaseSwitchAbility.Manual`, offered only where the device reports support.
- Diagnostic Channels: EVSE state, charge-stop reason, phases in use, capability bitmask, hardware
  current limit, watchdog timeout, DLM mode, serial numbers, firmware build.
- State-Channels: `EvseFault`, `FirmwareOutdated`, `PhaseSwitchFailed`.
- `readOnly` mode.

## Device and protocol

- Modbus/TCP, port 502, default Unit-ID 11.
- FC03/FC04 for reads, FC06 for writes; FC16 is rejected by the device.
- Register map source: Voltie Modbus API documentation v1.3 (2026-08-26). Only registers and codes
  described there are implemented.
- Minimum firmware build 357, the version the documented register map applies from. The readable
  area has grown with the firmware, and an over-long read is answered with exception `0x02`, so the
  driver probes the firmware build first and only then adds the full register map. On older
  firmware it raises `FirmwareOutdated` and adds nothing else.
- Two read blocks: status `0x0000..0x001A` (27 registers) and meter `0x2000..0x201D` (30). Writes:
  charging enable `0x000C`, current limit `0x0014`, forced single phase `0x0016`.

## Design decisions worth noting

**Both read tasks are `Priority.LOW`.** The charger serves a single Modbus master and forwards
requests only in the gaps between its own internal polls, roughly one per 500 ms. Against the
bridge's 500 ms timeout, two reads in one Cycle leave no margin, so the bridge round-robins one
block per Cycle and each block refreshes about every 2 s. The driver therefore needs its own
`Bridge.Modbus.Tcp`.

**`ActiveProductionEnergy` is integrated from `ActivePower`.** The device's lifetime energy register
only advances when a session ends, so mapping it directly would report 0 Wh for a whole session and
then jump. It is exposed separately as `LifetimeEnergy`.

**The set-point ability is `Ampere`.** The current-limit register takes milliamperes but the
firmware truncates to whole amperes, so a milliampere ability would let the cluster allocate
set-points the device discards.

**Phase switching is gated on the device's capability bitmask** (`0x0019` bit 0), which is set
exactly when the phase-switch register is writable. The documentation asks clients to read that bit
rather than probe with a write, because on unsupported hardware the write is rejected and a
retrying client can interrupt charging. No configuration option is therefore needed.

**The charging-enable register is written once per intended state change.** Writing 1 only permits
charging; with no vehicle connected the write is accepted but has no effect, so a client that
rewrites until the read-back matches would loop indefinitely.

## Hardware validation

Two production chargers on firmware build 360, driven by a local OpenEMS Edge.

Three-phase charger with a vehicle charging: the 27-register block reads cleanly; the capability
bit is set and the phase-switch ability is offered; `PhasesInUse` comes from the device and follows
the physical state (3 → 1 → 3); Channel values matched independent Modbus reads; the set-point
reached the vehicle; phase switching completed in both directions with a single write each; a
sustained run produced no `ModbusCommunicationFailed`, with request latency in the tens of
milliseconds.

Single-phase charger with the capability bit clear: no phase-switch ability is offered even with
three-phase wiring configured. Read-only mode issued no writes.

## Not included

- No UI live widget. The `evse-chargepoint.ts` case is easy to add, but it needs an
  `images.EVSE.*` entry, and those values live in theme files outside this repository — happy to
  add the case if you tell us how you would like the asset handled.
- Channels carry `PersistencePriority.LOW`; tell us if you would rather have specific values
  persisted differently.


voltie-eu:
We found and fixed an issue in the charging-enable handling: the enable was written while no vehicle was connected, so a vehicle plugged in later was never started. Details are in the updated description. The new push needs a workflow approval to run CI. Happy to answer any questions for the review.

io.openems.edge.evse.chargepoint.voltie/src/io/openems/edge/evse/chargepoint/voltie/EvseChargePointVoltieImpl.java
```diff
@@ -0,0 +1,803 @@
...
+	 * rejected write answers a Modbus exception, which the bridge reports as
+	 * {@link ExecuteState.Error}.
+	 *
+	 * @param state the {@link ExecuteState}
+	 */
+	protected void onControlWriteExecuted(ExecuteState state) {
+		if (!(state instanceof ExecuteState.Error)) {
+			return;
+		}
+		var now = Instant.now();
```
Sn0w3y: Could we use the OpenEMS clock instead of Instant.now() for the write/retry timers?

io.openems.edge.evse.chargepoint.voltie/src/io/openems/edge/evse/chargepoint/voltie/EvseChargePointVoltieImpl.java
```diff
@@ -0,0 +1,803 @@
...
+	 * rejects the write with Modbus exception 0x03 when control-by-Modbus is
+	 * disabled, on relay/EEPROM error, or while a charging session is active; the
+	 * bridge reports this as {@link ExecuteState.Error}. Hardware that does not
+	 * support phase switching at all is already excluded by the capability
+	 * bitmask, see {@link #getPhaseSwitchAbility()}.
+	 *
+	 * @param state the {@link ExecuteState}
+	 */
+	protected void onPhaseSwitchWriteExecuted(ExecuteState state) {
+		if (!(state instanceof ExecuteState.Error)) {
```
Sn0w3y: ExecuteState.Error is not limited to Modbus exception responses; AbstractWriteTask wraps any exception from executeRequest() in it, including transport/connection failures. Could we inspect Error.exception() here and only latch phaseSwitchUnavailable for a confirmed device rejection such as ModbusSlaveException with exception code 0x03? Otherwise a communication failure can permanently disable phase switching until the component configuration is changed... So it does not "self-recover" so to say :)

Sn0w3y (commented):
Please see my Comments :)

PR #3967, opened 2026-09-09
[UI] Enable TypeScript strict checks on login, overview, and registration

hvgollar:
- Remove '@ts-strict-ignore' from login, overview, and registration.
- Widen types / use definite assignment and optional chaining so strict checks pass.
- Treat a missing login password as "" (matches preprocessCredentials return type).

### DIFF
--- a/ui/src/app/index/login.component.ts
+++ b/ui/src/app/index/login.component.ts
@@ -1,5 +1,4 @@
-// @ts-strict-ignore
-import { AfterContentChecked, ChangeDetectorRef, Component, computed, effect, inject, OnDestroy, ChangeDetectionStrategy, } from "@angular/core";
+import { AfterContentChecked, ChangeDetectionStrategy, ChangeDetectorRef, Component, computed, effect, inject, OnDestroy, } from "@angular/core";
 import { FormGroup } from "@angular/forms";
 import { ActivatedRoute, Router } from "@angular/router";
 import { Capacitor } from "@capacitor/core";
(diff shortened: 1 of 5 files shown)


ui/src/app/index/overview/overview.component.ts
```diff
@@ -155,7 +154,7 @@ export class OverViewComponent implements ViewWillEnter, OnDestroy {
                 return;
             }
 
-            const searchParamsObj = {};
+            const searchParamsObj: Record<string, unknown> = {};
```
mat-kreisel, line 157: why not use the ChosenFilter["value"] from above instead of unknown?

ui/src/app/index/login.component.ts
```diff
@@ -72,9 +71,9 @@ export class LoginComponent implements ViewWillEnter, AfterContentChecked, OnDes
      * @param username The username
      * @returns Trimmed credentials
      */
-    public static preprocessCredentials(password: string, username?: string): { password: string; username?: string } {
+    public static preprocessCredentials(password: string | null, username?: string | null): { password: string; username?: string } {
```
mat-kreisel, line 74: why accept null for password and username? I don't see any call that needs it

sfeilmeier:
@mat-kreisel: Please update from develop to fix conflicts first. Your review comments should have been handled by my commit yesterday night already: https://github.com/OpenEMS/openems/commit/fdd9c7fd517cfff9b29a3e5d236ec440b2eb024c#diff-c681b02e49142717836a06f5598c9ad163132c5a772e392134db2c6024c7bddc

sfeilmeier (approved):

ui/src/app/index/overview/overview.component.ts
hvgollar: Yes, thank for pointing on this @mat-kreisel . We should have used ChosenFilter["value"]  instead of unknown.
Do you want me to go ahead make this change in the PR?

ui/src/app/index/login.component.ts
hvgollar: @mat-kreisel
" | null is there because login.spec.ts already calls preprocessCredentials(null) / (null, null). Production login only passes string and optional username.

If we drop | null, those spec calls will fail strict type-check"

PR #3970, opened 2026-09-09
Fix: Evse Controller Automatic phase switch never triggers with slower Core-Cycle-Time

vision1727:
Problem: 
Currently automatic phase switching uses a fixed rate of 60 samples in 2 minutes. If the core cycle time is higher than about 2 seconds it will probably never shift. 
Reason for that is that I'm using a core cycle time of 4 seconds, otherwise used hardware isn't fast enough. 

Fix:
Switching from fixed 60 sample window to a dynamic one. The sample window of the dynamic window depends on the average cycle time of taken samples x factor 90%. Reason for the 90% is that there might be a deviation of the cycle time.

Tests:
The test files are also changed to the same dynamic window.


vision1727:
locally   /tools/prepare-commit.sh script. runs without a mistake. tested this 3 times. the plcnext runs with an error here don`t know exactly why.

PR #3973, opened 2026-09-10
[Edge] PvInverter.SellToGridLimit: clamp setpoint to 0

### LINKED ISSUE #3972: PV-Inverter ActivePowerLimit goes negative & is written unchecked to unsigned Modbus registers
### Description

A PV-Inverter power limit can go below zero inside OpenEMS, and two drivers write that value into an unsigned Modbus register without checking it. The register wraps. The inverter runs unlimited at the moment the limit is needed. We found this in August 2026. Two separate defects line up to cause it. Each is a few lines of code.

### 1. The Sell-to-Grid-Limit controller produces negative setpoints

`ControllerPvInverterSellToGridLimitImpl.calculateRequiredPower` adds three numbers and writes the sum to the inverter:

```
setpoint = gridPower + pvInverterPower + maximumSellToGridPower
```

Nothing clips the sum to 0. So when exporting more than its managed PV production plus the configured limit gets a negative setpoint: eather a unmanage PV inverter or a battery discharging into the grid. Given a limit of 5000 W, a grid reading of -6000 W and 500 W of managed production give -500 W. The controller shouldn't ask for this.

### 2. Two drivers pass the value to unsigned registers unchecked

`ManagedSymmetricPvInverter.ACTIVE_POWER_LIMIT` is an `IntegerDoc` with no range, so each driver has to guard itself. The SunSpec drivers (Kaco, SMA, Fronius, SolarEdge, Kostal, Solarlog, Kaco Hybrid 10) do: `SetPvLimitHandler` converts watts to percent and clamps the result to 1 to 100 %. That clamp is the only reason they're safe and exists because SunSpecs inverter go to standby if set to 0, see `io.openems.edge.bridge.modbus/src/io/openems/edge/bridge/modbus/sunspec/pvinverter/SetPvLimitHandler.java:79-86`.

`HuaweiPvInverterImpl` maps the channel to `UnsignedWordElement(40235)` with `SCALE_FACTOR_2`. -500 W becomes raw -5. `UnsignedWordElement.valueToByteBuffer` calls `buff.putShort(i.shortValue())` with no range check, so the wire carries 65531 and the inverter reads a limit of 6553.1 kW. `HuaweiSmartLoggerPvInverterImpl` does the same through `UnsignedDoublewordElement(40424)` and `SCALE_FACTOR_3`, into 32 bits. `PvInverterCluster` caps each member at its `MaxActivePower` but has no floor, so it forwards negatives unchanged.

We haven't run the two Huawei drivers on hardware; the description above comes from reading the code. The affected files are identical in 2026.7.0, in release 2026.9.0 and on develop as of 2026-09-08. Our own driver for another inverter brand has the same register layout and showed the same result. The fix there was a clamp in the driver's BEFORE_WRITE handler.

### Suggested fix

Clamp the sum at 0 in `ControllerPvInverterSellToGridLimitImpl.run()` before the ramp logic. That's one line and one test. A PR follows.

Clamp the setpoint to the range 0 to `MaxActivePower` in both Huawei drivers before the Modbus write, the way `SetPvLimitHandler` does it. A shared helper in the PV-Inverter API would stop the next driver from repeating this, but the two clamps shouldn't wait for it.

A clamp at 0 also keeps a negative `maximumSellToGridPower` meaningful. It becomes "import at least X W", and the inverter never sees a value below 0.

### Screenshots

None

### Operating System

Debian 12, OpenEMS Edge 2026.7.0. The defect is in the Java code and doesn't depend on the OS.

Controller half, no hardware needed:

1. Configure `Controller.PvInverter.SellToGridLimit` with the default `maximumSellToGridPower = 5000`, a grid meter and any PV inverter.
2. Set the meter to `ActivePower = -6000` (export) and the PV inverter to `ActivePower = 500`.
3. `pvInverter0/ActivePowerLimit` is now -500.

Driver half, Huawei PV inverter or SmartLogger:

1. Write -500 to `pvInverter0/ActivePowerLimit`. The controller above does it, and so does the REST API or `Controller.PvInverter.FixPowerLimit`, whose config field is unbounded too.
2. Read register 40235 (or 40424). It holds the two's-complement wrap of the value, not 0.

JUnit test for the controller half:

```java
new ControllerTest(new ControllerPvInverterSellToGridLimitImpl())
    .addReference("componentManager", new DummyComponentManager())
    .addComponent(new DummyElectricityMeter("meter0"))
    .addComponent(new DummyManagedSymmetricPvInverter("pvInverter0"))
    .activate(MyConfig.create()
        .setId("ctrl0").setMeterId("meter0").setPvInverterId("pvInverter0")
        .setAsymmetricMode(false).setMaximumSellToGridPower(5_000).build())
    .next(new TestCase()
        .input("meter0", ACTIVE_POWER, -6000)
        .input("pvInverter0", ACTIVE_POWER, 500)
        .output("pvInverter0", ACTIVE_POWER_LIMIT, 0)); // currently -500
```

JonasBechler:
Fixes #3972

The setpoint calculated by `Controller.PvInverter.SellToGridLimit`
(`gridPower + pvInverterPower + maximumSellToGridPower`) has no lower bound
and goes negative whenever the site exports more than
"managed production + limit", e.g. with a second unmanaged PV-Inverter or a
discharging battery. Drivers that write the value to an unsigned Modbus
register wrap it to a large positive number and lift the limit instead of
tightening it (details in #3972).

This PR clamps the setpoint at 0 before the ramp logic and the write.
A negative `maximumSellToGridPower` ("always import at least X W") stays usable.

Tests: two new cases in `ControllerPvInverterSellToGridLimitImplTest`
(default 5 kW limit on an over-exporting site, and a negative limit).
`checkstyleMain`/`checkstyleTest` pass for the bundle.


sfeilmeier (approved):
Thank you!

PR #3975, opened 2026-09-10
[Common] JettyUtils: block on Response.write() to prevent IncompleteRead

### LINKED ISSUE #3974: REST API truncates responses above 44 KB: JettyUtils writes with Callback.NOOP & the handler completes the request before the body is flushed
### Description

Every REST or JSON-RPC response above roughly 44 KB reaches the client cut off. `JettyUtils.sendOkResponse` hands the body to Jetty with `Callback.NOOP` and returns at once, `RestHandler.handle` calls `callback.succeeded()` on the next line, and Jetty finishes the exchange while the body write is still running.

We hit this in March 2026 with a polling script that reads channels from OpenEMS Edge through the REST API. Python's `http.client` raised:

```
IncompleteRead(44758 bytes read, 92238 more expected)
```

The body was 136996 bytes long. 44758 of them arrived, then the connection closed. A single-channel `GET /rest/channel/_sum/GridActivePower` is a few hundred bytes and works every time, which is why the bug stays hidden until someone asks for a large result.

### 1. Response.write() is asynchronous in Jetty 12

PR #3026 moved OpenEMS to the Jetty 12 core API, and release 2026.9.0 ships `org.apache.felix.http.jetty12` 2.0.4 with Jetty 12.1.11 inside. In that API `Response.write(boolean last, ByteBuffer content, Callback callback)` returns before the bytes are on the wire and reports completion through the callback.

`sendOkResponse` at `JettyUtils.java:91` passes `Callback.NOOP`, so nothing waits for that completion. `sendErrorResponse` at line 120 does the same thing.

### 2. The handler completes the request before the write completes

`RestHandler.handle` in `io.openems.edge.controller.api.rest` calls `sendOkResponse` and then `callback.succeeded()` at line 101. Jetty's handler contract wants that callback completed after the last write has completed; completing it while a write is in flight lets Jetty close out the response and recycle the connection with bytes still pending. A body that fits into one write pass is already gone when `succeeded()` runs. Anything larger is cut where the first pass ended, and 44758 bytes is where it ended for us.

`io.openems.backend.b2brest/.../RestHandler.java:56` has the same two lines in the same order, so the Backend-to-Backend REST API is affected as well. Both handlers are otherwise fine; the fix belongs in the two helpers they share.

`JettyUtils.java` is identical in release 2026.9.0 and on develop as of 2026-09-08 (head `189ac916fd`).

### Suggested fix

Replace `Callback.NOOP` with `Blocker.callback()` and call `blocker.block()` right after `response.write()`, in both helpers. `org.eclipse.jetty.util.Blocker.Callback` is `AutoCloseable`; `block()` returns once Jetty reports the write as succeeded and throws if it failed. `sendOkResponse` turns that failure into an `OpenemsException`, `sendErrorResponse` drops it because it's already on the error path.

Method signatures, callers and the `callback.succeeded()` lines stay as they are, since by the time either helper returns the body has been handed to the socket. Our fork has carried this change since 2026-03-20. A PR follows.

### Screenshots

None

### Operating System

Debian 12, OpenEMS Edge 2026.9.0. The defect is in the Java code and doesn't depend on the OS.

No hardware needed, a Simulator setup does:

1. Run OpenEMS Edge 2026.9.0 with `Controller.Api.Rest.ReadOnly` on the default port 8084 and enough components that the full channel list is larger than 44 KB. About 300 channels are enough; a `_sum` component plus a simulated ESS, meter and PV inverter get there.
2. Request every channel through the regex form of the channel endpoint:

   ```
   curl -s -u x:user "http://localhost:8084/rest/channel/.*/.*" -o all.json
   ```

   curl exits with code 18, `transfer closed with <n> bytes remaining to read`, and `all.json` ends mid-object.
3. The same request from Python fails with the error above:

   ```python
   import urllib.request
   urllib.request.urlopen("http://x:user@localhost:8084/rest/channel/.*/.*").read()
   # http.client.IncompleteRead: IncompleteRead(<n> bytes read, <m> more expected)
   ```

4. Request a single channel, `GET /rest/channel/_sum/GridActivePower`. It returns complete, because the body is smaller than one write pass.

With `Blocker.callback()` in place of `Callback.NOOP`, step 2 returns the full JSON array and `wc -c all.json` matches the `Content-Length` header.

JonasBechler:
Fixes #3974

REST & JSON-RPC responses above roughly 44 KB reach the client truncated. `JettyUtils.sendOkResponse` hands the body to Jetty 12's asynchronous `Response.write()` with `Callback.NOOP` and returns at once; `RestHandler.handle` then calls `callback.succeeded()` and Jetty finishes the exchange with bytes still pending. A Python client reports `IncompleteRead(44758 bytes read, 92238 more expected)`.

This replaces `Callback.NOOP` with `Blocker.callback()` and `blocker.block()` in `sendOkResponse` & `sendErrorResponse`, so each helper returns only after Jetty confirms the write. `sendOkResponse` converts a failed write into an `OpenemsException`, `sendErrorResponse` drops it since it's already on the error path. 13 insertions, 3 deletions, one file; signatures and callers in `io.openems.edge.controller.api.rest` and `io.openems.backend.b2brest` are unchanged.

Tested: `:io.openems.common:test`, `:io.openems.edge.controller.api.rest:test` and checkstyle for `io.openems.common` pass on top of 2026.9.0.


sfeilmeier (approved):
Wow, thank you for the thorough analysis and the fix 🚀👍

PR #3976, opened 2026-09-10
[UI] Enable strict TypeScript checking for date-picker popover

### LINKED ISSUE #3995: UI: turn on `strict mode`
### Component

UI

### Description

In UI currently some checks are disabled. This ticket collects actions that try to change this, e.g.
- Remove `@ts-strict-ignore` from ts-files (https://github.com/allegro/typescript-strict-plugin)
- Enable `strictTemplates` in `tsconfig.json` (https://angular.dev/reference/configs/angular-compiler-options#stricttemplates)
- Enable `strictInjectionParameters` in `tsconfig.json` (https://angular.dev/reference/configs/angular-compiler-options#strictinjectionparameters)
- Enable `"noUncheckedIndexedAccess": true`

PunugupatiSaikumar:
## Changes
- Enable strict checking for the date-picker popover.
- Declare the callback supplied when creating the popover.
- Preserve existing date-function behavior for undefined input.
- Correct the allowed-history-periods helper's input and return types

### DIFF
--- a/ui/src/app/shared/components/pickdate/popover/popover.component.ts
+++ b/ui/src/app/shared/components/pickdate/popover/popover.component.ts
@@ -1,5 +1,4 @@
-// @ts-strict-ignore
-import { ChangeDetectorRef, Component, Input, OnInit, ChangeDetectionStrategy } from "@angular/core";
+import { ChangeDetectionStrategy, ChangeDetectorRef, Component, Input, OnInit } from "@angular/core";
 import { PopoverController } from "@ionic/angular";
 import { TranslateService } from "@ngx-translate/core";
 import { CalAnimation, IAngularMyDpOptions, IMyDate, IMyDateRangeModel } from "@nodro7/angular-mydatepicker";
@@ -24,7 +23,7 @@ import { Edge } from "../../edge/edge";
     ],
 })
 export class PickDatePopoverComponent implements OnInit {
-    @Input() public setDateRange: (period: DefaultTypes.HistoryPeriod) => void;
+    @Input() public setDateRange!: (period: DefaultTypes.HistoryPeriod) => void;
     @Input() public edge: Edge | null = null;
     @Input() public historyPeriods: DefaultTypes.PeriodStringValues[] = [];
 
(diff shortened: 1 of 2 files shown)


ui/src/app/shared/components/pickdate/popover/popover.component.ts
```diff
@@ -121,10 +120,12 @@ export class PickDatePopoverComponent implements OnInit {
 
     ngOnInit() {
         this.locale = Language.getCurrentLanguage().key;
+        const firstSetupProtocol = this.edge?.firstSetupProtocol;
+        const dateInput = firstSetupProtocol === undefined ? NaN : firstSetupProtocol;
         // Restrict user to pick date before ibn-date
         this.myDpOptions.disableUntil = {
-            day: Utils.subtractSafely(getDate(this.edge?.firstSetupProtocol), 1) ?? 1,
-            month: Utils.addSafely(getMonth(this.edge?.firstSetupProtocol), 1) ?? 1,
+            day: Utils.subtractSafely(getDate(dateInput), 1) ?? 1,
```
sfeilmeier: Utils is deprecated. Replace with `NumberUtils`

ui/src/app/shared/components/pickdate/popover/popover.component.ts
```diff
@@ -121,10 +120,12 @@ export class PickDatePopoverComponent implements OnInit {
...
+            month: Utils.addSafely(getMonth(dateInput), 1) ?? 1,
             year: this.edge?.firstSetupProtocol?.getFullYear() ?? 2013,
```
sfeilmeier: That entire logic was flawed. Please see my changes.

sfeilmeier:
## Review of session changes

1. **`Utils.subtractSafely`/`Utils.addSafely` → `NumberUtils.subtractSafely`/`NumberUtils.addSafely`** — good cleanup, the `Utils` variants are marked `@deprecated` in favor of `NumberUtils`. `Utils` import removed since it's now unused.

2. **Removed the `NaN`-substitution hack** in favor of an explicit branch: `disableUntil` is now only computed from `firstSetupProtocol` when it's actually defined; otherwise it falls back to a named default. This is clearer and removes the fragile `getDate(NaN)`/`NumberUtils.subtractSafely(NaN, 1) ?? 1` path we flagged earlier (which silently produced `NaN` instead of falling back). Good fix — this also resolves the pre-existing latent bug, not just the typing issue.

3. **`DEFAULT_DISABLE_UNTIL` constant** — extracting `{ day: 1, month: 1, year: 2013 }` into `private static readonly DEFAULT_DISABLE_UNTIL: IMyDate` removes duplication between the `myDpOptions` field initializer and the `ngOnInit()` fallback, and gives the magic literal a name. Good improvement; note the original `// TODO start with date since the edge is available` comment was dropped when the initializer changed to reference the constant — that TODO is still relevant (the field initializer still defaults to `2013` before `ngOnInit` runs), consider re-adding it near the constant declaration if that context should be preserved.

4. **Ternary formatting** — the final ternary is a bit dense (7 lines) but auto-formatted consistently with the project's Prettier/ESLint config (verified: lint passes clean). Readability is acceptable given it now clearly branches on a single condition (`firstSetupProtocol !== undefined`) instead of threading a synthetic `NaN` through math helpers.

No remaining concerns — this is a solid, self-contained improvement on top of the already-reviewed diff, and it's fully covered by the existing `EdgePermission`/`PickdatePopover` test suites (32/32 passing) plus `tsc --noEmit` and `ng lint` with zero errors.

sfeilmeier (approved):

